PR #7, opened 2019-07-16
Add API module, with basic mocked endpoint for namespaces.

newswangerd:
- Add `axios` for making HTTP requests.
- Add `axios-mock-adapter` for mocking HTTP requests.


src/api/base.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    constructor() {
+        this.http = axios.create({
+            baseURL: this.apiBaseURL,
+        });
+    }
+
+    list(params: object, apiPath?: string) {
+        const path = apiPath || this.apiPath;
+        return this.http.get(path, params);
```
ironfroggy: is this missing the prefix `this.apiBaseURL` used in all the other method below?

newswangerd: Oh, this is actually the only correct one. I meant to take the base url out of all the methods.

PR #8, opened 2019-07-19
Add edit namespaces form

newswangerd:
fixes: https://github.com/ansible/galaxy-dev/issues/26

## Edit Details
Screen Shot 2019-07-22 at 9 28 13 AM

## Edit Resources
Screen Shot 2019-07-22 at 9 28 35 AM

Todo:
- [x] Add breakpoint for resources form
- [x] Add breakpoint for tabs


ironfroggy:
Resizable textarea fields are useful, but in this case it breaks the layout, so maybe we should disable it?

Screenshot_20190722_132329
Screenshot_20190722_132027

src/components/namespace-form/namespace-form.tsx
```diff
@@ -0,0 +1,215 @@
...
+                                placeholder='Link Url'
+                                value={this.state.newLinkURL}
+                                onChange={value =>
+                                    this.setState({
+                                        newLinkURL: value,
+                                    })
+                                }
+                            />
+                        </div>
+                        <div className='link-button'>
```
ironfroggy: Users are probably going to try pressing [enter], can we include that to add the new link?

src/components/namespace-form/namespace-form.tsx
```diff
@@ -0,0 +1,215 @@
...
+                                size='md'
+                            />
+                        </div>
+                    </div>
+                </FormGroup>
+            </Form>
+        );
+    }
+
+    private udateField(value, event) {
```
ironfroggy: mispelled method here

PR #9, opened 2019-07-22
Add better documentation for running the app to the README

newswangerd:


profiles/local-frontend-and-api.js
```diff
@@ -1,17 +1,17 @@
 /*global module*/
 
 const SECTION = 'insights';
-const APP_ID = 'starter';
```
ironfroggy: Should you adjust `local-frontend.js` too?

README.md
```diff
@@ -1,24 +1,52 @@
...
+REALPATH=`python2 -c 'import os,sys;print os.path.realpath(sys.argv[1])' $SPANDX_CONFIG`
+```
+
+to use `python` instead of `python2`.
 
-Note: You will need to set up the Insights environment if you want to develop with the starter app due to the consumption of the chroming service as well as setting up your global/app navigation through the API.
+### Run Automation Hub
 
-## Build app
+Once the insights proxy is running simply execute
```
ironfroggy: Its easy to figure out, but maybe be clear they have to both run simultaneous in two different terminals?

README.md
```diff
@@ -1,24 +1,52 @@
...
+
+#### NOTE
+
+If you are on a Mac, you might have to make a small change to the `insights-proxy/scripts/run.sh` script. Update this line
+
+```
+REALPATH=`python2 -c 'import os,sys;print os.path.realpath(sys.argv[1])' $SPANDX_CONFIG`
+```
+
+to use `python` instead of `python2`.
```
ironfroggy: You shouldn't have changes to checked-in code that some users have to make, and then always carefully never check in.

maybe you can replace this line with

```
REALPATH=`realpath $SPANDX_CONFIG`
```

newswangerd: This is something that needs to get changed in the https://github.com/RedHatInsights/insights-proxy project. 

```
REALPATH=`realpath $SPANDX_CONFIG`
``` 

doesn't work on my mac either. I don't know how to change this file to work on everyone's system, so I think the best I can do for now is provide a work around until someone with better bash knowledge than me fixes the script.

PR #12, opened 2019-07-31
Add Partner detail page

newswangerd:
fixes: https://github.com/ansible/galaxy-dev/issues/25
fixes: https://github.com/ansible/galaxy-dev/issues/17


ironfroggy:
Screenshot_20190801_100414

By going to page 2 and then changing the items per page I can cause the pagination to be on a page that doesn't exist listing item ranges that don't exist.

src/api/base.ts
```diff
@@ -30,4 +35,19 @@ export class BaseAPI {
...
     }
+
+    private mapPageToOffset(p) {
+        const params = { ...p };
+        const pageSize =
+            parseInt(params['page_size']) || Constants.DEFAULT_PAGE_SIZE;
+        const page = parseInt(params['page']) || 1;
+
+        delete params['page'];
+        delete params['page_size'];
```
ironfroggy: the `p`->`params` copy is confusing when it seems to replace all the contents of the object anyway and never actually changes it, so what is the copy for? 

could we instead do `params = {}` and `parseInt(p[x])` so we have a clear delineation between the input and output params?

src/components/patternfly-wrappers/sort.tsx
```diff
@@ -0,0 +1,132 @@
...
+                (descending ? '-' : '') + field.id,
+            ),
+        );
+    }
+
+    private getIsDescending(params) {
+        const sort = params[this.props.sortParamName];
+        if (!sort) {
+            return true;
+        }
```
ironfroggy: it is not clear to me why descending is `true` by default, or why we would be in a situation where there is no sort. It may not be clear to another reader in the future. Can you add some comments here?

newswangerd:
> By going to page 2 and then changing the items per page I can cause the pagination to be on a page that doesn't exist listing item ranges that don't exist.

https://github.com/patternfly/patternfly-react/issues/2629

PR #15, opened 2019-08-02
Add developer guidelines

newswangerd:


developer_guidelines.md
```diff
@@ -0,0 +1,89 @@
...
+## Imports
+
+### Ordering
+
+Imports should be ordered as follows:
+
+1. Any imports from 'react'
+2. Stylesheets
+3. Any imports from 3rd party libraries such as patternfly, lodash, etc.
+4. Local imports
```
ironfroggy: What about ordering within those groups? Like should they be lexically ordered?

developer_guidelines.md
```diff
@@ -0,0 +1,89 @@
...
+
+Incorrect way to import:
+
+```
+import { CollectionList } from '../../components/collection-list/collection-list'
+import { PartnerHeader } from '../../components/headers/partner-header'
+```
+
+Not only is this a lot cleaner to read and update, but it also makes it easier to reorganize
+code without breaking imports.
```
ironfroggy: Can these be enforced by a linter?

PR #22, opened 2019-08-21
Add collections page

newswangerd:
Note, I still need to get the UX people to review this page, so the design might change slightly.

Fixes: https://github.com/ansible/galaxy-dev/issues/23


src/components/card-list-switcher/card-list-switcher.tsx
```diff
@@ -0,0 +1,49 @@
...
+    updateParams: (params) => void;
+}
+
+export class CardListSwitcher extends React.Component<IProps> {
+    render() {
+        let disp = this.props.params.view_type;
+        const { updateParams, params } = this.props;
+
+        if (!disp) {
+            disp = 'card';
```
ironfroggy: Should this be set with `defaultProps`?

src/components/card-list-switcher/card-list-switcher.tsx
```diff
@@ -0,0 +1,49 @@
...
+                    }
+                    onClick={() =>
+                        updateParams(
+                            ParamHelper.setParam(params, 'view_type', 'card'),
+                        )
+                    }
+                />
+                <ListIcon
+                    className={
+                        'icon clickable ' + (disp === 'list' ? 'selected' : '')
```
ironfroggy: if they same `className` is used on both, should it be computed in a variable first?

src/components/cards/collection-card.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+    render() {
+        const { name, description, latest_version, namespace } = this.props;
+
+        return (
+            <Card className='collection-card-container'>
+                <CardHead className='logo-row'>
+                    <Logo
+                        image={namespace.avatar_url}
+                        alt={namespace.company + ' logo'}
```
ironfroggy: Is `company` a required/guaranteed field? Can it be `null`, which would break the string concat?

src/components/cards/collection-card.tsx
```diff
@@ -0,0 +1,101 @@
...
+                </CardBody>
+                <CardFooter className='type-container'>
+                    {Object.keys(latest_version.content_summary.contents).map(
+                        k => {
+                            return this.renderTypeCount(
+                                k,
+                                latest_version.content_summary.contents[k]
+                                    .length,
+                            );
+                        },
```
ironfroggy: the `k => { ... }` form is useful when there are multiple statements or no return value (for loops, not mapping). Here we can simply do `k => this.renderTypeCount(...)` right?

src/components/cards/collection-card.tsx
```diff
@@ -0,0 +1,101 @@
...
+                    )}
+                </CardFooter>
+            </Card>
+        );
+    }
+
+    private getDescription(d: string) {
+        if (d.length > this.MAX_DESCRIPTION_LENGTH) {
+            return d.slice(0, this.MAX_DESCRIPTION_LENGTH) + '...';
```
ironfroggy: Aren't we already doing this with the CSS `text-decoration: ellipsis` rule? Or, could we?

src/components/collection-list/collection-list-item.tsx
```diff
@@ -6,88 +6,109 @@ import {
...
-                                key='stats'
-                            >
-                                <div className='right-col entry'>
+        const cells = [];
+
+        if (showNamespace) {
+            cells.push(
+                <DataListCell isFilled={false} alignRight={false} key='ns'>
+                    <Logo
+                        alt={namespace.company + ' logo'}
```
ironfroggy: same previous question about `company` being potentially `null` or undefined

src/components/collection-list/collection-list-item.tsx
```diff
@@ -6,88 +6,109 @@ import {
...
+                </div>
+                <div className='entry pf-l-flex pf-m-wrap'>
+                    {latest_version.metadata.tags.map((tag, index) => (
+                        <Tag key={index}>{tag}</Tag>
+                    ))}
+                </div>
+                <div className='entry pf-l-flex pf-m-wrap'>
+                    {Object.keys(latest_version.content_summary.contents).map(
+                        k => {
+                            return (
```
ironfroggy: repeat previous `k => { return ... }` comment

src/components/logo/logo.tsx
```diff
@@ -0,0 +1,35 @@
...
+        // use inline css so we can set size
+        return (
+            <div
+                className={className}
+                style={{
+                    width: size,
+                    height: size,
+                    display: 'flex',
+                    justifyContent: 'center',
+                    alignItems: 'center',
```
ironfroggy: Should the non-variable styles by part of the CSS?

src/containers/search/search.scss
```diff
@@ -0,0 +1,31 @@
+@import '../../global-variables.scss';
+
+.collection-container {
+    // Should be the height of the page minus the header and the footer.
+    // This makes it so the footer stays at the bottom.
+    min-height: calc(100vh - 285px);
```
ironfroggy: How is the exact value of 285 determined?

src/containers/search/search.tsx
```diff
@@ -1,23 +1,218 @@
...
+                        />
+                    }
+                    title='Collections'
+                >
+                    <div
+                        style={{
+                            marginBottom: '16px',
+                            display: 'flex',
+                            justifyContent: 'space-between',
+                        }}
```
ironfroggy: I think this belongs in a layout CSS class. Spreading the styling between stylesheets and direct style attributes might make it difficult to trace later, tracking properties down between disparate sources.

That's just my thoughts, tho.

src/containers/search/search.scss
newswangerd: By adding up the heights of the header, chrome, toolbar and footer. I know this is kind of a janky solution. I tried using `position: absolute` to pin the footer to the bottom, but that cause all kinds of other issues. Is there a better way to do this?

src/containers/search/search.tsx
newswangerd: Woops. I meant to put this in the css file when I added it.

src/components/cards/collection-card.tsx
newswangerd: In my experience, doing ellipses with css only really works if the div is one line.

src/components/card-list-switcher/card-list-switcher.tsx
newswangerd: I tried that and it doesn't work because `params` can be set without containing the `view_type` key.

src/components/logo/logo.tsx
newswangerd: I think I'd prefer to keep all the styling in one place.

PR #26, opened 2019-08-30
changes for build directory

ryandillinfelton:
This should be all of the changes necessary to start pushing to the build repo located [here](https://github.com/RedHatInsights/ansible-hub-ui-build)


config/dev.webpack.config.js
```diff
@@ -18,4 +18,7 @@ const newWebpackConfig = {
 module.exports = {
     ...newWebpackConfig,
     plugins,
+    paths: {
```
newswangerd: Something about this is causing webpack to break:

```
✖ ｢wds｣: Invalid configuration object. Webpack has been initialised using a configuration object that does not match the API schema.
 - configuration has an unknown property 'paths'. These properties are valid:
   object { amd?, bail?, cache?, context?, dependencies?, devServer?, devtool?, entry?, externals?, infrastructureLogging?, loader?, mode?, module?, name?, node?, optimization?, output?, parallelism?, performance?, plugins?, profile?, recordsInputPath?, recordsOutputPath?, recordsPath?, resolve?, resolveLoader?, serve?, stats?, target?, watch?, watchOptions? }
   For typos: please correct them.
   For loader options: webpack >= v2.0.0 no longer allows custom properties in configuration.
     Loaders should be updated to allow passing options via loader options in module.rules.
     Until loaders are updated one can use the LoaderOptionsPlugin to pass these options to the loader:
     plugins: [
       new webpack.LoaderOptionsPlugin({
         // test: /\.xxx$/, // may apply this only for some modules
         options: {
           paths: …
         }
       })
     ]
```

ryandillinfelton:
> @ryelo @ryandillinfelton can either of you give me some context for these changes?
> 
> Also what's the strategy for deployment? I see you've created another repo here: https://github.com/RedHatInsights/ansible-hub-ui-build. Is the plan for that repo to be a fork of this one?

@ryelo could probably provide a more in depth explanation than I can. But essentially we have separate build repositories that we deploy from. This repo will remain the main repo for the source files, the new one will hold the built versions of the app. 

As an example you can take a look at how the landing page is set up: ([Suild Repo](https://github.com/RedHatInsights/landing-page-frontend-build) | [Source Repo](https://github.com/RedHatInsights/landing-page-frontend))

We need to create a build repo for this app so that we can deploy it with the new Ansible bundle on cloud.redhat.com.

Hopefully that clears some things up. If not let me know and I'd be happy to explain anything else I can.

newswangerd:
Ah, I think I understand. So https://github.com/RedHatInsights/ansible-hub-ui-build is basically just going to contain the 'dist' directory from building this repo?

If I understand correctly, pushes to `master`, `ci-stable` etc. on this repo will trigger builds that land in the new repo?

ryelo:
@newswangerd Yep. That repo will just be the dist from this repo.

By default, master will push to ci-beta, then you'll have to make the other environments as branches here. However (and what I recommend) is that you set up a custom config like the landing page: https://github.com/RedHatInsights/landing-page-frontend/blob/master/.travis/custom_release.sh

In that case, when you push to master, you'll be building to ci-beta and qa-beta. If you push to master-stable, then it'll build out ci-stable and qa-stable.

config/dev.webpack.config.js
ryandillinfelton: Just pushed an update that should fix this

config/dev.webpack.config.js
```diff
@@ -9,6 +10,16 @@ const { config: webpackConfig, plugins } = config({
     https: true,
 });
 
+webpackConfig.serve = {
```
newswangerd: One question before I merge this: do any of these changes need to be added to `config/prod.webpack.config`?

ryelo: Yes they do

ryandillinfelton: I've added the changes to prod

PR #58, opened 2019-10-15
Adding a table of contents to plugin docs

romanblanco:
Fixes: https://github.com/ansible/galaxy-dev/issues/123

These changes introduce a table of content identical to [Ansible docs](https://docs.ansible.com/ansible/latest/modules/copy_module.html#see-also).

cc/ @newswangerd

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
         "react-markdown": "^4.2.2",
         "react-redux": "^5.1.1",
         "react-router-dom": "^4.2.2",
+        "react-router-hash-link": "^1.2.2",
         "redux": "^3.7.2",
         "redux-logger": "^3.0.6",
         "redux-promise-middleware": "^5.1.1"
--- a/src/components/collection-detail/render-plugin-doc.tsx
+++ b/src/components/collection-detail/render-plugin-doc.tsx
@@ -435,12 +435,14 @@ export class RenderPluginDoc extends React.Component<IProps, IState> {
     private renderSynopsis(doc: PluginDoc) {
         return (
             <React.Fragment>
-                <h2>Synopsis</h2>
-                <ul>
-                    {doc.description.map((d, i) => (
-                        <li key={i}>{this.applyDocFormatters(d)}</li>
-                    ))}
-                </ul>
+                <div id="synopsis">
+                    <h2>Synopsis</h2>
+                    <ul>
+                        {doc.description.map((d, i) => (
+                            <li key={i}>{this.applyDocFormatters(d)}</li>
+                        ))}
+                    </ul>
+                </div>
             </React.Fragment>
         );
     }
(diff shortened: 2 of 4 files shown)


romanblanco:
@newswangerd I've renamed `renderDescription` to `renderSynopsis` so it's consistent with how we call it in the content and menu. 

I'm not sure where does the content for "See Also" and "Status" comes from.

I'm guessing [renderRequirements](https://github.com/ansible/ansible-hub-ui/blob/master/src/components/collection-detail/render-plugin-doc.tsx#L597-L612) could be the content for "See Also", is that right?

src/components/collection-detail/render-plugin-doc.tsx
```diff
@@ -192,19 +203,56 @@ export class RenderPluginDoc extends React.Component<IProps, IState> {
...
 
+    private renderTableOfContents(content: any) {
+      console.log(content);
+      // FIXME: test content's presence
+      // FIXME: link inside of the current document
+      return (
+            <ul>
+              {content["synopsis"] !== null &&
+                  <li>
+                      <a href="#synopsis">Synopsis</a>
```
newswangerd: These links are probably going to break. You need to use the `Link` component from react router instead.

https://reacttraining.com/react-router/web/api/Link

src/components/collection-detail/render-plugin-doc.tsx
```diff
@@ -64,20 +64,31 @@ export class RenderPluginDoc extends React.Component<IProps, IState> {
...
+
         if (!this.state.renderError) {
             return (
                 <div className='pf-c-content'>
                     <h1>
                         {plugin.content_type} > {plugin.content_name}
                     </h1>
                     <br />
+                    {this.renderTableOfContents(content)}
                     {this.renderShortDescription(doc)}
```
newswangerd: Lets put the short description above the table of contents

newswangerd:
I decided not to render "See also" and "Status" because both of those fields are specific to modules inside ansible and don't make to much sense for collections, so I think you can probably leave them out for now

newswangerd:
I'm not sure why this is happening, but there used to be a space between the header and the end of the previous section which has disappeared in this PR, making the spacing on the documentation feel kind of cramped.

Original spacing
Screen Shot 2019-10-15 at 2 18 35 PM

this PR
Screen Shot 2019-10-15 at 2 18 59 PM

romanblanco:
> I'm not sure why this is happening, but there used to be a space between the header and the end of the previous section which has disappeared in this PR, making the spacing on the documentation feel kind of cramped.

I'm looking into it, the `<div>` around `<h2>` seems to be causing a problem.

I've tried to follow the [PatternFly 4 documentation for content](https://www.patternfly.org/v4/documentation/core/components/content#documentation), but as as the document contains elements as `<table>`, `<pre>`, etc., it does not allow me to use `<p>` which would fix the issue.

I keep looking for a different solution.

romanblanco:
I've tried moving the `id` linking the document content to the headers instead of wrapping the content inside of `<div>` (https://github.com/ansible/ansible-hub-ui/pull/58/commits/0b6df475349e90a4b06ac79f67a6bbb24965bc80).

However, I've noticed the link does not scroll to the point where the div begins, it just shifts the page a little (having the id's both on `<div>` and `<h1>`). I'm checking where's the problem

package.json
```diff
@@ -22,6 +22,7 @@
         "react-markdown": "^4.2.2",
         "react-redux": "^5.1.1",
         "react-router-dom": "^4.2.2",
+        "react-router-hash-link": "^1.2.2",
```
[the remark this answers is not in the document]
newswangerd: If we have to add this as a dependency, that's probably okay. We're going to need to solve this problem for markdown files too (because hashlinks are broken on markdown too), and I'm wondering if this library might help us fix that too

PR #63, opened 2019-10-17
Enable tags, remove sort

newswangerd:
- Enabling tags to be passed to the API
- Removing sort from collections page because the API doesn't support it yet


romanblanco:
Tags enabled on the API in https://github.com/ansible/galaxy-api/pull/105

PR #76, opened 2019-11-11
Add import status at end of import log.

newswangerd:
Fixes: https://github.com/ansible/galaxy-dev/issues/135

### DIFF
--- a/src/components/my-imports/import-console.tsx
+++ b/src/components/my-imports/import-console.tsx
@@ -93,6 +93,20 @@ export class ImportConsole extends React.Component<IProps, {}> {
                             </span>
                         </div>
                     ) : null}
+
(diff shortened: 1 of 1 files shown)


src/components/my-imports/import-console.tsx
```diff
@@ -93,6 +93,20 @@ export class ImportConsole extends React.Component<IProps, {}> {
                             </span>
                         </div>
                     ) : null}
+
+                    {task.state === PulpStatus.completed && (
+                        <div className='message'>
+                            <br />
+                            <span className='success'>Done</span>
```
romanblanco: could we have a ternary operator here instead of two very similar objects?

```javascript
<span className={`${task.state !== PulpStatus.failed ? "success" : "failure"`}}>
```

PR #87, opened 2020-01-13
Adding namespace creation modal

romanblanco:
Fixes: https://github.com/ansible/galaxy-dev/issues/176
Closes: https://github.com/ansible/galaxy-dev/issues/176

Waiting for: 
- (create): ansible/galaxy-api#145
- (update): ansible/galaxy-api#149

TODO:

- [x] move modal from 'container' to a 'component' in namespace-list.tsx
- [x] refresh page after submit
- [x] success / failure alert
- [x] style the button in the toolbar
- [x] processing groups - comma separated string (tokenize using [chipgroup](https://patternfly-react.surge.sh/patternfly-4/documentation/react/components/chipgroup))
- [x] use [API](https://github.com/ansible/ansible-hub-ui/blob/master/src/containers/certification-dashboard/certification-dashboard.tsx#L107-L114) to test if user has permissions to see the button

### DIFF
--- a/src/components/namespace-form/namespace-form.tsx
+++ b/src/components/namespace-form/namespace-form.tsx
@@ -55,55 +55,6 @@ export class NamespaceForm extends React.Component<IProps, IState> {
                         <br />
 
                         <FormGroup
-                            fieldId='groups'
-                            label='Red Hat accounts'
-                            helperTextInvalid={errorMessages['groups']}
-                            isValid={!('groups' in errorMessages)}
-                        >
-                            <br />
-                            <ChipGroup>
-                                {this.props.namespace.groups.map(group => (
-                                    <Chip
-                                        key={group}
-                                        onClick={() => this.deleteItem(group)}
-                                    >
-                                        {group}
-                                    </Chip>
-                                ))}
-                            </ChipGroup>
-                            <div className='account-ids'>
-                                <div className='account-id'>
-                                    <TextInput
-                                        id='url'
-                                        type='text'
-                                        placeholder='Red Hat account ID'
-                                        value={this.state.newNamespaceGroup}
-                                        onChange={value =>
-                                            this.setState({
-                                                newNamespaceGroup: value,
-                                            })
-                                        }
-                                        onKeyDown={e => {
-                                            if (e.key === 'Enter') {
-                                                this.addGroup();
-                                            }
-                                        }}
-                                    />
-                                </div>
-                                <div className='account-add'>
-                                    <div className='clickable account-button'>
-                                        <PlusCircleIcon
-                                            onClick={() => this.addGroup()}
-                                            size='md'
-                                        />
-                                    </div>
-                                </div>
-                            </div>
-                        </FormGroup>
-
-                        <br />
-
-                        <FormGroup
                             fieldId='company'
                             label='Company name'
                             helperTextInvalid={errorMessages['company']}
@@ -124,6 +75,54 @@ export class NamespaceForm extends React.Component<IProps, IState> {
                         <NamespaceCard {...namespace} />
                     </div>
                 </div>
+
+                <FormGroup
+                    fieldId='groups'
+                    label='Namespace owners'
+                    helperTextInvalid={errorMessages['groups']}
+                    isValid={!('groups' in errorMessages)}
+                >
+                    <br />
+
+                    <ChipGroup>
+                        {this.props.namespace.groups.map(group => (
+                            <Chip
+                                key={group}
+                                onClick={() => this.deleteItem(group)}
+                            >
+                                {group}
+                            </Chip>
+                        ))}
+                    </ChipGroup>
+
+                    <div className='account-ids'>
+                        <br />
+                        <TextInput
+                            id='url'
+                            type='text'
+                            placeholder='Red Hat account ID'
+                            value={this.state.newNamespaceGroup}
+                            onChange={value =>
+                                this.setState({
+                                    newNamespaceGroup: value,
+                                })
+                            }
+                            onKeyDown={e => {
+                                if (e.key === 'Enter') {
+                                    this.addGroup();
+                                }
+                            }}
+                        />
+                        <div className='account-add'>
+                            <div className='clickable account-button'>
+                                <PlusCircleIcon
+                                    onClick={() => this.addGroup()}
+                                    size='md'
+                                />
+                            </div>
+                        </div>
+                    </div>
+                </FormGroup>
 
                 <FormGroup
                     fieldId='avatar_url'
--- a/src/components/namespace-form/namespace-form.tsx
+++ b/src/components/namespace-form/namespace-form.tsx
@@ -256,7 +256,7 @@ export class NamespaceForm extends React.Component<IProps, IState> {
     }
 
     private addGroup() {
-        let groups = this.state.namespaceGroups;
+        let groups = [...this.state.namespaceGroups]
         groups.push(this.state.newNamespaceGroup.trim());
         this.setState({
             namespaceGroups: groups,
@@ -265,7 +265,7 @@ export class NamespaceForm extends React.Component<IProps, IState> {
     }
 
     private deleteItem(id) {
-        const groups = this.state.namespaceGroups;
+        let groups = [...this.state.namespaceGroups]
         const index = groups.indexOf(id);
         if (index !== -1) {
             groups.splice(index, 1);
--- a/src/components/namespace-form/namespace-form.tsx
+++ b/src/components/namespace-form/namespace-form.tsx
@@ -103,6 +103,9 @@ export class NamespaceForm extends React.Component<IProps, IState> {
                             type='text'
                             placeholder='Red Hat account ID'
                             value={this.state.newNamespaceGroup}
+                            isValid={
+                                !isNaN(Number(this.state.newNamespaceGroup))
+                            }
                             onChange={value =>
                                 this.setState({
                                     newNamespaceGroup: value,
@@ -254,6 +257,12 @@ export class NamespaceForm extends React.Component<IProps, IState> {
 
     private addGroup() {
         const update = { ...this.props.namespace };
+        if (
+            this.state.newNamespaceGroup.trim() == '' ||
+            isNaN(Number(this.state.newNamespaceGroup.trim()))
+        ) {
+            return;
+        }
         update.groups.push(
             'rh-identity-account:' + this.state.newNamespaceGroup.trim(),
         );
--- a/src/components/namespace-form/namespace-form.tsx
+++ b/src/components/namespace-form/namespace-form.tsx
@@ -89,8 +89,9 @@ export class NamespaceForm extends React.Component<IProps, IState> {
                             <Chip
                                 key={group}
                                 onClick={() => this.deleteItem(group)}
+                                isReadOnly={group === 'system:partner-engineers'}
                             >
-                                {group}
+                                {group.split(':')[1]}
                             </Chip>
                         ))}
                     </ChipGroup>
--- a/src/components/namespace-form/namespace-form.tsx
+++ b/src/components/namespace-form/namespace-form.tsx
@@ -18,7 +18,6 @@ interface IProps {
 interface IState {
     newLinkName: string;
     newLinkURL: string;
-    namespaceGroups: string[];
     newNamespaceGroup: string;
 }
 
@@ -29,14 +28,12 @@ export class NamespaceForm extends React.Component<IProps, IState> {
         this.state = {
             newLinkURL: '',
             newLinkName: '',
-            namespaceGroups: props.namespace.groups,
             newNamespaceGroup: '',
         };
     }
 
     render() {
         const { namespace, errorMessages } = this.props;
-        const { namespaceGroups } = this.state;
 
         if (!namespace) {
             return null;
@@ -65,7 +62,7 @@ export class NamespaceForm extends React.Component<IProps, IState> {
                         >
                             <br />
                             <ChipGroup>
-                                {namespaceGroups.map(group => (
+                                {this.props.namespace.groups.map(group => (
                                     <Chip
                                         key={group}
                                         onClick={() => this.deleteItem(group)}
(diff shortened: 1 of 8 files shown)


romanblanco:
waiting for https://github.com/ansible/galaxy-api/pull/145

src/components/namespace-form/namespace-form.tsx
```diff
@@ -17,6 +18,8 @@ interface IProps {
 interface IState {
     newLinkName: string;
     newLinkURL: string;
+    namespaceGroups: Array<string>;
```
newswangerd: For consistency it would be better to use `string[]` because that's what we've been using for array definitions.

src/components/namespace-form/namespace-form.tsx
```diff
@@ -51,6 +57,52 @@ export class NamespaceForm extends React.Component<IProps, IState> {
 
                         <br />
 
+                        <FormGroup
+                          fieldId='groups'
+                          label='Red Hat Accounts'
```
newswangerd: Red Hat accounts (sentence case)

src/components/namespace-form/namespace-form.tsx
```diff
@@ -51,6 +57,52 @@ export class NamespaceForm extends React.Component<IProps, IState> {
...
+                                        {group}
+                                    </Chip>
+                                ))}
+                            </ChipGroup>
+                            <div className='account-ids'>
+                                <div className='account-id'>
+                                    <TextInput
+                                        id='url'
+                                        type='text'
+                                        placeholder='Red Hat Account ID'
```
newswangerd: Sentence case

src/components/namespace-form/namespace-form.tsx
```diff
@@ -200,6 +252,21 @@ export class NamespaceForm extends React.Component<IProps, IState> {
         );
     }
 
+    private addGroup() {
+      let groups = this.state.namespaceGroups;
+      groups.push("rh-identity-account:" + this.state.newNamespaceGroup.trim());
```
newswangerd: Two things here: 

1. Create a copy of the `this.state.namespaceGroups` property before you modify it (https://reactjs.org/tutorial/tutorial.html#why-immutability-is-important)
2. This doesn't allow the user to add groups that don't start with `rh-identity-account` (such as `system:partner-engineer`)

src/components/namespace-form/namespace-form.tsx
```diff
@@ -200,6 +252,21 @@ export class NamespaceForm extends React.Component<IProps, IState> {
...
+      this.setState({ namespaceGroups: groups, newNamespaceGroup: ''});
+    }
+
+    private deleteItem(id) {
+      const groups = this.state.namespaceGroups;
```
newswangerd: Same as no. 1 above.

src/components/namespace-form/namespace-form.tsx
romanblanco: > This doesn't allow the user to add groups that don't start with `rh-identity-account` (such as `system:partner-engineer`)

Group `system:partner-engineer` is assigned by default. Are we using other types than `rh-identity-account` and `system:partner-engineer`?

src/components/namespace-modal/namespace-modal.tsx
```diff
@@ -0,0 +1,124 @@
+import * as React from 'react';
```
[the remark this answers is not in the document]
romanblanco: @newswangerd I've decided to go with the modal component to save user's time by giving him just the most important fields to fill while creating a new namespaces. All the secondary information can be filled in the mentioned `NamespaceForm`.

src/components/namespace-form/namespace-form.tsx
newswangerd: Oh, in that case it's probably fine to leave it like this for now.

src/components/namespace-modal/namespace-modal.tsx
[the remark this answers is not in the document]
newswangerd: There are a few reasons why this is a bad idea:
1. Duplication: we now have two components that perform the same functionality to maintain
2. Inconsistent UX: this adds different workflows that do the same thing
3. Inconvenience: if the user wants to add more than just the namespace and owners, they need to create an incomplete entry first and then find and fill out the update form.

If you're dead set on having this in a modal, you should at least try to use the existing `NamespaceForm` component inside the modal instead of creating a new form.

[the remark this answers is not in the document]
romanblanco: @sbuenafe-rh

Screencast from 2020-01-23 07-43-17

romanblanco:
Marking this as `[WIP]`, updating namespaces with changes in ansible/galaxy-api#145 does not work.

Still waiting for ansible/galaxy-api#145 to be merged.

src/components/namespace-modal/namespace-modal.tsx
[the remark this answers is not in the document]
newswangerd: We've decided to keep the modal, but put the existing form component in it.

src/components/namespace-form/namespace-form.tsx
newswangerd: One way to create a copy of the array is to use `let groups = [...this.state.namespaceGroups]`.

src/containers/namespace-list/namespace-list.tsx
```diff
@@ -119,6 +151,9 @@ export class NamespaceList extends React.Component<IProps, IState> {
                                 }
                             />
                         </div>
+                        <div className='create-button'>
```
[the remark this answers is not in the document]
newswangerd: @sbuenafe-rh requested that we move the button down to the toolbar, next to the sort button.

Screen Shot 2020-01-27 at 16 16 29-fullpage

src/components/namespace-modal/namespace-modal.tsx
```diff
@@ -0,0 +1,144 @@
...
+                    title='Create a new namespace'
+                    isOpen={this.props.isOpen}
+                    onClose={this.toggleModal}
+                    actions={[
+                        <Button
+                            key='confirm'
+                            variant='primary'
+                            onClick={this.handleSubmit}
+                        >
+                            Confirm
```
newswangerd: Maybe change this to "Create"

src/components/namespace-modal/namespace-modal.tsx
```diff
@@ -0,0 +1,144 @@
...
+}
+
+interface IState {
+    newNamespaceName: string;
+    newNamespaceGroupIds: string;
+    errorMessages: any;
+}
+
+export class NamespaceModal extends React.Component<IProps, IState> {
+    handleSubmit;
```
newswangerd: With TypeScript, It's better not to assign these functions in the constructor. Instead you can

```
    private handleSubmit = event => {
        const ids = this.state.newNamespaceGroupIds.split(',');
        const groups = ids.map(id => id.trim());
        ...
```

I recommend adding `private` to all your function declarations. It makes it easier for us to clean up old functions when they aren't used anymore because linters will know to flag them as unused if they are market as private.

src/components/namespace-modal/namespace-modal.tsx
```diff
@@ -0,0 +1,144 @@
...
+            newNamespaceName: '',
+            newNamespaceGroupIds: '',
+            errorMessages: {},
+        };
+
+        this.handleSubmit = event => {
+            const ids = this.state.newNamespaceGroupIds.split(',');
+            const groups = ids.map(id => id.trim());
+            groups.push('system:partner-engineers');
+            const data: any = {
```
newswangerd: Avoid using `any` as much as possible. You can find most of our type definitions here: https://github.com/ansible/ansible-hub-ui/tree/master/src/api/response-types.

src/components/namespace-modal/namespace-modal.tsx
```diff
@@ -0,0 +1,144 @@
...
+            };
+            NamespaceAPI.createNamespace(data)
+                .then(results => {
+                    this.toggleModal();
+                    this.setState({
+                        newNamespaceName: '',
+                        newNamespaceGroupIds: '',
+                        errorMessages: {},
+                    });
+                    window.location.href = window.location + '/' + data['name'];
```
newswangerd: Use the `react-router` `history` object to update the window location. Here's an example of that: https://github.com/ansible/ansible-hub-ui/blob/master/src/containers/namespace-detail/namespace-detail.tsx#L133. 

The history prop is available to any component that is created by react router, which are all exported using the `withRouter` function like this: https://github.com/ansible/ansible-hub-ui/blob/master/src/containers/namespace-detail/manage-namespace.tsx#L22. In this case you'll need to pass it down to this component.

Additionally, whenever you put a URL in, you need to use the `Paths` enum to define the route and the `formatPath` function to get add any dynamic routes to it. For example if you want a link to the edit namespace form you would do something like this: https://github.com/ansible/ansible-hub-ui/blob/master/src/containers/namespace-detail/namespace-detail.tsx#L289

src/components/namespace-modal/namespace-modal.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+    render() {
+        const {
+            newNamespaceName,
+            newNamespaceGroupIds,
+            errorMessages,
+        } = this.state;
+
+        return (
+            <React.Fragment>
```
newswangerd: Since everything is already in the `Modal` tag, you can probably get rid of `React.Fragment`. React requires that each component only return one child, and `React.Fragment` is just a handy component you can use to wrap multiple child nodes in.

src/containers/namespace-list/namespace-list.tsx
```diff
@@ -45,6 +62,8 @@ interface IProps extends RouteComponentProps {
 
 export class NamespaceList extends React.Component<IProps, IState> {
     nonURLParams = ['tenant'];
+    handleModalToggle;
```
newswangerd: Same as above. You can declare these outside of the constructor.

src/components/namespace-modal/namespace-modal.tsx
newswangerd: Here's a better description for why we use the `Paths` enum: https://github.com/ansible/ansible-hub-ui/blob/master/developer_guidelines.md#urls

src/containers/namespace-list/namespace-list.tsx
```diff
@@ -63,9 +81,19 @@ export class NamespaceList extends React.Component<IProps, IState> {
             itemCount: 0,
             params: params,
             hasPermission: true,
+            isModalOpen: false,
+            partnerEngineer: false,
         };
+
+        this.isPartnerEngineer();
```
newswangerd: Don't make API calls in the constructor. This should go in `componentDidMount`.  API calls in the constructor can cause performance issues and crashes if the state gets updated before the component has been mounted.

src/components/namespace-modal/namespace-modal.tsx
```diff
@@ -0,0 +1,143 @@
...
+    errorMessages: any;
+}
+
+export class NamespaceModal extends React.Component<IProps, IState> {
+    toggleModal;
+
+    constructor(props) {
+        super(props);
+
+        this.toggleModal = this.props.toggleModal;
```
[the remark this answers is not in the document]
romanblanco: I'll be looking into this. This change throws errors:

```
ERROR in [at-loader] ./src/components/namespace-modal/namespace-modal.tsx:47:28
    TS2349: This expression is not callable.
  Type '{}' has no call signatures.

ERROR in [at-loader] ./src/components/namespace-modal/namespace-modal.tsx:80:17
    TS2322: Type 'object' is not assignable to type '() => void'.
  Type '{}' provides no match for the signature '(): void'.

ERROR in [at-loader] ./src/components/namespace-modal/namespace-modal.tsx:92:25
    TS2322: Type 'object' is not assignable to type '(event: MouseEvent<HTMLButtonElement, MouseEvent>) => void'.
  Type '{}' provides no match for the signature '(event: MouseEvent<HTMLButtonElement, MouseEvent>): void'.
```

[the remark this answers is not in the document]
newswangerd: You probably need to pass it as `onClose={() => this.props.toggleModal()}`.

This actually hits on something that I've been thinking about. My linter for Collection Explorer always complains about passing anonymous arrow functions as props. After looking through your PRs I'm starting to think you might have stumbled on the right way to do this by binding all the callbacks to class methods.

I'll leave this one up to you to figure out. If you can come up with a more efficient way to pass callbacks from props, let me know.

Once you get this fixed, feel free to merge it.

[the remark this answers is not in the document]
romanblanco: @newswangerd tried this, still getting `TS2349: This expression is not callable.`:

```diff
@@ -22,12 +23,10 @@ interface IState {
 }

 export class NamespaceModal extends React.Component<IProps, IState> {
-    toggleModal;

     constructor(props) {
         super(props);

-        this.toggleModal = this.props.toggleModal;
         this.state = {
             newNamespaceName: '',
             newNamespaceGroupIds: '',
@@ -45,12 +51,6 @@ export class NamespaceModal extends React.Component<IProps, IState> {
         };
         NamespaceAPI.create(data)
             .then(results => {
-                this.toggleModal();
-                this.setState({
-                    newNamespaceName: '',
-                    newNamespaceGroupIds: '',
-                    errorMessages: {},
-                });
                 this.props.onCreateSuccess(data);
             })
             .catch(error => {
@@ -77,7 +77,7 @@ export class NamespaceModal extends React.Component<IProps, IState> {
                 isLarge
                 title='Create a new namespace'
                 isOpen={this.props.isOpen}
-                onClose={this.toggleModal}
+                onClose={() => this.props.toggleModal()}
                 actions={[
                     <Button
                         key='confirm'
@@ -89,7 +89,7 @@ export class NamespaceModal extends React.Component<IProps, IState> {
                     <Button
                         key='cancel'
                         variant='link'
-                        onClick={this.toggleModal}
+                        onClick={() => this.props.toggleModal()}
                     >
                         Cancel
                     </Button>,
```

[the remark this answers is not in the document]
romanblanco: checked with @karelhala, continue working on this in https://github.com/ansible/ansible-hub-ui/pull/96

PR #89, opened 2020-01-24
Add import log to colletion detail page.

newswangerd:
Fixes: https://github.com/ansible/galaxy-dev/issues/222

Screen Shot 2020-01-24 at 15 19 44-fullpage

### DIFF
--- a/src/Routes.js
+++ b/src/Routes.js
@@ -47,6 +47,13 @@ const CollectionDocs = asyncComponent(() =>
     ),
 );
 
+const CollectionImportLog = asyncComponent(() =>
+    import(
+        /* webpackChunkName: "collection_detail" */
+        './containers/collection-detail/collection-import-log'
+    ),
+);
+
 const NotFound = asyncComponent(() =>
     import(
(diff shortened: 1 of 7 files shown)


romanblanco:
@newswangerd displaying import log on collection detail page adds the same information as we already have on 'My imports' page.

Do we plan to keep using both of them?

It feels easier to navigate to the import log on collection detail page, but on 'My imports' page it's possible to switch collection version:

Screencast from 2020-02-03 19-30-41



(`/automation-hub/my-imports?namespace=rblanco`)
Screenshot_2020-02-03 Ansible Automation Hub

(`/automation-hub/rblanco/collection_demo/import-log`)
Screenshot_2020-02-03 Ansible Automation Hub(1)

/cc @chouseknecht

newswangerd:
The plan is to keep both of them. My imports is designed for collection maintainers and it lets them see an overview of all their imports as well as view imports as their happening, which is functionality that isn't in collection details. 

Also, failed imports won't be viewable on collection details because the data for the collection detail page isn't available until after the import succeeds and is approved by.

PR #91, opened 2020-02-03
2 spaces for a tab in prettier

romanblanco:
Even though there does not seem to be a set standard for React/TypeScript code, I would still like to suggest to take back changes done in commit 39e0f8369690331522e4917018805c34c19751f9 and use two spaces instead of four to indent our TypeScript code.

Using two spaces seems to be the most common style, and also helps to read code as the longer lines are breaked in less cases. Especially reading the code on vertical monitor would be much easier.

### DIFF
--- a/.prettierrc.yaml
+++ b/.prettierrc.yaml
@@ -1,6 +1,7 @@
 printWidth: 80
 singleQuote: true
 trailingComma: 'all'
-tabWidth: 4
+tabWidth: 2
 jsxSingleQuote: true
 jsxBracketSameLine: false
+bracketSpacing: true
--- a/src/api/base.ts
+++ b/src/api/base.ts
@@ -3,71 +3,69 @@ import { Constants } from '../constants';
 import { ParamHelper } from '../utilities';
 
 export class BaseAPI {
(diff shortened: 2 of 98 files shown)


bmclaughlin:
I suspected the 4 spaces were related to us being a Python shop.  I'm all for better readability and ease of use with React. 

From the limited searches I've done, React looks to be 2 spaces and Typescript looks to lean towards 4 but don't have a hard recommendation.

romanblanco:
I'd also suggest to configure prettier to prefer 

```
if (expression) {
  expression;
} else {
  alt_expression;
}
```

rather than

```
if (expression) {
  expression;
} else alt_expression;
```

my suggestion is give +1/-1 if agree/disagree.

newswangerd:
> I'd also suggest to configure prettier to prefer
> 
> ```
> if (expression) {
>   expression;
> } else {
>   alt_expression;
> }
> ```
> 
> rather than
> 
> ```
> if (expression) {
>   expression;
> } else alt_expression;
> ```
> 
> my suggestion is give +1/-1 if agree/disagree.

That's fine with me. I didn't know that was an option.

PR #92, opened 2020-02-03
Allow the UI to run without insights.

newswangerd:
This PR changes the way that the app is compiled to make it possible to easily configure the app to be compiled for different environments.

This provides a base webpack config at `config/webpack.base.config.js` that can be imported into a new configuration and initialized with set of customizable settings to determine how the app should compile. This is an example of how that looks like:

```
const webpackBase = require('./config/webpack.base.config');

module.exports = webpackBase({
    // The host where the API lives. EX: https://localhost:5001
    API_HOST: '',

    // Path to the API on the API host. EX: /api/automation-hub
    API_BASE_PATH: '/api/automation-hub/',

    // Path on the host where the UI is found. EX: /apps/automation-hub
    UI_BASE_PATH: '',

    // Determines if the app should be compiled to run on insights or on
    // another platform. Options: insights, standalone
    DEPLOYMENT_MODE: 'insights',

    // Serve the UI over http or https. Options: true, false
    UI_USE_HTTPS: true,

    // Enables webpack debug mode. Options: true, false
    UI_DEBUG: true,

    // Target compilation environment. Options: dev, prod
    TARGET_ENVIRONMENT: 'dev',
});
```

Once a new configuration file is created it can be used to build the app the same way as any other webpack app, by running `webpack --config <path to new config here>`.

The rationale behind these changed is that this allows users to create a simple configuration file that can be used as a template in an Ansible playbook, Dockerfile or any other CI system to configure the UI for any user's environment.


cutwater:
@newswangerd pluggable configuration with webpack seems reasonable to me.

romanblanco:
To run the standalone UI on a different port, I've changed:

```ack
config/webpack.base.config.js
45:        port: 8013,

config/spandx.config.js
12:        '/apps/automation-hub': { host: `https://${localhost}:8013` },
13:        '/ansible/automation-hub': { host: `https://${localhost}:8013` },

profiles/local-frontend.js
5:const FRONTEND_PORT = 8013;

profiles/local-frontend-and-api.js
5:const FRONTEND_PORT = 8013;
```

Running `npm run start` still tries to run on port 8002:

```
Error: listen EADDRINUSE: address already in use 127.0.0.1:8002
``` 

Checking whether it is problem in the PR or in my setup

newswangerd:
@romanblanco making the UI port configurable is a good idea. I need to look into whether or not this is possible in the webpack configs that insights provides us. This might be a good project for you.

newswangerd:
Another potential feature we could consider (for another PR) is creating a build container for the UI.

romanblanco:
> Another potential feature we could consider (for another PR) is creating a build container for the UI.

seems like @alikins is already working on that: https://github.com/ansible/galaxy-dev/compare/master...alikins:docker_ui

romanblanco:
> Another potential feature we could consider (for another PR) is creating a build container for the UI.

seems like @alikins is already working on that: https://github.com/ansible/galaxy-dev/compare/master...alikins:docker_ui

PR #97, opened 2020-02-05
Fix UX nitpicks on Collections page

newswangerd:
Screen Shot 2020-02-05 at 16 44 52-fullpage
Screen Shot 2020-02-05 at 16 44 39-fullpage
Screen Shot 2020-02-05 at 16 44 34-fullpage

### DIFF
--- a/src/components/collection-list/collection-list-item.tsx
+++ b/src/components/collection-list/collection-list-item.tsx
@@ -79,11 +79,6 @@ export class CollectionListItem extends React.Component<IProps, {}> {
                 <div className='entry'>
                     {latest_version.metadata.description}
                 </div>
-                <div className='entry pf-l-flex pf-m-wrap'>
-                    {latest_version.metadata.tags.map((tag, index) => (
-                        <Tag key={index}>{tag}</Tag>
-                    ))}
-                </div>
                 <div className='entry pf-l-flex pf-m-wrap content'>
                     {Object.keys(contentSummary.contents).map(k => (
                         <div key={k}>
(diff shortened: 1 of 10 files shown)


romanblanco:
@newswangerd to be consistent with b68a37073bc5f17dcc73395fb347f530af4e5d99, could we adjust the empty state for search on Partners and Namespaces page as well?

The changes in b68a37073bc5f17dcc73395fb347f530af4e5d99 look good!

Screenshot_2020-02-07 Ansible Automation Hub(1)

Screenshot_2020-02-07 Ansible Automation Hub

romanblanco:
Maybe to consider: at the empty state, we could give the user a hint on what would be the next action (see https://github.com/ManageIQ/manageiq-ui-classic/pull/5434#issue-267795874).

@sbuenafe-rh could know, if that is something we want to do.

sbuenafe-rh:
@newswangerd Collections pages look good. I would want to make sure that there's 24px between the card/list view buttons and the compact pagination component. 

Also, for the filter input field, are we filtering by the name of the collection? If so, I would change the placeholder text to be "Filter by collection".

sbuenafe-rh:
@romanblanco the Partners page, no results found/matched, the message should be the same as the no collections found. See mock for reference: https://marvelapp.com/project/4271938/screen/65612952

sbuenafe-rh:
@romanblanco also, for the My namespaces empty page (Day 1 user), the toolbar should be removed and the action button to create a namespace should be in the empty state message. See mock for reference: https://marvelapp.com/project/4271938/screen/65785967

sbuenafe-rh:
For a PF Empty state component reference, the pages that we've talked about to should be using the **Empty State - Large**. See PF documentation: https://www.patternfly.org/v4/documentation/core/components/emptystate#large

newswangerd:
@sbuenafe-rh 

> Also, for the filter input field, are we filtering by the name of the collection? If so, I would change the placeholder text to be "Filter by collection".

I checked. It performs a search, not a filter.

sbuenafe-rh:
> @sbuenafe-rh
> 
> > Also, for the filter input field, are we filtering by the name of the collection? If so, I would change the placeholder text to be "Filter by collection".
> 
> I checked. It performs a search, not a filter.

Ok knowing this, the no matches found message will probably need to be tweaked.

PR #102, opened 2020-02-18
Prevent user to enter namespace editation after he disowns it

romanblanco:
This PR corrects the situation where error 403 is received after user attempts to save change after loosing access. After the action, user is redirected back on home screen.

The PR also fixes the situation where the user tries to open edit page even though he is not an owner.

Links
----------------

* https://github.com/ansible/galaxy-dev/issues/253#issue-561867562

Steps for Testing/QA

Prevent to enter the edit page:
1. navigate to namespace edit page (`https://ci.foo.redhat.com:1337/beta/ansible/automation-hub/my-namespaces/edit/<namespace_name>`)
2. remove the user as owner
3. try to access the edit page again

Forward away on attempt to save after loosing access:
1. Open edit namespace page logged as User1
2. Remove User1 as namespace owner
3. Try to save changes

### DIFF
--- a/src/containers/edit-namespace/namespace-form.tsx
+++ b/src/containers/edit-namespace/namespace-form.tsx
@@ -221,6 +221,7 @@ class EditNamespace extends React.Component<RouteComponentProps, IState> {
                                     `Not possible to proceed with the action.` +
                                     `Plese contact the administrator of the namespace`,
                             }),
+                            saving: false,
                         });
                     }
                 });
--- a/src/containers/edit-namespace/namespace-form.tsx
+++ b/src/containers/edit-namespace/namespace-form.tsx
@@ -217,9 +217,7 @@ class EditNamespace extends React.Component<RouteComponentProps, IState> {
                             alerts: this.state.alerts.concat({
                                 variant: 'danger',
                                 title: `API Error: ${error.response.status}`,
-                                description:
-                                    `Not possible to proceed with the action.` +
-                                    `Plese contact the administrator of the namespace`,
+                                description: `You don't have permissions to update this namespace.`,
                             }),
                             saving: false,
                         });
--- a/src/containers/edit-namespace/namespace-form.tsx
+++ b/src/containers/edit-namespace/namespace-form.tsx
@@ -203,6 +203,12 @@ class EditNamespace extends React.Component<RouteComponentProps, IState> {
                             errorMessages: messages,
                             saving: false,
                         });
+                    } else if (result.status === 403) {
+                        this.setState({
+                            redirect: formatPath(Paths.myNamespaces, {
+                                namespace: '',
+                            }),
+                        });
                     }
                 });
         });
--- a/src/containers/edit-namespace/namespace-form.tsx
+++ b/src/containers/edit-namespace/namespace-form.tsx
@@ -7,6 +7,9 @@ import {
     PartnerHeader,
     NamespaceForm,
     ResourcesForm,
+    AlertList,
+    closeAlertMixin,
+    AlertType,
     Main,
 } from '../../components';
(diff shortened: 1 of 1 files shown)


newswangerd:
Closing because we solved this in #105

romanblanco:
@newswangerd I think we should still use the part that prevents user to get stuck on edit page, in case he looses permissions while editing:

https://github.com/ansible/ansible-hub-ui/pull/102/files#diff-cb99783ac2e06b5caa0da844d4149302R217-R222
```

                    } else if (result.status === 403) {
                        this.setState({
                            redirect: formatPath(Paths.myNamespaces, {
                                namespace: '',
                            }),
                        });
```

I suggest to reopen this.

newswangerd:
oh, yeah, that's a good point. I would suggest adding an error message saying that the user doesn't have access to edit the page instead of redirecting them without any explanation

src/containers/edit-namespace/namespace-form.tsx
```diff
@@ -54,7 +54,21 @@ class EditNamespace extends React.Component<RouteComponentProps, IState> {
     }
 
     componentDidMount() {
-        this.loadNamespace();
+        const editedNamespace = this.props.match.params['namespace'];
+        NamespaceAPI.update(editedNamespace, { name: editedNamespace })
```
newswangerd: We don't need this part anymore.

src/containers/edit-namespace/namespace-form.tsx
```diff
@@ -200,6 +214,12 @@ class EditNamespace extends React.Component<RouteComponentProps, IState> {
                             errorMessages: messages,
                             saving: false,
                         });
+                    } else if (result.status === 403) {
```
newswangerd: Instead of redirecting if it returns a 403, just show the user an error message saying they don't have permission to edit the form

src/containers/edit-namespace/namespace-form.tsx
```diff
@@ -203,10 +212,23 @@ class EditNamespace extends React.Component<RouteComponentProps, IState> {
                             errorMessages: messages,
                             saving: false,
                         });
+                    } else if (result.status === 404) {
+                        this.setState({
+                            alerts: this.state.alerts.concat({
+                                variant: 'danger',
+                                title: `API Error: ${error.response.status}`,
+                                description:
+                                    `Not possible to proceed with the action.` +
```
newswangerd: Lets change the error message to say: 'You don't have permissions to update this namespace.'

src/containers/edit-namespace/namespace-form.tsx
```diff
@@ -203,10 +212,23 @@ class EditNamespace extends React.Component<RouteComponentProps, IState> {
                             errorMessages: messages,
                             saving: false,
                         });
+                    } else if (result.status === 404) {
+                        this.setState({
+                            alerts: this.state.alerts.concat({
```
newswangerd: Don't forget to set loading to false when the result is returned from the API, even if the result is bad

src/containers/edit-namespace/namespace-form.tsx
romanblanco: @newswangerd 

I'm not sure if mentioning permissions is correct approach in the situation where we return `404 Not Found` instead of `403 Forbidden`.
The message stating the user does not have permission might be misleading.

Also, in case of malicious request, I'm concerned this message gives away the information that the namespace actually exists, even though the user is not permitted to see it.

Would you still prefer to update the message to _"You don't have permissions to update this namespace."_?

newswangerd: the reason the API returns a 404 is because the my namespace API is filtered to only show namespaces the user owns, so in this case I think the message about not having permissions is the most accurate.

PR #104, opened 2020-02-19
Use constants for admin user group.

newswangerd:
Adds `ADMIN_GROUP` constant that's set to `system:partner-engineers` to make it a little easier to update the system if we ever need to change the name name of the partner engineer group.

### DIFF
--- a/src/containers/edit-namespace/namespace-form.tsx
+++ b/src/containers/edit-namespace/namespace-form.tsx
@@ -10,6 +10,7 @@ import {
     Main,
 } from '../../components';
 import { MyNamespaceAPI, NamespaceType } from '../../api';
+import { Constants } from '../../constants';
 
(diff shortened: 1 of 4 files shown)


src/containers/edit-namespace/namespace-form.tsx
```diff
@@ -149,7 +150,7 @@ class EditNamespace extends React.Component<RouteComponentProps, IState> {
     }
 
     private removeGroupsPrefix(groups) {
-        const partnerEngineerOwner = 'system:partner-engineers';
+        const partnerEngineerOwner = Constants.ADMIN_GROUP;
```
romanblanco: using the `Constants.ADMIN_GROUP`, the `partnerEngineerOwner` variable is not necessary

newswangerd: oh, good point

PR #105, opened 2020-02-19
Return 404 when users try to access namespace for on namespaces they don't own

newswangerd:
Use the `my-namespaces` endpoint for the edit namespace form. This automatically filters out namespaces that the user doesn't have access to.


romanblanco:
Fixes: ansible/galaxy-dev#253

PR #120, opened 2020-04-20
Add option for configuring webpack public path.

newswangerd:


config/standalone.prod.webpack.config.js
```diff
@@ -9,4 +9,5 @@ module.exports = webpackBase({
   UI_USE_HTTPS: false,
   UI_DEBUG: false,
   TARGET_ENVIRONMENT: 'prod',
+  WEBPACK_PUBLIC_PATH: '/static/galaxy_ng/',
```
fabricio-aguiar, line 12: Could you set it on `config/standalone.dev.webpack.config.js`?

newswangerd: That's a good idea, I can do that.

PR #116, opened 2020-04-03
Move collection filtering by tags into toolbar

romanblanco:
These changes are part of adjustments related to content sync feature.
The goal of this PR is to move filtering by tags from sidebar into a toolbar together with filtering by name and sync status.

TODO
----------------
- [x] params properties are not restored from URL for tags and sync status
- [x] styling Chip group
- [ ] filtering for Sync can not be applied yet (waiting for https://github.com/ansible/galaxy_ng/issues/45, https://github.com/ansible/galaxy_ng/pull/68)
- [ ] test that the filtering is applied correctly

Links
----------------

* story: https://github.com/ansible/galaxy_ng/issues/17
* subtask https://github.com/ansible/galaxy_ng/issues/19
* [ux mockup](https://marvelapp.com/4dijjai/screen/66793821)

Steps for Testing/QA
-------------------------------

- Go to _Collections_ page in Automation Hub UI
- Toolbar at the top contains the options

PR #130, opened 2020-05-05
Search tags compound filter

newswangerd:
Updated version of  #116


src/containers/search/search.tsx
```diff
@@ -70,14 +71,6 @@ class Search extends React.Component<RouteComponentProps, IState> {
       params['page_size'] = Constants.CARD_DEFAULT_PAGE_SIZE;
     }
 
-    if (!params['tag']) {
-      params['tag'] = [];
-    }
-
-    if (!params['sync']) {
-      params['sync'] = ['Sync on'];
```
romanblanco: @newswangerd the purpose of this line was to have `'Sync on'` as a default option: https://github.com/ansible/ansible-hub-ui/pull/116#issuecomment-608508671

newswangerd: I figured that's what it's for. I'm removing it for the moment because I want to merge this PR with just the new tag filter and then add the sync param in later when we have an API for it so that we can test the new multiselect filter without breaking the collection page.

PR #131, opened 2020-05-11
Add login page to on prem

newswangerd:
Adds:
- Login page
- Masthead with user's  name and username
- Authentication checks on every page

Screen Shot 2020-05-11 at 13 39 19-fullpage
Screen Shot 2020-05-11 at 13 38 17-fullpage

https://github.com/ansible/galaxy_ng/issues/75
https://github.com/ansible/galaxy_ng/issues/16


src/api/response-types/user.ts
```diff
@@ -17,6 +17,16 @@ export class UserAuthType {
...
 }
 
+export class UserAuthType {
+  id: number;
+  username: string;
+  first_name?: string;
+  last_name?: string;
+  email?: string;
+  groups: { id: number; name: string }[];
+  is_partner_engineer: boolean;
```
ironfroggy, line 27: Do on-premise instances _have_ "partner engineer" users?

src/containers/login/login.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+  private onLoginButtonClick = event => {
+    ActiveUserAPI.login(this.state.usernameValue, this.state.passwordValue)
+      .then(result => {
+        ActiveUserAPI.getUser().then(() =>
+          this.setState({ redirect: this.redirectPage }),
+        );
+      })
+      .catch(result =>
+        this.setState({ passwordValue: '', showHelperText: true }),
```
ironfroggy, line 87: Is there any other kind of error that could come back than an incorrect password?

Invalid user?
Deactivated account?
500 Internal server error?

src/api/response-types/user.ts
newswangerd: This is for backwards compatibility. The API in cloud.redhat.com returns an `is_partner_engineer` flag and doesn't have a `groups` attribute yet. We'll probably rename this to `admin` or something like that once we can do so without messing up cloud.

src/containers/login/login.tsx
newswangerd: Internal error is a good idea, I'll add that.

My thought for not returning an error like `invalid user` or `deactivated account` is that this gives a potential attacker less information about what user accounts are available to break into. I don't want an attacker to be able to determine if there is an `admin` or `root` user just by putting in the username with a blank user.

newswangerd: Though now that I think about it, an attacker could just use the api...

PR #140, opened 2020-05-22
Add SortTable component

ZitaNemeckova:
TODO: 
- [x] correct icons
- [x] save state
- [x] remove old select for that page
- [x] add active CSS 

Fixes https://github.com/ansible/galaxy_ng/issues/168

Before:
Screenshot 2020-05-22 at 15 19 11

After:
Screenshot 2020-05-29 at 12 18 54

### DIFF
--- a/src/components/sort-table/sort-table.tsx
+++ b/src/components/sort-table/sort-table.tsx
@@ -21,18 +21,15 @@ interface IProps {
 }
 
 export class SortTable extends React.Component<IProps> {
-  private sort(type, id) {
-    let isDescending = this.props.params['sort'].includes('-');
+  private sort(id, isMinus) {
     // Alphabetical sorting is inverted in Django, so flip it here to make
     // things match up with the UI.
-    if (type == 'alpha') {
-      isDescending = !isDescending;
-    }
+    isMinus = !isMinus;
     this.props.updateParams(
       ParamHelper.setParam(
         this.props.params,
         'sort',
-        (isDescending ? '' : '-') + id,
+        (isMinus ? '-' : '') + id,
       ),
     );
   }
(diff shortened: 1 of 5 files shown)


src/components/sort-table/sort-table.tsx
```diff
@@ -0,0 +1,70 @@
...
+import {
+  LongArrowAltUpIcon,
+  LongArrowAltDownIcon,
+  ArrowsAltVIcon,
+} from '@patternfly/react-icons';
+import { ParamHelper } from '../../utilities';
+import './sort-table.scss';
+
+interface IProps {
+  options: object;
```
newswangerd, line 12: Please add a type for options. Something like
```
{
title: string;
type: string;
id: string;
}[]
```

should do.

src/components/sort-table/sort-table.tsx
```diff
@@ -0,0 +1,70 @@
...
+  private sort(type, id) {
+    this.props.updateParams(
+      ParamHelper.setParam(
+        this.props.params,
+        'sort',
+        (type.includes('up') ? '-' : '') + id,
+      ),
+    );
+  }
+  private getIcon(type, id) {
```
newswangerd, line 27: The state of this component should be derived from the `params` inside of the component instead of setting `type` by checking params in the parent. In this case you can just compare `item.id` to `params['sort']`
- if sort isn't set or doesn't match, return `ArrowsAltVIcon`
- if sort does match, return `LongArrowAltDownIcon` or `LongArrowAltUpIcon` depending on if it starts with `-` or not.

src/containers/certification-dashboard/certification-dashboard.tsx
```diff
@@ -248,22 +241,58 @@ class CertificationDashboard extends React.Component<
         </EmptyState>
       );
     }
+    let sortTableOptions = {
+      headers: [
+        {
+          title: 'Namespace',
+          type: 'none',
```
newswangerd, line 248: Type should be used to declare whether the field is using alphabetical or numeric sort. In the `Sort` component this matters because we change the icon depending on the field type, but I think this is also relevant here because of how of some differences with how Django interprets the `-` between numbers and strings

https://github.com/ansible/ansible-hub-ui/blob/master/src/components/patternfly-wrappers/sort.tsx#L63

ZitaNemeckova:
@newswangerd Fixed. Storybook added. Please have a look, thanks :)

src/containers/certification-dashboard/certification-dashboard.tsx
```diff
@@ -248,22 +241,53 @@ class CertificationDashboard extends React.Component<
...
+        },
+        {
+          title: 'Version',
+          type: 'number',
+          id: 'version',
+        },
+        {
+          title: 'Date created',
+          type: 'number',
+          id: 'date',
```
newswangerd, line 264: The id for date is `pulp_created`

ZitaNemeckova: Fixed.

PR #155, opened 2020-06-26
Replace Sort with SortTable in Users page

ZitaNemeckova:
Before:
Screenshot 2020-06-26 at 16 11 02

After:
Screenshot 2020-06-29 at 14 41 49

Added sort for `Created`.

Note: `api/automation-hub/v3/_ui/users/?sort=-last_name&offset=0&limit=10` returns `400`


ZitaNemeckova:
@newswangerd Please have a look, thanks :)

PR #163, opened 2020-07-14
Update my namespaces to work with inbound repos

ZitaNemeckova:
Fixes https://github.com/ansible/galaxy_ng/issues/290

Steps to test:
- if https://github.com/ansible/galaxy_ng/pull/272 not merged create an inbound repo `inbound-[namespace.name]` 
```python
>>> from pulp_ansible.app.models import AnsibleRepository, AnsibleDistribution
>>> repo = AnsibleRepository.objects.create(name='inbound-[namespace.name]')
>>> AnsibleDistribution.objects.create(name='inbound-[namespace.name]', base_path='inbound-[namespace.name]', repository=repo)
<AnsibleDistribution: automation-hub>
>>> # Press <CTRL+D> to exit.
```

   else create a namespace
- upload collection
- `/content/inbound-[namespace.name]/v3/artifacts/collection` is called instead of `automation-hub/v3/artifacts/collection`

*Add a new view where the user can copy the repository URL for the purpose of configuring the ansible-galaxy cli:*

This shows only in My Namespaces.

Screenshot 2020-08-03 at 16 33 35

### DIFF
--- a/src/containers/namespace-detail/namespace-detail.tsx
+++ b/src/containers/namespace-detail/namespace-detail.tsx
@@ -7,6 +7,7 @@ import {
   DropdownItem,
   Alert,
   AlertActionCloseButton,
+  ClipboardCopy,
 } from '@patternfly/react-core';
 
 import * as ReactMarkdown from 'react-markdown';
@@ -111,7 +112,7 @@ export class NamespaceDetail extends React.Component<IProps, IState> {
       return <LoadingPageWithHeader></LoadingPageWithHeader>;
     }
 
-    const tabs = ['Collections', 'Configuration'];
+    const tabs = ['Collections', 'CLI Configuration'];
 
     const tab = params['tab'] || 'collections';
 
@@ -188,8 +189,14 @@ export class NamespaceDetail extends React.Component<IProps, IState> {
             ) : (
               this.renderResources(namespace)
             )}
-            {tab.toLowerCase() === 'configuration' ? (
-              <div>{repositoryUrl}</div>
+            {tab.toLowerCase() === 'cli configuration' ? (
+              <div>
+                <ClipboardCopy isReadOnly>{repositoryUrl}</ClipboardCopy>
+                <div>
+                  <b>Note:</b> Use this URL to configure ansible-galaxy to
+                  upload collections to.
+                </div>
+              </div>
             ) : (
               this.renderResources(namespace)
             )}
--- a/src/containers/namespace-detail/namespace-detail.tsx
+++ b/src/containers/namespace-detail/namespace-detail.tsx
@@ -7,6 +7,11 @@ import {
   DropdownItem,
   Alert,
   AlertActionCloseButton,
+  ButtonVariant,
+  TextInput,
+  Form,
+  FormGroup,
+  ActionGroup,
 } from '@patternfly/react-core';
 
 import * as ReactMarkdown from 'react-markdown';
@@ -111,13 +116,15 @@ export class NamespaceDetail extends React.Component<IProps, IState> {
       return <LoadingPageWithHeader></LoadingPageWithHeader>;
     }
 
-    const tabs = ['Collections'];
+    const tabs = ['Collections', 'Configuration'];
 
     const tab = params['tab'] || 'collections';
 
     if (namespace.resources) {
       tabs.push('Resources');
     }
+
+    const repositoryUrl = 'inbound-' + namespace.name;
 
     return (
       <React.Fragment>
@@ -179,6 +186,11 @@ export class NamespaceDetail extends React.Component<IProps, IState> {
                   this.handleCollectionAction(id, action)
                 }
               />
+            ) : (
+              this.renderResources(namespace)
+            )}
+            {tab.toLowerCase() === 'configuration' ? (
+              <div>{repositoryUrl}</div>
             ) : (
               this.renderResources(namespace)
             )}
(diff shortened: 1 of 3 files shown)


src/containers/namespace-detail/namespace-detail.tsx
```diff
@@ -111,14 +111,20 @@ export class NamespaceDetail extends React.Component<IProps, IState> {
...
 
     const tab = params['tab'] || 'collections';
 
     if (namespace.resources) {
       tabs.push('Resources');
     }
 
+    const repositoryUrl =
+      window.location.href.split('/ui')[0].replace('8002', '5001') +
+      '/api/automation-hub/content/inbound-' +
```
newswangerd, line 124: Please use the API base path for the URL because this can change. I think you can get it with `CollectionAPI.apiBaseUrl`

https://github.com/ansible/ansible-hub-ui/blob/stable/src/api/base.ts#L7

ZitaNemeckova: @newswangerd Did you mean to replace `/api/automation-hub/` by `CollectionAPI.apiBaseUrl`? How can I check that user owns the namespace? Or is it enough that the user belongs to same group as a namespace?

newswangerd: `CollectionAPI.apiBaseUrl` should work.

You can just hide it if `showControls` is false. https://github.com/ansible/ansible-hub-ui/blob/master/src/containers/namespace-detail/namespace-detail.tsx#L55

ZitaNemeckova: @newswangerd 
Hiding done in
https://github.com/ansible/ansible-hub-ui/pull/163/files#diff-155af35dcff67011de5d6c80469d78d3R117 
and `CollectionAPI.apiBaseUrl` used.

src/containers/namespace-detail/namespace-detail.tsx
```diff
@@ -113,12 +114,22 @@ export class NamespaceDetail extends React.Component<IProps, IState> {
...
+      tabs.push('CLI Configuration');
+    }
     const tab = params['tab'] || 'collections';
 
     if (namespace.resources) {
       tabs.push('Resources');
     }
 
+    const repositoryUrl =
+      window.location.href.split('/ui')[0].replace('8002', '5001') +
```
newswangerd, line 127: Don't split on `/ui`. The UI base path is configurable and cloud.redhat.com doesn't use that prefix for the UI.

I recommend using `window.location.origin`

newswangerd, line 127: No need to replace the port. On the dev environment `8002` should redirect to `5001`.

src/containers/namespace-detail/namespace-detail.tsx
```diff
@@ -113,12 +114,22 @@ export class NamespaceDetail extends React.Component<IProps, IState> {
...
+      CollectionAPI.apiBaseURL +
```
newswangerd, line 128: On second thought, `CollectionAPI.apiBaseURL` might be a bad choice. We have the option to configure the UI to look for an api on a different host entirely, that's what `API_HOST` does here:

```
export class BaseAPI {
  apiBaseURL = API_HOST + API_BASE_PATH;
  apiPath: string;
  http: any;
```

`API_HOST` and `API_BASE_PATH` are both webpack constants that are defined here https://github.com/ansible/ansible-hub-ui/blob/master/config/standalone.dev.webpack.config.js#L8 if anyone decides to configure an `API_HOST` the URLs are going to break.

I would recommend is to check if `API_HOST` exists. If it does, use `API_HOST + API_BASE_PATH` for the first part of the URL. If it doesn't, use `window.location.origin + API_BASE_PATH`.

src/containers/namespace-detail/namespace-detail.tsx
```diff
@@ -182,6 +193,17 @@ export class NamespaceDetail extends React.Component<IProps, IState> {
             ) : (
               this.renderResources(namespace)
             )}
+            {tab.toLowerCase() === 'cli configuration' ? (
+              <div>
+                <ClipboardCopy isReadOnly>{repositoryUrl}</ClipboardCopy>
+                <div>
+                  <b>Note:</b> Use this URL to configure ansible-galaxy to
```
newswangerd, line 200: Reword this to "Use this URL to configure `ansible-galaxy` to upload collections to this namespace. More information on `ansible-galaxy` configurations can be found [here](https://docs.ansible.com/ansible/latest/galaxy/user_guide.html#configuring-the-ansible-galaxy-client)"

ZitaNemeckova: Done :)

src/containers/namespace-detail/namespace-detail.tsx
ZitaNemeckova: Done :)

src/containers/namespace-detail/namespace-detail.tsx
```diff
@@ -113,12 +114,23 @@ export class NamespaceDetail extends React.Component<IProps, IState> {
...
       tabs.push('Resources');
     }
 
+    const repositoryUrl = !!API_HOST
+      ? API_HOST
+      : window.location.origin +
+        API_BASE_PATH +
+        'content/inbound-' +
+        namespace.name +
+        '/v3/collections/';
```
newswangerd, line 132: I think the client expects the URL to be the endpoint that returns this

```
{
    "available_versions": {
        "v3": "v3/"
    },
    "distro_base_path": "automation-hub"
}
```

So you can drop `v3/collections`. Make sure to leave the`/` at the end though.

ZitaNemeckova: Fixed :)

PR #167, opened 2020-07-29
Check password confirm after Save

ZitaNemeckova:
https://github.com/ansible/galaxy_ng/issues/179#issuecomment-663170597

I'm not sure it should be fixed given [PF guidelines](https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/forms#errors-and-validation) that says that `When a form field submission results in an error, let users know as soon as possible.`


ZitaNemeckova:
This goes against PF guidelines. Closing

PR #173, opened 2020-08-10
[WIP] Create upload input component

ZitaNemeckova:
https://github.com/ansible/ansible-hub-ui/pull/164 needs upload input so creating one instead of copy-pasting.


newswangerd:
I'm actually not sure if reusing this particular component is a good idea. This upload component is one that we brought over from galaxy, and the UX team has already complained about it and told us to change it to match their standards. We might be better off using a different component for the requirements.yml upload

ZitaNemeckova:
@newswangerd Ok, I'll use `FileUploadField` from PF instead. Thanks :)

PR #187, opened 2020-09-07
Add MultiRepo select to navbar

ZitaNemeckova:
Related to https://github.com/ansible/galaxy_ng/issues/296

TODO: 
- [x] Add select to navbar
- [x] Fix select styling
- [x] Reload page when selectedRepo changes
- [x] Check `:repo` param and if needed change `this.context.selectedRepo` or 404 
- [x] Add new routes
- [x] New routes and check for namespace detail
- [ ] Breadcrumbs
- [x] Remove it from Insights
- [x] Use API according to selected repo
  - [x] Collection list by repo: `api/automation-hub/v1/_ui/repo/<repo_name>/`
  - [x] Collection detail by repo: `api/automation-hub/v1/_ui/repo/<repo_name>/<namespace_name>/<collection_name>/`
  - [x] Collection list by repo, filter by namespace: `api/automation-hub/v1/_ui/repo/<repo_name>/?namespace=<namespace_name>`
 - [x] Test Insights!

Navbar:
Screenshot 2020-09-07 at 15 56 22

Note: Dear future self if you ever need it here's commit with propagated props f1632b6 . It worked.


src/api/collection.ts
```diff
@@ -19,6 +21,22 @@ export class API extends BaseAPI {
...
 
+  list(params?: {}, repo?: string) {
+    const path =
+      DEPLOYMENT_MODE === Constants.INSIGHTS_DEPLOYMENT_MODE
+        ? this.apiPath
+        : '_ui/v1/repo/' + repo + '/';
+    return this.http.get(path, params);
+  }
+
+  listByNamespace(namespace: string, params?: {}, repo?: string) {
```
newswangerd, line 32: what is listByNamespace for? Can't we just use list with the `?namespace=` param?

src/api/collection.ts
```diff
@@ -19,6 +21,22 @@ export class API extends BaseAPI {
     // new MockCollection(this.http, this.apiPath);
   }
 
+  list(params?: {}, repo?: string) {
+    const path =
+      DEPLOYMENT_MODE === Constants.INSIGHTS_DEPLOYMENT_MODE
```
newswangerd, line 26: I noticed you're calling `INSIGHTS_DEPLOYMENT_MODE` a lot to toggle between the `collections/` and `repo/<reponame>/` apis. My suggestion is to exclusively use the new `repo/` api endpoint. We want to get rid of the old `collections` anyway. 

Instead of checking the depolyment mode and toggling between the two endpoints, just set `appContext.selectedRepo = 'automation-hub'` for insights mode and don't add any logic to change it. That way, when the app is in insights mode it will always just look up `repo/automation-hub/` which is what `collections/` does anyway.

src/components/collection-list/collection-list-item.tsx
```diff
@@ -54,17 +57,22 @@ export class CollectionListItem extends React.Component<IProps, {}> {
 
     const contentSummary = convertContentSummaryCounts(latest_version.contents);
 
+    const path =
+      DEPLOYMENT_MODE === Constants.INSIGHTS_DEPLOYMENT_MODE
```
newswangerd, line 61: I'm tempted to say we should just use the new `repo/` paths for all our links in insights mode. As long as the old URLs still work for people to link to stuff in hub, it doesn't matter that all the internal links prefix everything with `repo/`

src/constants.tsx
```diff
@@ -70,4 +70,12 @@ export class Constants {
...
     },
   ];
+  static REPOSITORYNAMES = {
+    'Red Hat Certified': 'rh-certified',
+    Community: 'community',
+    Published: 'published',
+  };
+
+  static ALLOWEDREPOS = ['automation-hub', 'community', 'published'];
+  static DEAFAULTREPO = 'Red Hat Certified';
```
newswangerd, line 80: make the default `published` for on prem and `automation-hub` for insights. (We'll probably change insights to use published as soon)

src/containers/collection-detail/collection-content.tsx
```diff
@@ -26,20 +28,42 @@ class CollectionContent extends React.Component<
...
       collection: undefined,
       params: params,
+      repo: props.match.params.repo,
     };
   }
 
   componentDidMount() {
-    this.loadCollection();
+    const { repo } = this.state;
+    if (DEPLOYMENT_MODE === Constants.STANDALONE_DEPLOYMENT_MODE) {
```
newswangerd, line 37: These checks to see if there's a repo set and to set a default repo should just go in `standalone-loader` and `insights-loader` so you only need to me this check once for the whole app.

src/containers/collection-detail/collection-detail.tsx
```diff
@@ -26,20 +29,43 @@ class CollectionDetail extends React.Component<
...
       params: params,
+      repo: props.match.params.repo,
+      redirect: false,
     };
   }
 
   componentDidMount() {
-    this.loadCollection();
+    const { repo } = this.state;
+    if (DEPLOYMENT_MODE === Constants.STANDALONE_DEPLOYMENT_MODE) {
```
newswangerd, line 39: Same here.

src/containers/collection-detail/collection-docs.tsx
```diff
@@ -43,23 +50,45 @@ class CollectionDocs extends React.Component<
...
+      repo: props.match.params.repo,
     };
 
     this.docsRef = React.createRef();
   }
 
   componentDidMount() {
-    this.loadCollection();
+    const { repo } = this.state;
+    if (DEPLOYMENT_MODE === Constants.STANDALONE_DEPLOYMENT_MODE) {
```
newswangerd, line 61: same here

src/containers/collection-detail/collection-import-log.tsx
```diff
@@ -35,10 +37,28 @@ class CollectionImportLog extends React.Component<RouteComponentProps, IState> {
       selectedImportDetail: undefined,
       selectedImport: undefined,
       apiError: undefined,
+      repo: props.match.params.repo,
     };
   }
 
   componentDidMount() {
+    const { repo } = this.state;
```
newswangerd, line 45: same here

src/containers/search/search.tsx
```diff
@@ -83,18 +87,42 @@ class Search extends React.Component<RouteComponentProps, IState> {
       numberOfResults: 0,
       loading: true,
       synclist: undefined,
+      redirect: false,
+      repo: props.location.pathname.split('/')[2],
     };
   }
 
   componentDidMount() {
+    const { repo } = this.state;
```
newswangerd, line 96: same here

src/loaders/standalone/standalone-loader.tsx
```diff
@@ -105,31 +113,65 @@ class App extends React.Component<RouteComponentProps, IState> {
...
+                    }}
+                    onSelect={(event, value) =>
+                      this.setState({ selectedRepo: value.toString() })
+                    }
+                  >
+                    <SelectOption
+                      key={'rh-certified'}
+                      value={'Red Hat Certified'}
+                    />
+                    <SelectOption key={'published'} value={'Published'} />
```
newswangerd, line 136: move published to the top

src/loaders/standalone/standalone-loader.tsx
```diff
@@ -105,31 +113,65 @@ class App extends React.Component<RouteComponentProps, IState> {
...
-              </NavItem>
-              <NavItem>
-                <Link to={Paths.myNamespaces}>My Namespaces</Link>
-              </NavItem>
-              <NavItem>
-                <Link to={Paths.token}>API Token</Link>
-              </NavItem>
-              {!!user && user.model_permissions.view_user && (
+              <NavGroup title='Content'>
+                <NavItem className={'nav-select'}>
```
newswangerd, line 117: Not a huge issue, but I the background behind the dropdown is highlighted when you mouse over it, which looks a little strange.

newswangerd (commented):
I also left you some notes on slack about why the context doesn't seem to update the app.

src/containers/collection-detail/collection-content.tsx
ZitaNemeckova: I didn't do that because it would switch to default repo if I went to "no repo" page like Users.

newswangerd: Oh, that's a good point. We need to solve this problem at some point, but we can figure it out later

newswangerd: I found a way to do this on the loader. You can use `componentDidUpdate` and check to see if the current location is a match for a repository and compare it the repo that is stored in state.

```
  componentDidUpdate(prevProps) {
    const match = matchPath(this.props.location.pathname, {
      path: '/repo/:repo',
    });
    if (match && Constants.ALLOWEDREPOS.includes(match.params['repo'])) {
      const newRepoName = Object.keys(Constants.REPOSITORYNAMES).find(
        key => Constants.REPOSITORYNAMES[key] === match.params['repo'],
      );

      if (newRepoName !== this.state.selectedRepo) {
        this.setState({ selectedRepo: newRepoName });
      }
    }
  }
```

This also appears to solve the issue pages other than search would get stuck on repo/published

src/loaders/standalone/standalone-loader.tsx
```diff
@@ -105,31 +113,79 @@ class App extends React.Component<RouteComponentProps, IState> {
...
+                    className='nav-select'
+                    variant='single'
+                    isOpen={this.state.selectExpanded}
+                    selections={this.state.selectedRepo}
+                    isPlain={false}
+                    onToggle={isExpanded => {
+                      this.setState({ selectExpanded: isExpanded });
+                    }}
+                    onSelect={(event, value) => {
+                      const originalRepo = this.state.selectedRepo;
```
newswangerd, line 128: repo toggle doesn't work on http://localhost:8002/ui/

src/loaders/standalone/routes.tsx
```diff
@@ -78,6 +78,8 @@ class AuthHandler extends React.Component<IProps, IState> {
 }
 
 export class Routes extends React.Component<{}> {
+  static contextType = AppContext;
+
   routes = [
```
newswangerd, line 83: I want to figure out how to use a sub router to route the `repo/:repo` routes in the future. I don't think it's too hard to do. No need to worry about that now.

newswangerd (changes requested):
One last note: legacy routes such as http://localhost:8002/ui/newswangerd/collection-demo don't work on insights or standalone mode. I think this is probably because of the ordering on the routes.

src/api/collection.ts
```diff
@@ -19,6 +20,11 @@ export class API extends BaseAPI {
     // new MockCollection(this.http, this.apiPath);
   }
 
+  list(params?: any, repo?: string) {
+    const path = '_ui/v1/repo/' + repo + '/';
```
newswangerd, line 24: use `this.getUIPath('repo')` instead of `_ui/v1/repo/`

src/api/collection.ts
```diff
@@ -75,9 +81,11 @@ export class API extends BaseAPI {
   getCached(
     namespace,
     name,
+    repo,
     params?,
     forceReload?: boolean,
   ): Promise<CollectionDetailType> {
+    const path = `_ui/v1/repo/${repo}/${namespace}/${name}/`;
```
newswangerd, line 88: use this.getUIPath('repo') instead of _ui/v1/repo/

src/api/collection.ts
newswangerd: since we don't use `collections/` anymore you can change the `apiPath=this.getUIPath('repo/')` and use `apiPath` here

newswangerd (approved):

PR #192, opened 2020-09-16
Add distro_base_path to download url pattern

bmclaughlin:
Closes issue: https://github.com/ansible/galaxy_ng/issues/442

### DIFF
--- a/src/api/collection.ts
+++ b/src/api/collection.ts
@@ -116,11 +116,13 @@ export class API extends BaseAPI {
     }
   }
 
-  getDownloadURL(namespace, name, version) {
+  getDownloadURL(distro_base_path, namespace, name, version) {
     // UI API doesn't have tarball download link, so query it separately here
     return new Promise((resolve, reject) => {
       this.http
-        .get(`v3/collections/${namespace}/${name}/versions/${version}/`)
+        .get(
+          `content/${distro_base_path}/v3/collections/${namespace}/${name}/versions/${version}/`,
+        )
         .then(result => {
           resolve(result.data['download_url']);
         })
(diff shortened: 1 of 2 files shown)


src/components/collection-detail/collection-info.tsx
```diff
@@ -89,7 +89,12 @@ export class CollectionInfo extends React.Component<IProps> {
                     variant='link'
                     icon={<DownloadIcon />}
                     onClick={() =>
-                      this.download(namespace, name, latest_version)
+                      this.download(
+                        Constants.REPOSITORYNAMES[this.context.selectedRepo],
```
newswangerd, line 93: you can just do `this.context.selectedRepo`. I just changed the selected repo so that it's the basepath and doesn't need to be passed through the `REPOSITORYNAMES` constant.

newswangerd (changes requested):

src/api/collection.ts
```diff
@@ -116,11 +116,13 @@ export class API extends BaseAPI {
...
   }
 
-  getDownloadURL(namespace, name, version) {
+  getDownloadURL(reponame, namespace, name, version) {
     // UI API doesn't have tarball download link, so query it separately here
     return new Promise((resolve, reject) => {
       this.http
-        .get(`v3/collections/${namespace}/${name}/versions/${version}/`)
+        .get(
+          `content/${reponame}/v3/collections/${namespace}/${name}/versions/${version}/`,
```
alikins, line 124: (non binding discussion comment on minor minor naming thing...)  for stuff like this, wonder if instead of 'reponame' if a more explicit 'distro_base_path' or 'distro_name' might be less confusing in the future. There is at least one place (the repo/distro's associated with synclists depending on if the synclist is empty or not) where this kind of comes into play.

alikins (commented):
just a comment about reponame/distroname/distrobasepath to consider in the future, doesn't need to be addressed at the moment.

src/api/collection.ts
newswangerd: that's a good point. We should refactor these variable names

newswangerd (approved):

PR #197, opened 2020-09-21
Prepare UI for branding

ZitaNemeckova:
Related to https://github.com/ansible/galaxy_ng/issues/465

#### Logos
- [x] create a `LargeLogo` and `SmallLogo` component that import and display `'static/images/logo_large.svg'` and `'static/images/logo_small.svg'` respectively. 
- [x] duplicate `static/images/galaxy_logo.svg` and rename it to `logo_small` and `logo_large` so that it gets used by default for both the large and small logos
- [x] use `SmallLogo` on the masthead 
- [x] use `LargeLogo` on the login page - changed to pic path

#### Add the following build time variables
- [x] `NAMESPACE_TERM` -> this will replace all instances term "partner" with "namespace". Update the partners page, and partners url to use `NAMESPACE_TERM` instead of "partners". Set this to "namespace" for the standalone builds and "partner" for the insights builds.
- [x] `APPLICATION_NAME` -> default this to `Galaxy NG`. This should be used for the app name on nave as well as the title in `index.html`

#### Navigation
- [x] update the app navigation
- [x] disable blue border when selected - to be done in a follow up PR
Screenshot 2020-09-21 at 14 13 27

#### Misc
- [x] on the certification dashboard, change the phrase "certified" to "approved"
- [x] on group management, rename the permissions to use more user friendly name and remove the `galaxy.` prefix. For example `galaxy.view_user` should display as `View user` Done in https://github.com/ansible/ansible-hub-ui/pull/198


newswangerd (changes requested):
You're missing the Application Name on the side bar
Screen Shot 2020-09-25 at 10 42 06 AM
Screen Shot 2020-09-25 at 10 40 52 AM

newswangerd:
Certification dashboard still uses the term "certification". We need to rename this to approval. So the dashboard should be renamed to the "Approval Dashboard".

Screen Shot 2020-09-25 at 10 46 00-fullpage

src/loaders/standalone/standalone-loader.tsx
```diff
@@ -129,81 +124,78 @@ class App extends React.Component<RouteComponentProps, IState> {
...
-                      repo: this.state.selectedRepo,
-                    })}
-                  >
-                    Collections
-                  </Link>
-                </NavItem>
-                <NavItem>
-                  <Link to={Paths.partners}>Namespaces</Link>
-                </NavItem>
+              <NavItem className={'nav-title'}>{APPLICATION_NAME}</NavItem>
```
newswangerd, line 127: I don't think `NavItem` is the right component for us to use here. `NavItem` causes the background to be highlighted and makes the application name look like a button, which is confusing. You might have to create your own CSS to make this work, it doesn't look like patternfly has the functionality we want here.

newswangerd (changes requested):

src/loaders/standalone/standalone-loader.tsx
ZitaNemeckova: @newswangerd Fixed so the nav title doesn't change apperance on hover and look like this
Screenshot 2020-09-29 at 17 28 51

Does it look ok?

newswangerd: Looks great!

newswangerd (approved):

PR #198, opened 2020-09-21
Convert Group permissions so a User sees them in human format

ZitaNemeckova:
Related to https://github.com/ansible/galaxy_ng/issues/465

Before:
Screenshot 2020-09-21 at 15 45 28

After:
Screenshot 2020-09-21 at 15 43 50

### DIFF
--- a/src/constants.tsx
+++ b/src/constants.tsx
@@ -99,6 +99,10 @@ export class Constants {
     'galaxy.view_synclist': 'View synclist',
     'galaxy.add_synclist': 'Add synclist',
   };
+  static GROUP_HUMAN_PERMISSIONS = {
+    change_namespace: 'Change namespace',
+    upload_to_namespace: 'Upload to namespace',
+  };
   static UPSTREAM_HOSTS = [
     'galaxy.ansible.com',
     'galaxy-dev.ansible.com',
--- a/src/constants.tsx
+++ b/src/constants.tsx
@@ -77,6 +77,28 @@ export class Constants {
       ],
     },
   ];
+  static HUMAN_PERMISSIONS = {
+    'galaxy.add_namespace': 'Add namespace',
+    'galaxy.change_namespace': 'Change namespace',
+    'galaxy.upload_to_namespace': 'Upload to namespace',
+    'ansible.modify_ansible_repo_content': 'Modify Ansible repo content',
+    'galaxy.view_user': 'View user',
+    'galaxy.delete_user': 'Delete user',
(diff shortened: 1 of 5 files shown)


newswangerd (commented):
Can you also add the human readable names to the namespace permission field on the create and edit namespaces forms?

Screen Shot 2020-09-22 at 9 01 41 AM
Screen Shot 2020-09-22 at 9 02 01 AM

src/containers/group-management/group-detail.tsx
```diff
@@ -297,6 +303,16 @@ class GroupDetail extends React.Component<RouteComponentProps, IState> {
     );
   }
 
+  private getPermissionCode(permission) {
```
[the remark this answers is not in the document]
ZitaNemeckova: Good idea. Created `twoWayMapper`

src/utilities/two-way-mapper.ts
```diff
@@ -0,0 +1,10 @@
+// Returns human value for DB permission and vise verse
+export function twoWayMapper(permission: string, mapper: any) {
```
newswangerd, line 2: One small nitpick.

This function is useful for any mapping, so I'd recommend renaming `permission` to something like `key` as well as updating the comment to reflect that it can be used for any map (not just permissions)

src/constants.tsx
```diff
@@ -77,6 +77,32 @@ export class Constants {
...
+    'ansible.view_collectionremote': 'View collection remote',
+    'ansible.change_ansibledistribution': 'Change Ansible distribution',
+    'ansible.view_ansibledistribution': 'View Ansible distribution',
+    'galaxy.delete_synclist': 'Delete synclist',
+    'galaxy.change_synclist': 'Change synclist',
+    'galaxy.view_synclist': 'View synclist',
+    'galaxy.add_synclist': 'Add synclist',
+  };
+  static GROUP_HUMAN_PERMISSIONS = {
+    change_namespace: 'Change namespace',
```
newswangerd, line 103: we really should just use the `galaxy.change_namespace` and `galaxy.upload_to_namespace` permissions, but it looks like the API doesn't include the app label when it returns object permissions on namespaces. I guess it's okay to leave it like this until we update the API.

newswangerd (approved):
A couple of small nitpicks. Once you fix them feel free to merge.

src/utilities/two-way-mapper.ts
ZitaNemeckova: Done :)

PR #202, opened 2020-09-23
Switch UI to using published repo in insights mode.

newswangerd:
...

### DIFF
--- a/src/constants.tsx
+++ b/src/constants.tsx
@@ -9,14 +9,11 @@ export class Constants {
   static readonly STANDALONE_DEPLOYMENT_MODE = 'standalone';
 
   static readonly ADMIN_GROUP = 'system:partner-engineers';
-  static PUBLISHED =
-    DEPLOYMENT_MODE === Constants.INSIGHTS_DEPLOYMENT_MODE
-      ? 'automation-hub'
-      : 'published';
+  static PUBLISHED = 'published';
 
   static CERTIFIED_REPO =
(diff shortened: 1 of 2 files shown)


src/constants.tsx
```diff
@@ -9,14 +9,11 @@ export class Constants {
...
-    DEPLOYMENT_MODE === Constants.INSIGHTS_DEPLOYMENT_MODE
-      ? 'automation-hub'
-      : 'published';
+  static PUBLISHED = 'published';
 
   static CERTIFIED_REPO =
     DEPLOYMENT_MODE === Constants.INSIGHTS_DEPLOYMENT_MODE
-      ? 'automation-hub'
+      ? 'published'
       : 'rh-certified';
```
awcrosby, line 17: Curious how the `CERTIFIED_REPO` variable is used? ... in standalone mode there will be both a `published` repo and an `rh-certified` repo, each with content that has been certified by either RH or the org.

newswangerd: We use `CERTIFIED_REPO` to determine which repo to put the certified flag on collections. Right now it just checks to see if the selected repo is the `CERTIFIED_REPO` and if it is, it puts the flag on all the collections the user is looking at. Right now we don't have a way to tell what other repos a collection is in, so if a collection is copied from rh-certified to published, it won't have the certified flag on it when it's viewed in the published repo.

awcrosby (approved):

PR #205, opened 2020-09-30
[TODO] Add Cog dropdown with  dropdown items

ZitaNemeckova:
TODO:
- [ ] create dropdown with correct items
- [ ] add correct urls (open in new tab/window? ask Susan)
- [ ] About modal has all relevant data (found out which are those)

Screenshot 2020-09-30 at 11 38 03
Screenshot 2020-09-30 at 12 31 13

Note to self: How to get Browser info =>  https://github.com/ManageIQ/manageiq-ui-classic/blob/master/app/assets/javascripts/miq_browser_detect.js ?


src/loaders/standalone/standalone-loader.tsx
```diff
@@ -107,11 +224,26 @@ class App extends React.Component<RouteComponentProps, IState> {
...
             ) : (
-              <StatefulDropdown
-                defaultText={userName}
-                toggleType='dropdown'
-                items={dropdownItems}
-              ></StatefulDropdown>
+              <div>
+                <Dropdown
+                  toggle={
+                    <CogIcon
```
newswangerd, line 230: This should be a question mark instead of the cog

newswangerd (changes requested):

src/loaders/standalone/standalone-loader.tsx
ZitaNemeckova: Fixed. And Browser detection updated.

src/loaders/app.scss
```diff
@@ -27,6 +27,9 @@ section.pf-c-nav__section.nav-title > h2 {
   font-size: var(--pf-c-nav__link--FontSize);
   font-weight: var(--pf-c-nav__link--FontWeight);
 }
+.cog-dropdown {
+  padding-right: 10px;
```
newswangerd, line 31: Add styling to make the button use the pointer cursor to indicate that it's clickable

src/loaders/standalone/standalone-loader.tsx
```diff
@@ -1,6 +1,7 @@
 // import PropTypes from 'prop-types';
 import * as React from 'react';
 import '../app.scss';
+const { detect } = require('detect-browser');
```
newswangerd, line 4: Don't use node style imports.

You should be able to import it as `import * as detect from 'detect-browser`. You might have to add `detect-browser` in `index.d.ts`

src/loaders/standalone/standalone-loader.tsx
```diff
@@ -58,9 +74,23 @@ class App extends React.Component<RouteComponentProps, IState> {
     }
   }
 
+  componentDidMount() {
+    ApplicationInfoAPI.get('').then(result => {
+      console.log(result);
```
newswangerd, line 79: Remove console

src/loaders/standalone/standalone-loader.tsx
```diff
@@ -89,6 +120,94 @@ class App extends React.Component<RouteComponentProps, IState> {
           Logout
         </DropdownItem>,
       ];
+      dropdownItemsCog = [
+        <DropdownItem
+          key='customer_support'
+          onClick={() =>
+            window.open('https://github.com/ansible/galaxy_ng/issues', '_blank')
```
newswangerd, line 127: Lets point users to https://access.redhat.com/support instead of our github issues. I don't want people flooding us with support questions.

src/loaders/standalone/standalone-loader.tsx
```diff
@@ -89,6 +120,94 @@ class App extends React.Component<RouteComponentProps, IState> {
...
+              '_blank',
+            )
+          }
+        >
+          Training
+        </DropdownItem>,
+        <DropdownItem
+          key='documentation'
+          onClick={() =>
+            window.open('https://galaxy.ansible.com/docs/', '_blank')
```
newswangerd, line 146: Lets point people here for documentation: https://access.redhat.com/documentation/en-us/red_hat_ansible_automation_platform/1.0/

The galaxy docs aren't really relevant.

newswangerd (changes requested):

newswangerd (approved):

PR #207, opened 2020-10-07
Remove mocked responses for api.

newswangerd:
I used these mocked apis in the very beginning of the project before we had a real api and there isn't any reason to keep them around anymore.


ZitaNemeckova (approved):
LGTM :+1:

PR #208, opened 2020-10-07
Remove packages with old node-fetch

ZitaNemeckova:
Removing Storybook as it's using old buggy node-fetch and we don't really need it.


newswangerd (changes requested):
Can we remove storybook as a dependency without removing all our stories? I would like to get storybook working again once we can sort out the deps.

ZitaNemeckova:
@newswangerd Why not just revert this PR with some minor changes once deps are sorted? That way there's no unusable code around in the meantime.

newswangerd (approved):

PR #209, opened 2020-10-08
Add distribution and group permission checks

ZitaNemeckova:
Related to https://github.com/ansible/galaxy_ng/issues/467

TODO:
- [x] check `view_group`
- [x] check `view_distribution`
- [x] check `change_group`
- [x] check `change_distribution`
- [x] check `add_group`
- [x] check `delete_group`
- [x] check `change_remote`

\+ removing some unused imports

### DIFF
--- a/src/api/response-types/user.ts
+++ b/src/api/response-types/user.ts
@@ -18,8 +18,10 @@ export class InsightsUserType {
 }
 
 export class Permissions {
+  add_group: boolean;
   add_namespace: boolean;
   upload_to_namespace: boolean;
+  change_distribution: boolean;
   change_namespace: boolean;
   move_collection: boolean;
   view_distribution: boolean;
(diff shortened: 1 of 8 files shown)


src/api/response-types/user.ts
```diff
@@ -18,12 +18,18 @@ export class InsightsUserType {
 }
 
 export class Permissions {
```
newswangerd, line 20: Remote permissions are missing here. Need to hide the sync and edit remote option is if the user doesn't have permissions to edit remotes.

newswangerd (changes requested):
Just missing remotes

src/api/response-types/user.ts
ZitaNemeckova: @newswangerd Done :)

newswangerd (approved):

PR #212, opened 2020-10-22
Allow collection detail view to display content from repos other than published

newswangerd:
Fixes https://issues.redhat.com/browse/AAH-46.


ZitaNemeckova:
It works when going directly to the URL but not when clicking it from Approval dashboard. Otherwise LGTM.

newswangerd:
> It works when going directly to the URL but not when clicking it from Approval dashboard. Otherwise LGTM.
> 
> Nevermind. LGTM and works in Insights mode.

@ZitaNemeckova  I doubled checked and it was broken in standalone mode just like you said. Should be fixed now

PR #214, opened 2020-10-27
Fix namespace links

newswangerd:
Fixes: https://issues.redhat.com/browse/AAH-18

Updates namespace links to do the following:
- Use the new field names introduced in https://github.com/ansible/galaxy_ng/pull/531
- The fields for adding links now automatically add links when the user starts typing rather than requiring the user to hit the plus icon first.

It's currently impossible to associate an error message with a specific link, but the error messages have been updated to to provide the name of the invalid URL as well as the field name that the error is from.


ZitaNemeckova (approved):
Tested in UI. LGTM :+1:

PR #236, opened 2020-11-11
Fix misc console errors.

newswangerd:

PR #239, opened 2020-11-12
Fix logout after changes to toolbar

ZitaNemeckova:
Fix after #236 that made some changes to toolbar.


ZitaNemeckova:
Closing in favor of #270

PR #240, opened 2020-11-13
Add test for Group Management

ZitaNemeckova:
TODO: 
- [x] create group
- [x] delete group
- [x] add permissions to a group
- [x] add all permissions to a group
- [x] remove permissions from a group
- [x] add user to a group
- [x] remove user from a group

### DIFF
--- a/test/cypress/integration/group_management.js
+++ b/test/cypress/integration/group_management.js
@@ -12,7 +12,8 @@ describe('Hub Group Management Tests', () => {
         cy.login(adminUsername, adminPassword);
         cy.createGroup(name);
         cy.contains(name).should('exist');
-        cy.deleteGroup(name);
-        cy.contains(name).should('not.exist');
+        cy.addAllPermissions(name);
+        //cy.deleteGroup(name);
+        //cy.contains(name).should('not.exist');
     });
 });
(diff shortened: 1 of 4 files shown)


ZitaNemeckova:
@hendersonreed please review. Thanks :)

test/cypress/integration/group_management.js
```diff
@@ -0,0 +1,54 @@
...
+    });
+
+    it('admin user can add/remove permissions to/from a group', () => {
+        let name = 'testGroup';
+        cy.login(adminUsername, adminPassword);
+        cy.createGroup(name);
+        cy.contains(name).should('exist');
+        cy.addAllPermissions(name);
+        [ 'namespaces', 'collections', 'users', 'groups', 'remotes' ].forEach(permGroup => cy.get(`.pf-l-flex.pf-m-align-items-center.${permGroup}  [placeholder="No permission"]`).should('not.exist'));;
+        cy.removePermissions(name, [{
```
hendersonreed, line 38: Should we just use `removeAllPermissions` here?

hendersonreed (approved):
Verified that it runs on my machine in both FF and Chromium!

test/cypress/integration/group_management.js
ZitaNemeckova: Fixed :)

hendersonreed:
Looks good to me! They continue to pass on my machine as well.

himdel:
Backporting because #305 (backported) depends on `createGroup` from this one + https://github.com/ansible/ansible-hub-ui/pull/301 (which conflicts without it).

PR #244, opened 2020-11-16
Update and add instructions on connect to hub page.

newswangerd:
- Rephrase some of the descriptions to not be specific to the `ansible-galaxy` client.
- Connect Private Automation Hub section

Screen Shot 2020-11-16 at 15 22 23-fullpage

### DIFF
--- a/src/containers/token/token-insights.tsx
+++ b/src/containers/token/token-insights.tsx
@@ -58,7 +58,7 @@ class TokenPage extends React.Component<RouteComponentProps, IState> {
           <Section className='body pf-c-content'>
             <h2>Connect the ansible-galaxy client</h2>
             <p>
-              Documentation how to configure the <code>ansible-galaxy</code>{' '}
+              Documentation on how to configure the <code>ansible-galaxy</code>{' '}
               client can be found{' '}
               <a
                 href='https://access.redhat.com/documentation/en-us/red_hat_ansible_automation_platform/1.0/html/getting_started_with_red_hat_ansible_automation_hub/proc-configure-automation-hub-server'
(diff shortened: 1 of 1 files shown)


cbudz:
Nav:Connect to Hub > I assume *Connect to private Automation Hub* won't fit but to the best of my knowledge we don't have confirmed short-form names for these services yet, but I'd try to stick with 'Sync to ...' since everywhere else we use 'sync' and not 'connect' as the primary action. 
- Offline token - consider replacing the second sentence with "This is a secret token used to protect your content. Store your API token in a secure location."  Not sure about purposefully telling people it works for all apps is best practice.
- Connect Private Automation Hub - consider moving this to the top of the page since it parallels the nav item.  Also follow the construction of your descriptions under Server URL and SSO URL, since both are ideally prescriptive. Ex, "Use the Repository Management page to sync collections curated by your organization to the Red Hat Certified repository in your private Automation Hub."
If the last sentence applies to the same persona then just start it off with "You can use the sync toggle ...." - if not then its fine as it stands.

newswangerd:
> but I'd try to stick with 'Sync to ...' since everywhere else we use 'sync' and not 'connect' as the primary action

This page is also meant for connecting the `ansible-galaxy` client to hub, so sync isn't really the right term.

Since this page is also used for connecting the galaxy client, should there be a section for that as well?

cbudz:
One more - if the last bit about syncing can be broken out into its own header it makes the flow of information about using the toggle->create sync list its own unique activity thats performed separately from syncing to private hub.

cbudz:
> > but I'd try to stick with 'Sync to ...' since everywhere else we use 'sync' and not 'connect' as the primary action
> 
> This page is also meant for connecting the `ansible-galaxy` client to hub, so sync isn't really the right term.
> 
> Since this page is also used for connecting the galaxy client, should there be a section for that as well?

I'd say that's its own activity so its own header would help differentiate things.

newswangerd:
How does this look?

Screen Shot 2020-11-16 at 16 30 05-fullpage

cbudz:
One quick fix : "Documentation *on* how to configure ..."

New layout is much easier to read and understand what you're trying to do.

PR #248, opened 2020-11-18
Make logo clickable

ZitaNemeckova:
Fix-Issue: AAH-162

### DIFF
--- a/src/loaders/standalone/standalone-loader.tsx
+++ b/src/loaders/standalone/standalone-loader.tsx
@@ -223,7 +223,15 @@ class App extends React.Component<RouteComponentProps, IState> {
 
     const Header = (
       <PageHeader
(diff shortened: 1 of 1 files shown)


src/loaders/standalone/standalone-loader.tsx
```diff
@@ -223,7 +223,11 @@ class App extends React.Component<RouteComponentProps, IState> {
 
     const Header = (
       <PageHeader
-        logo={<SmallLogo alt='Galaxy Logo'></SmallLogo>}
+        logo={
+          <a href={'/'}>
```
newswangerd, line 227: Why not `Link`?

Also, unrelated, but we should change `Galaxy Logo` to use the `APPLICATION_NAME`

ZitaNemeckova: Fixed :)

newswangerd (approved):

PR #254, opened 2020-11-19
Fix missing versions in Aboutt Modal

ZitaNemeckova:
Move AboutModal to be a component to get componentDidMount to work properly.

Fix-Issue: AAH-161

Steps to reproduce:
- log out of hub
- refresh the page to clear out state
- log in
- open modal

Before:
Screenshot 2020-11-19 at 14 33 25

After:
Screenshot 2020-11-19 at 14 27 11

### DIFF
--- a/src/containers/about-modal/about-modal.tsx
+++ b/src/containers/about-modal/about-modal.tsx
@@ -1,0 +1,110 @@
+import * as React from 'react';
+import {
+  AboutModal,
+  TextContent,
+  TextList,
+  TextListItem,
+  TextListItemVariants,
+  TextListVariants,
+} from '@patternfly/react-core';
+import Logo from '../../../static/images/logo_large.svg';
+import { ApplicationInfoAPI, UserType } from '../../api';
+import { detect } from 'detect-browser';
+
+interface IProps {
+  isOpen: boolean;
+  trademark: string;
+  brandImageSrc: string;
+  onClose: () => void;
+  brandImageAlt: string;
+  productName: string;
+  user: UserType;
+  userName: string;
+}
+
+interface IState {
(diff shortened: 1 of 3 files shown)


src/components/about-modal/about-modal.tsx
```diff
@@ -0,0 +1,114 @@
...
+    this.state = {
+      applicationInfo: { server_version: '', pulp_ansible_version: '' },
+    };
+  }
+
+  componentDidMount() {
+    console.log('componentDidMount');
+    ApplicationInfoAPI.get('')
+      .then(result => {
+        console.log(result);
```
newswangerd, line 41: Might want to remove this

src/components/about-modal/about-modal.tsx
```diff
@@ -0,0 +1,114 @@
...
+  productName: string;
+  user: UserType;
+  userName: string;
+}
+
+interface IState {
+  applicationInfo: { server_version: string; pulp_ansible_version: string };
+}
+
+export class AboutModalWindow extends React.Component<IProps, IState> {
```
newswangerd, line 29: I'm a little conflicted about classifying this as a component instead of a container. On one hand, it's making API requests and it has zero re-usability. On the other hand It's not a standalone page like all of our other containers are.

My gut reaction is to move this into containers. What do you think?

ZitaNemeckova: Moved :)

src/components/about-modal/about-modal.tsx
ZitaNemeckova: Removed all console.logs

newswangerd (approved):

PR #259, opened 2020-11-23
Surface API error on Group form

ZitaNemeckova:
Fix-Issue: AAH-171

Steps to reproduce:
- try to create a group with a name that already exists
- click Create

Before:
Screenshot 2020-11-23 at 16 40 03

After:
Screenshot 2020-11-23 at 16 38 49

### DIFF
--- a/src/components/group-management/group-modal.tsx
+++ b/src/components/group-management/group-modal.tsx
@@ -12,12 +12,11 @@ interface IProps {
   onSave?: (string) => void;
   onChange?: () => void;
   group?: any;
-  errorMessage?: string;
+  errorMessage?: any;
 }
 
 interface IState {
   name: string;
-  errorMessage: string;
 }
 
 export class GroupModal extends React.Component<IProps, IState> {
@@ -28,17 +27,7 @@ export class GroupModal extends React.Component<IProps, IState> {
         !this.props.group || !this.props.group.name
           ? ''
           : this.props.group.name,
-      errorMessage: this.props.errorMessage,
     };
-  }
-  componentDidUpdate(
-    prevProps: Readonly<IProps>,
-    prevState: Readonly<IState>,
-    snapshot?: any,
-  ): void {
-    if (prevState.errorMessage !== this.props.errorMessage) {
-      this.setState({ errorMessage: this.props.errorMessage });
-    }
   }
 
   render() {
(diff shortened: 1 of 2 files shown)


src/containers/group-management/group-list.tsx
```diff
@@ -247,12 +257,16 @@ class GroupList extends React.Component<RouteComponentProps, IState> {
...
-    });
+    GroupAPI.create({ name: value })
+      .then(result => {
+        this.setState({
+          redirect: '/group/' + result.data.id,
+          createModalVisible: false,
+        });
+      })
+      .catch(error =>
+        this.setState({ groupError: error.response.data.errors[0].detail }),
```
newswangerd, line 268: Assuming the error is a standard error response, you should use this function to parse the error message and then check for the field that you're interested in: https://github.com/ansible/ansible-hub-ui/blob/master/src/utilities/map-error-messages.ts

src/components/group-management/group-modal.tsx
```diff
@@ -19,14 +28,28 @@ export class GroupModal extends React.Component<IProps, IState> {
...
           ? ''
           : this.props.group.name,
+      errorMessage: this.props.errorMessage,
     };
   }
+  componentDidUpdate(
+    prevProps: Readonly<IProps>,
+    prevState: Readonly<IState>,
+    snapshot?: any,
+  ): void {
```
newswangerd, line 38: Setting the error message as state and a prop is a really bad idea because it leads to this messy logic here, which is error prone and hard to debug. When dealing with react, you should always have a single source of truth whenever possible. (As a side note, we don't have a single source of truth for the global repo, and it trying to keep the URL state and internal state in sync is a nightmare and leads to weird errors like this: #262)

I recommend only using a prop and passing in a `clearErrors` or `updateErrors` function if you need to update the error message.

newswangerd (changes requested):

src/containers/group-management/group-list.tsx
```diff
@@ -247,12 +256,14 @@ class GroupList extends React.Component<RouteComponentProps, IState> {
...
-    GroupAPI.create({ name: value }).then(result => {
-      this.setState({
-        redirect: '/group/' + result.data.id,
-        createModalVisible: false,
-      });
-    });
+    GroupAPI.create({ name: value })
+      .then(result => {
+        this.setState({
+          redirect: '/group/' + result.data.id,
```
newswangerd, line 262: This should use `Paths.groupDetail`

src/components/group-management/group-modal.tsx
```diff
@@ -1,10 +1,18 @@
...
+  Form,
+  FormGroup,
+  Modal,
+  TextInput,
+} from '@patternfly/react-core';
 
 interface IProps {
   onCancel?: () => void;
   onSave?: (string) => void;
+  onChange?: () => void;
```
newswangerd, line 13: `onChange` is a little bit confusing because that's a fairly common prop name that's used by a lot of different components in react. Typically on change would also pass in the value that's changing and would be used to update state on the parent.

I recommend being explicit and using something like `clearErrors` or `updateErrors`.

newswangerd (changes requested):

newswangerd (approved):

PR #266, opened 2020-12-01
Add tooltips to user's password field and fixes namespace's to PF version

ZitaNemeckova:
Closes-Issue: AAH-159

**NAMESPACE CREATE FORM**
Before:
Screenshot 2020-12-01 at 15 09 27

After:
Screenshot 2020-12-02 at 14 20 49

**USER CREATE FORM**
Before:
Screenshot 2020-12-01 at 15 08 30

After:
Screenshot 2020-12-02 at 14 20 40

**REMOTE REPO TABLE**
Before:
Screenshot 2020-12-03 at 16 15 06

After:
Screenshot 2020-12-03 at 15 59 10

Used password validators:
```
class MinimumLengthValidator(min_length=8)¶
Validates whether the password meets a minimum length. The minimum length can be customized with the min_length parameter.

class UserAttributeSimilarityValidator(user_attributes=DEFAULT_USER_ATTRIBUTES, max_similarity=0.7)¶
Validates whether the password is sufficiently different from certain attributes of the user.

The user_attributes parameter should be an iterable of names of user attributes to compare to. If this argument is not provided, the default is used: 'username', 'first_name', 'last_name', 'email'. Attributes that don’t exist are ignored.

The minimum similarity of a rejected password can be set on a scale of 0 to 1 with the max_similarity parameter. A setting of 0 rejects all passwords, whereas a setting of 1 rejects only passwords that are identical to an attribute’s value.

class CommonPasswordValidator(password_list_path=DEFAULT_PASSWORD_LIST_PATH)¶
Validates whether the password is not a common password. This converts the password to lowercase (to do a case-insensitive comparison) and checks it against a list of 20,000 common password created by Royce Williams.

The password_list_path can be set to the path of a custom file of common passwords. This file should contain one lowercase password per line and may be plain text or gzipped.

class NumericPasswordValidator¶
Validates whether the password is not entirely numeric.
```
from https://docs.djangoproject.com/en/3.1/topics/auth/passwords/

### DIFF
--- a/src/components/helper-text/helper-text.tsx
+++ b/src/components/helper-text/helper-text.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import { Popover, PopoverPosition, Button } from '@patternfly/react-core';
 import { OutlinedQuestionCircleIcon } from '@patternfly/react-icons';
+import './helper-text.scss';
 
 interface IProps {
   /** Value to display in the tag */
@@ -15,7 +16,7 @@ export class HelperText extends React.Component<IProps, {}> {
         position={PopoverPosition.top}
         bodyContent={this.props.content}
       >
-        <Button iconPosition={'left'} variant={'plain'} padding={'0px'}>
+        <Button iconPosition={'left'} variant={'plain'} className={'helper'}>
           <OutlinedQuestionCircleIcon />
         </Button>
       </Popover>
--- a/src/components/helper-text/helper-text.tsx
+++ b/src/components/helper-text/helper-text.tsx
@@ -1,0 +1,24 @@
+import * as React from 'react';
+import { Popover, PopoverPosition, Button } from '@patternfly/react-core';
+import { OutlinedQuestionCircleIcon } from '@patternfly/react-icons';
+
+interface IProps {
+  /** Value to display in the tag */
+  content: string;
+}
+
+export class HelperText extends React.Component<IProps, {}> {
+  render() {
+    return (
+      <Popover
+        aria-label='popover example'
+        position={PopoverPosition.top}
+        bodyContent={this.props.content}
+      >
+        <Button iconPosition={'left'} variant={'plain'} padding={'0px'}>
+          <OutlinedQuestionCircleIcon />
+        </Button>
+      </Popover>
+    );
+  }
+}
--- a/src/components/namespace-modal/namespace-modal.tsx
+++ b/src/components/namespace-modal/namespace-modal.tsx
@@ -5,7 +5,7 @@ import { Button, InputGroup, TextInput } from '@patternfly/react-core';
 import { OutlinedQuestionCircleIcon } from '@patternfly/react-icons';
 import { NamespaceAPI, GroupObjectPermissionType } from '../../api';
 
-import { ObjectPermissionField } from '../../components';
+import { HelperText, ObjectPermissionField } from '../../components';
 
 interface IProps {
   isOpen: boolean;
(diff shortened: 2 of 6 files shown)


ZitaNemeckova:
@sbuenafe-rh @fionayhlin What do you think about wording of password tooltip? Any suggestions to make it better? Thanks :)
Screenshot 2020-12-01 at 14 55 48

sbuenafe-rh:
@ZitaNemeckova I would use the popover component (white bkg) with the info icon. I would also format at least the password requirements in a bulleted list (see screenshot below). We can run the text copy by @cbudz for better wording. 

Screen Shot 2020-12-01 at 11 50 20

cbudz:
Hey @ZitaNemeckova and @sbuenafe-rh -Does the password field support special characters? Also if it doesn't validate against what's up there about user attributes and common passwords i'd just say you can have one or two simple and prescriptive sentence like "Create a password using at least 9 characters, including special characters <or list of supported special characters>, ex <x!!f3@pl$%>. Avoid using common names or expressions.

src/components/user-form/user-form.tsx
```diff
@@ -99,6 +99,23 @@ export class UserForm extends React.Component<IProps, IState> {
             label={v.title}
             validated={this.toError(!(v.id in errorMessages))}
             helperTextInvalid={errorMessages[v.id]}
+            labelIcon={
+              v.id === 'password' && (
+                <Tooltip
+                  position='top'
+                  content={
+                    <div>
+                      Password must contain at least 9 characters. Password
```
newswangerd, line 108: Can this be formatted like a list?

src/components/namespace-modal/namespace-modal.tsx
```diff
@@ -122,6 +115,16 @@ export class NamespaceModal extends React.Component<IProps, IState> {
             helperText='Please, provide the namespace name'
             helperTextInvalid={this.state.errorMessages['name']}
             validated={this.toError(this.state.newNamespaceNameValid)}
+            labelIcon={
+              <Tooltip
+                position='top'
+                content={
+                  'Namespace names are limited to lowercase word characters ([a-zA-Z0-9_]), must have a minimum length of 2 characters and cannot start with an ‘_’.'
```
newswangerd, line 122: instead of `([a-zA-Z0-9_])` we should just say "alphanumeric characters and underscores".

src/components/namespace-modal/namespace-modal.tsx
```diff
@@ -122,6 +115,16 @@ export class NamespaceModal extends React.Component<IProps, IState> {
             helperText='Please, provide the namespace name'
             helperTextInvalid={this.state.errorMessages['name']}
             validated={this.toError(this.state.newNamespaceNameValid)}
+            labelIcon={
+              <Tooltip
```
newswangerd, line 119: I think we should create a `HelperText` component that just take the text to display. That way whenever we want to add helper texts to forms we can just do

```
<HelperText>This is how you do the thing.</HelperText>
```

Instead of
```
<Tooltip position={...} content={}>
<Icon/>
</Tooltip>
```

newswangerd (changes requested):
This is a good start, but there are still a lot of other confusing fields that need helper texts.

newswangerd:
Notes from the meeting today:

Use popovers and not tooltips.

- upload collection form: explain character requirments
- namespace form:
  - useful links
    - change name to related links
  - logo url: link to image, size requirements, file types etc

- User form:
  - Username - any validation requirments
  - Password - validation requirments
  - group create - validation for name

- Edit group permissions form
  - add desciption text underneath permission fields
    - add for all fields
  - also rename modify ansible repo content to something that makes more sense

- remote form
  - add popovers for everything except name
  - use a wider modal for this

ZitaNemeckova:
@sbuenafe-rh @cbudz updated as you suggested. Please have a look that it's ok, thank you :)

ZitaNemeckova:
@newswangerd this is ready for another round of review :) Other tooltips/popups will be added in a followup PR.

src/components/helper-text/helper-text.tsx
```diff
@@ -0,0 +1,25 @@
+import * as React from 'react';
+import { Popover, PopoverPosition, Button } from '@patternfly/react-core';
+import { OutlinedQuestionCircleIcon } from '@patternfly/react-icons';
+import './helper-text.scss';
+
+interface IProps {
+  /** Value to display in the tag */
+  content: string;
```
newswangerd, line 8: You don't have to do this, but if you change the prop name to `children` you can pass it in as `<HelperText>foo</HelperText> instead of `<HelperText content='foo' />`. Both are fine and you don't need to change this, but i just wanted to call this out.

newswangerd (approved):

PR #276, opened 2020-12-10
Verify 'nopermissions' user can't see stuff they shouldn't.

hendersonreed:
The current test for how permissions interact with what appears on the sidebar only verifies that users without permissions **can** see what they **should** see. i.e. that they can see all the ordinary things like 'Collections', 'Namespaces', 'API Token', etc.

This PR extends that test to verify that users without permissions **cannot** see things that they *shouldn't** see, i.e. that they can't see 'Users', 'Groups', and 'Approval'.


ZitaNemeckova:
LGTM 👍

PR #280, opened 2020-12-14
Add test to ensure 'download_concurrency' config is surfaced.

hendersonreed:
This test just verifies that the option exists in the UI - I think that's all the coverage required in the UI tests. Thoughts/opinions/reviews/etcetera from @ZitaNemeckova and @ironfroggy are greatly appreciated.

### DIFF
--- a/test/cypress/integration/remote_options.js
+++ b/test/cypress/integration/remote_options.js
@@ -1,0 +1,19 @@
+describe('Hub Menu Tests', () => {
+    let host = Cypress.env('host');
+    let adminUsername = Cypress.env('username');
+    let adminPassword = Cypress.env('password');
+
+    beforeEach(() => {
+        cy.visit(host)
+    });
+
+    it('admin user sees download_concurrency in remote config', () => {
+       cy.login(adminUsername, adminPassword);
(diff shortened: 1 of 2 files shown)


ZitaNemeckova:
@hendersonreed It fails sporadically for me on
Screenshot 2020-12-15 at 14 35 59

I think it needs to wait for `http://localhost:8002/ui/repositories?page_size=10&tab=remote` to finish.

hendersonreed:
Good to know - I'll add that wait after standup! Thanks.

hendersonreed:
@ZitaNemeckova I added a wait but that ended up causing a different kind of error. I elected to use `cy.visit` to navigate directly to the repo page - please let me know if you're still seeing intermittent failures - I haven't seen any yet.

ZitaNemeckova:
My remote repos are failed and the test tries to click the `failed` label instead of the kebab
Screenshot 2020-12-16 at 16 22 43

Kebab has a `aria-label="Actions"`. Maybe that will help.

hendersonreed:
@ZitaNemeckova Updated - don't know why I didn't see that aria-label at first. Thanks for working through this with me!

ZitaNemeckova:
LGTM :+1:

PR #285, opened 2021-01-04
Add Empty state components and use them

ZitaNemeckova:
How it looks so far:
Screenshot 2021-01-26 at 13 53 44
Screenshot 2021-01-26 at 13 52 30
Screenshot 2021-01-26 at 14 49 05
Screenshot 2021-01-26 at 14 51 27

Screenshot 2021-01-26 at 16 09 37

Screenshot 2021-01-26 at 16 20 05

Screenshot 2021-01-25 at 15 56 28

TODO: 
- [x] add EmptyState component for "filter", "unauthorised" and "no data yet"
- [x] make them fully functional
  - [x] Clear all filters - removed based on UX feedback
  - [x] Add button 
- [x] replace code with components
  - [x] filter
  - [x] unauthorised
  - [x] no data yet
- [x] add check for filter/no data
- [x] src/containers/collection-detail/collection-docs.tsx -> Use EmptyState with red ! icon - followup PR
- [x] EmptyStateNoData -> if possible to add data use "add" icon + primary button else use "search" icon
- [x] "No stuff yet" -> no headers, no footer, no pagination, just title
- [x] fix redirect as it's not always `not-found`
- [x] Exe. Env. needs to use the new nes too.

List of all pages:

Empty state - No data yet:
- Collections
- Namespaces
- My namespaces
- Users (it's there but it shouldn't happen)
- Groups
- Groups - select a group - click on Users tab
- Repo management - Remote tab (it's there but it shouldn't happen)
- Approval 

Empty state - No data for filter:
- Collections
- Namespaces
- My namespaces
- Users
- Groups
- Groups - select a group - click on User tab
- Approval

Empty state - Unauthorised:
- Groups (view and add/edit)
- Users (view and add/edit)
- Approval

Custom "errror" state:
- Collections - select a collection - click documentation tab (it's there but it shouldn't happen)

### DIFF
--- a/src/components/index.ts
+++ b/src/components/index.ts
@@ -52,3 +52,4 @@ export { WriteOnlyField } from './patternfly-wrappers/write-only-field';
 export { EmptyStateFilter } from './empty-state/empty-state-filter';
 export { EmptyStateUnauthorised } from './empty-state/empty-state-unauthorised';
 export { EmptyStateNoData } from './empty-state/empty-state-no-data';
+export { EmptyStateCustom } from './empty-state/empty-state-custom';
--- a/src/components/index.ts
+++ b/src/components/index.ts
@@ -49,3 +49,6 @@ export { LocalRepositoryTable } from './repositories/local-repository-table';
 export { StatusIndicator } from './status/status-indicator';
 export { HelperText } from './helper-text/helper-text';
 export { WriteOnlyField } from './patternfly-wrappers/write-only-field';
+export { EmptyStateFilter } from './empty-state/empty-state-filter';
+export { EmptyStateUnauthorised } from './empty-state/empty-state-unauthorised';
+export { EmptyStateNoData } from './empty-state/empty-state-no-data';
--- a/src/components/index.ts
+++ b/src/components/index.ts
@@ -50,6 +50,6 @@ export { StatusIndicator } from './status/status-indicator';
 export { HelperText } from './helper-text/helper-text';
 export { WriteOnlyField } from './patternfly-wrappers/write-only-field';
 export { EmptyStateFilter } from './empty-state/empty-state-filter';
-export { EmptyStateUnauthorised } from './empty-state/empty-state-unauthorised';
+export { EmptyStateUnauthorized } from './empty-state/empty-state-unauthorized';
 export { EmptyStateNoData } from './empty-state/empty-state-no-data';
 export { EmptyStateCustom } from './empty-state/empty-state-custom';
--- a/src/components/repositories/remote-repository-table.tsx
+++ b/src/components/repositories/remote-repository-table.tsx
@@ -5,7 +5,7 @@ import { Button, DropdownItem, Tooltip } from '@patternfly/react-core';
 import { ExclamationCircleIcon } from '@patternfly/react-icons';
 
 import { RemoteType, UserType, PulpStatus } from '../../api';
-import { EmptyStateNoData, HelperText, SortTable, StatefulDropdown } from '..';
+import { HelperText, SortTable, StatefulDropdown } from '..';
 import { StatusIndicator } from '../../components';
 
 import { Constants } from '../../constants';
@@ -55,14 +55,6 @@ export class RemoteRepositoryTable extends React.Component<IProps> {
 
   render() {
     const { remotes } = this.props;
-    if (remotes.length !== 0) {
-      return (
-        <EmptyStateNoData
-          title={'No remote repositories yet'}
-          description={''}
-        />
-      );
-    }
     return this.renderTable(remotes);
   }
 
--- a/src/components/repositories/remote-repository-table.tsx
+++ b/src/components/repositories/remote-repository-table.tsx
@@ -1,19 +1,11 @@
 import * as React from 'react';
 import * as moment from 'moment';
 
-import {
-  Button,
-  DropdownItem,
-  EmptyState,
-  EmptyStateIcon,
-  EmptyStateVariant,
-  Title,
-  Tooltip,
-} from '@patternfly/react-core';
-import { WrenchIcon, ExclamationCircleIcon } from '@patternfly/react-icons';
+import { Button, DropdownItem, Tooltip } from '@patternfly/react-core';
+import { ExclamationCircleIcon } from '@patternfly/react-icons';
 
 import { RemoteType, UserType, PulpStatus } from '../../api';
-import { EmptyStateFilter, HelperText, SortTable, StatefulDropdown } from '..';
+import { EmptyStateNoData, HelperText, SortTable, StatefulDropdown } from '..';
 import { StatusIndicator } from '../../components';
 
 import { Constants } from '../../constants';
@@ -63,19 +55,8 @@ export class RemoteRepositoryTable extends React.Component<IProps> {
 
   render() {
     const { remotes } = this.props;
-    if (remotes.length == 0) {
-      return (
-        <EmptyState className='empty' variant={EmptyStateVariant.full}>
-          <EmptyStateIcon icon={WrenchIcon} />
-          <Title headingLevel='h2' size='lg'>
-            No remote repos
-          </Title>
-        </EmptyState>
-      );
-    }
-    // TODO only with search
-    if (remotes.length === 0) {
-      return <EmptyStateFilter />;
+    if (remotes.length !== 0) {
+      return <EmptyStateNoData />;
     }
     return this.renderTable(remotes);
   }
--- a/src/components/repositories/remote-repository-table.tsx
+++ b/src/components/repositories/remote-repository-table.tsx
@@ -5,21 +5,15 @@ import {
   Button,
   DropdownItem,
   EmptyState,
-  EmptyStateBody,
   EmptyStateIcon,
   EmptyStateVariant,
   Title,
   Tooltip,
-  Popover,
 } from '@patternfly/react-core';
-import {
-  WarningTriangleIcon,
-  WrenchIcon,
-  ExclamationCircleIcon,
-} from '@patternfly/react-icons';
+import { WrenchIcon, ExclamationCircleIcon } from '@patternfly/react-icons';
 
 import { RemoteType, UserType, PulpStatus } from '../../api';
-import { HelperText, SortTable, StatefulDropdown } from '..';
+import { EmptyStateFilter, HelperText, SortTable, StatefulDropdown } from '..';
 import { StatusIndicator } from '../../components';
 
 import { Constants } from '../../constants';
@@ -81,17 +75,7 @@ export class RemoteRepositoryTable extends React.Component<IProps> {
     }
     // TODO only with search
     if (remotes.length === 0) {
-      return (
-        <EmptyState className='empty' variant={EmptyStateVariant.full}>
-          <EmptyStateIcon icon={WarningTriangleIcon} />
-          <Title headingLevel='h2' size='lg'>
-            No matches
-          </Title>
-          <EmptyStateBody>
-            Please try adjusting your search query.
-          </EmptyStateBody>
-        </EmptyState>
-      );
+      return <EmptyStateFilter />;
     }
     return this.renderTable(remotes);
   }
--- a/src/components/repositories/remote-repository-table.tsx
+++ b/src/components/repositories/remote-repository-table.tsx
@@ -56,7 +56,12 @@ export class RemoteRepositoryTable extends React.Component<IProps> {
   render() {
     const { remotes } = this.props;
     if (remotes.length !== 0) {
-      return <EmptyStateNoData />;
+      return (
+        <EmptyStateNoData
+          title={'No remote repositories yet'}
+          description={''}
+        />
+      );
     }
     return this.renderTable(remotes);
   }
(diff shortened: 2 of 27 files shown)


ZitaNemeckova:
@sbuenafe-rh @fionayhlin does it look to be heading the right way?

sbuenafe-rh:
@ZitaNemeckova for the **No users added to this group**, you can shorten that title to **No users**

fionayhlin:
@ZitaNemeckova @sbuenafe-rh Should that be **No users yet** to match the other titles?

sbuenafe-rh:
@fionayhlin @ZitaNemeckova yes! good catch.

ZitaNemeckova:
Container Registry:
Screenshot 2021-02-03 at 12 39 09
Screenshot 2021-02-03 at 12 39 52

Groups:
Screenshot 2021-02-03 at 12 48 18

Collection documentation:
Screenshot 2021-02-03 at 12 50 46

Group detail - users:
Screenshot 2021-02-03 at 12 54 21

@sbuenafe-rh @fionayhlin here's new Container Registry empty pages and fixed ones. WDYT?

ZitaNemeckova:
@himdel please have a look code-wise, thanks :)

himdel:
Sure, thanks for the help getting it working :), some testing...

Repo Management > Remote tab - always shows empty state, on master there are 2 items
My namespaces - Create button doesn't work (the empty state one, the filter one does, but shouldn't be there)
Namespace & My namespaces - filter still there on empty state

(I have yet to go through the code, next comment :))

(I also noticed the UI doesn't stop me from removing the current user, which is probably a bug too, though unrelated :))

And do we have a list of all the screens with an EmptyState?

src/components/index.ts
```diff
@@ -48,3 +48,7 @@ export { RemoteRepositoryTable } from './repositories/remote-repository-table';
 export { LocalRepositoryTable } from './repositories/local-repository-table';
 export { StatusIndicator } from './status/status-indicator';
 export { HelperText } from './helper-text/helper-text';
+export { EmptyStateFilter } from './empty-state/empty-state-filter';
+export { EmptyStateUnauthorised } from './empty-state/empty-state-unauthorised';
```
himdel, line 52: this should probably be spelled Unauthori**z**ed (`s` -> `z`)?

src/components/repositories/remote-repository-table.tsx
```diff
@@ -69,28 +55,12 @@ export class RemoteRepositoryTable extends React.Component<IProps> {
 
   render() {
     const { remotes } = this.props;
-    if (remotes.length == 0) {
+    if (remotes.length !== 0) {
```
himdel, line 58: this is causing the Remote Repositories issue

chouseknecht:
@himdel Thanks for reviewing our code! Appreciate the comments. I think @alikins is working on the issue regarding users deleting their own account. [View the issue here](https://issues.redhat.com/browse/AAH-265).

ZitaNemeckova:
Screenshot 2021-02-08 at 12 52 07
Screenshot 2021-02-08 at 13 18 14
Screenshot 2021-02-08 at 13 18 24
Screenshot 2021-02-08 at 13 28 08
Screenshot 2021-02-08 at 13 28 18
Screenshot 2021-02-08 at 13 56 32

@sbuenafe-rh @fionayhlin some more Empty State pages. WDYT?

fionayhlin:
@ZitaNemeckova @sbuenafe-rh I believe the no results found from search keyword ones should have a "Clear all filters" link button in the empty state under the text, but otherwise looks good!

ZitaNemeckova:
@fionayhlin We agreed with @sbuenafe-rh that filter handling should stay in filter component only for this case
Screenshot 2021-02-03 at 12 39 09 I assume it included all searches. WDYT?

ZitaNemeckova:
@himdel It's ready for another round of review. Thanks :)

src/components/repositories/remote-repository-table.tsx
```diff
@@ -69,30 +55,6 @@ export class RemoteRepositoryTable extends React.Component<IProps> {
 
   render() {
     const { remotes } = this.props;
-    if (remotes.length == 0) {
```
himdel, line 72: :+1: handled by `repository-list` now, the only user of `RemoteRepositoryTable`

himdel:
I think `EmptyStateUnauthorised` is currently missing a screenshot, and unless it's supposed to be different..

```diff
diff --git a/src/components/empty-state/empty-state-unauthorised.tsx b/src/components/empty-state/empty-state-unauthorised.tsx
index a205ca6..bef7e2e 100644
--- a/src/components/empty-state/empty-state-unauthorised.tsx
+++ b/src/components/empty-state/empty-state-unauthorised.tsx
@@ -15,9 +15,9 @@ interface IProps {}
 export class EmptyStateUnauthorised extends React.Component<IProps> {
   render() {
     return (
-      <EmptyState variant={EmptyStateVariant.xs}>
+      <EmptyState variant={EmptyStateVariant.small}>
         <EmptyStateIcon icon={LockIcon} />
-        <Title headingLevel='h4'>
+        <Title headingLevel='h4' size='lg'>
           You do not have have access to Automation Hub
         </Title>
         <EmptyStateBody>
```

(And I still think Unauthori**z**ed is the proper spelling, unless Ansible is using non-US spellings?)

---

Would it make sense to DRY it up by implementing the other empty states on top of `EmptyStateCustom`?

If we move the `{this.props.button && (...` logic there, all the other empty states could just be built on top of it, like..

```js
export const EmptyStateFilter = ({}) => (<EmptyStateCustom
  title="No resuls found"
  description="No results match the filter criteria. Remove all filters or clear all filters to show results."
  icon={SearchIcon}
/>);

export const EmptyStateNoData = ({ title, description, button }) => (<EmptyStateCustom
  title={title}
  description={description}
  icon={button ? PlusCircleIcon : SearchIcon}
  button={button}
/>);

export const EmptyStateUnauthorized = ({}) => (<EmptyStateCustom
  title="You do not have have access to Automation Hub"
  description="Contact you organization administrator for more information."
  icon={LockIcon}
/>);
```

(+- typescript and the component syntax anyway)

---

I'm seeing some icon changes, `InfoIcon` -> `SearchIcon`, `WarningTriangleIcon` -> `SearchIcon`, but those are all intentional, right? :)

And I'll continue tomorrow :)

src/utilities/filter-is-set.ts
```diff
@@ -0,0 +1,12 @@
+// Checks that at least one filter is set
+export function filterIsSet(params, filters) {
```
himdel, line 2: This could be `_.some(filters, (filter) => filter in params));`, if lodash is preferred.

ZitaNemeckova:
@himdel Icon changes are intentional based on UX requirements. `s/z` is fixed. I'll do the DRY you suggested as it's really cool! 

Unathorised:
Screenshot 2021-02-17 at 11 46 01

src/utilities/filter-is-set.ts
ZitaNemeckova: Fixed :)

ZitaNemeckova:
@himdel last review addressed. Ready for another round. Thanks :)

src/containers/namespace-list/namespace-list.tsx
```diff
@@ -140,48 +135,65 @@ export class NamespaceList extends React.Component<IProps, IState> {
...
-              sortOptions={[{ title: 'Name', id: 'name', type: 'alpha' }]}
-              searchPlaceholder={'Search ' + title.toLowerCase()}
-              updateParams={p =>
-                this.updateParams(p, () => this.loadNamespaces())
-              }
-              extraInputs={extra}
-            />
-            <div>
-              <Pagination
+          {filterIsSet(this.state.params, ['keywords']) ||
```
himdel, line 138: consider using a `noData` var as well?

src/containers/repositories/repository-list.tsx
```diff
@@ -188,36 +185,50 @@ class RepositoryList extends React.Component<RouteComponentProps, IState> {
...
+      ) : (
+        <Main className='repository-list'>
+          <Section className='body'>
+            {loading ? (
+              <LoadingPageSpinner />
+            ) : (
+              <RemoteRepositoryTable
+                remotes={content}
+                updateParams={this.updateParams}
+                editRemote={(remote: RemoteType) => {
```
himdel, line 217: `this.selectRemoteToEdit` becomes unused (and does the same thing except on a *copy* of `remote`?)

src/containers/user-management/user-create.tsx
```diff
@@ -31,17 +36,27 @@ class UserCreate extends React.Component<RouteComponentProps, IState> {
 
   render() {
     const { user, errorMessages } = this.state;
-    if (!this.context.user || !this.context.user.model_permissions.add_user) {
-      return <Redirect to={Paths.notFound}></Redirect>;
-    }
-    return (
+    const notAutthorised =
```
himdel, line 39: typo, double T

himdel:
Thanks, looks great :)

---

Is `redirect` ever set from outside the component, or can I assume `setState({ redirect: ...` only happens in the component itself?

If it's local only, would it make sense to split redirect from 404 and use a separate state param? It looks like some components (`edit-namespace`) still need to Redirect after update, so you're catching `Paths.notFound` as a special case and dealing with it differently, while others (`certification-dashboard`) only used `Paths.notFound` so the `Redirect` is removed and `redirect` is used to just mean "not there".

I'm not seeing any logic errors in the conversion :+1: , but if it's safe to split, I'd keep `redirect` just for `Redirect` and use something like `setState({ notFound: true })` instead for the unauthorized empty state.

EDIT: or `unauthorized` like in `user-edit`

---

I do notice that `filterIsSet` needs to know the list of params that *could* be filters while `AppliedFilters` seems to need a list of params to *ignore*.. sounds like the 2 lists could get out of sync fairly quickly, new params would get ignored by `filterIsSet` but not by `AppliedFilters`, we may want to fix that by having a per screen list of filers, used by both.

And perhaps the meta params could be app global, enforcing `page_size`, `id` etc. can never be treated as a filter, regardles of whether the screen is using it or not (and we could drop the `ignoredParams` list). But that's probably a separate refactoring as that's not really related to this change?

---

I can't seem to be able to rebuild now (`pkg_resources.VersionConflict: (pulp-container 2.1.0 (/venv/lib/python3.6/site-packages), Requirement.parse('pulp-container==2.3.1'))`, possibly related to https://github.com/ansible/galaxy_ng/pull/650 ?), so I have not tested the new changes yet.

src/containers/user-management/user-create.tsx
ZitaNemeckova: fixed :)

ZitaNemeckova:
@himdel

`redirect` shouldn't be set from outside. I'll fix it as suggested.

Params refactoring sounds like good idea but it's out of scope for this PR.

How to rebuild:  `galaxy_ng` must be on latest `master`, `LOCK_REQUIREMENTS=1` added  in `.compose.env` then build and run migration (to be safe).

src/containers/namespace-list/namespace-list.tsx
ZitaNemeckova: Added `noData` var. Leaving 209 as is.

src/containers/repositories/repository-list.tsx
ZitaNemeckova: I cannot remember why I did that. Fixing back at `this.selectRemoteToEdit`

src/containers/namespace-list/namespace-list.tsx
```diff
@@ -109,6 +104,8 @@ export class NamespaceList extends React.Component<IProps, IState> {
     const { namespaces, params, itemCount } = this.state;
     const { title, filterOwner } = this.props;
     const { user } = this.context;
+    const noData =
+      !filterIsSet(this.state.params, ['keywords']) && namespaces.length === 0;
```
himdel, line 108: `namespaces` can be undefined, we can't check length if it is

himdel:
Ah, managed to rebuild, `docker system prune -a --volumes` helped, seems like the build doesn't notice changes to requirements :).

---

Trying to load `/ui/namespaces` or `/ui/my-namespaces` (with a fresh db):

```
 Cannot read property 'length' of undefined
    at NamespaceList../src/containers/namespace-list/namespace-list.tsx.NamespaceList.render (App.js:12364)
```

(https://github.com/ansible/ansible-hub-ui/pull/285#discussion_r578834943)

---

`src/containers/certification-dashboard/certification-dashboard.tsx` is still using the `redirect: Paths.notFound` pattern to display `EmptyStateUnauthorized`

`src/containers/namespace-detail/namespace-detail.tsx` is still redirecting (actual redirect) to `Paths.notFound`, should it be doing that?

src/containers/namespace-list/namespace-list.tsx
ZitaNemeckova: Fixed :+1:

ZitaNemeckova:
@himdel
> `src/containers/namespace-detail/namespace-detail.tsx` is still redirecting (actual redirect) to Paths.notFound, should it be doing that?

Yes, it gets all `http://localhost:8002/ui/<ramdom stuff>` calls. This my be worth refactoring later.

himdel (approved):
LGTM, not seeing any issues in code or when testing :+1:

PR #288, opened 2021-01-12
Add Pulp API to UI

ZitaNemeckova:
Closes-Issue: AAH-255

Populate DB via console:
```
for foo in  postgres treafik mongo mariadb redis node mysql busybox alpine docker python hhtpd nginx memcached golang  ; do  docker pull $foo; docker image tag $foo localhost:5001/$foo; docker push localhost:5001/$foo; done
```

Last modified will be added in followup UI PR.

TODO:
- [x] pagination
- [x] sorting - param ordering
- [x] filter - no partial filter

### DIFF
--- a/src/api/base.ts
+++ b/src/api/base.ts
@@ -20,12 +20,6 @@ export class BaseAPI {
     return '';
   }
 
-  // Use this function to get paths in the _ui API. That will ensure the API version
-  // gets updated when it changes
-  getUIPath(url: string) {
-    return url;
-  }
-
   list(params?: object, apiPath?: string) {
     // The api uses offset/limit for pagination. I think this is confusing
     // for params on the front end, so we're going to use page/page size
--- a/src/api/base.ts
+++ b/src/api/base.ts
@@ -7,17 +7,13 @@ export class BaseAPI {
   apiPath: string;
   http: any;
 
-  constructor() {
+  constructor(apiBaseUrl) {
     this.http = axios.create({
-      baseURL: this.apiBaseURL(),
+      baseURL: apiBaseUrl,
       paramsSerializer: params => ParamHelper.getQueryString(params),
     });
 
     this.http.interceptors.request.use(request => this.authHandler(request));
-  }
-
-  apiBaseURL() {
-    return '';
   }
 
   public mapPageToOffset(p) {
--- a/src/api/base.ts
+++ b/src/api/base.ts
@@ -20,6 +20,24 @@ export class BaseAPI {
     return '';
   }
 
+  public mapPageToOffset(p) {
+    // Need to copy the object to make sure we aren't accidentally
+    // setting page state
+    const params = { ...p };
+
+    const pageSize =
+      parseInt(params['page_size']) || Constants.DEFAULT_PAGE_SIZE;
+    const page = parseInt(params['page']) || 1;
+
+    delete params['page'];
+    delete params['page_size'];
+
+    params['offset'] = page * pageSize - pageSize;
+    params['limit'] = pageSize;
+
+    return params;
+  }
+
   list(params?: object, apiPath?: string) {
     // The api uses offset/limit for pagination. I think this is confusing
     // for params on the front end, so we're going to use page/page size
@@ -50,7 +68,7 @@ export class BaseAPI {
     return this.http.patch(this.getPath(apiPath) + id + '/', data);
   }
 
-  private getPath(apiPath: string) {
+  getPath(apiPath: string) {
     return apiPath || this.apiPath;
   }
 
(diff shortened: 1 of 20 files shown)


ZitaNemeckova:
@newswangerd please have a look, thanks :)

src/api/pulp.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  constructor() {
+    this.apiPath = this.apiBaseURL;
+    this.http = axios.create({
+      baseURL: this.apiBaseURL,
+      paramsSerializer: params => ParamHelper.getQueryString(params),
+    });
+  }
+
+  list(params?: object, apiPath?: string) {
```
newswangerd, line 19: These are all copy pasted from BaseAPI. They should really be moved to a reusable file or inherited from a shared base class.

I know we were attempting to not change the base class for the full API, but I think it should be fairly easy to make a new base class that contains the `list`, `get`, `update`, `create`, `delete`, `patch`, `getPath`, `authHandler` and `mapPageToOffset` methods ad the `apiPath` and `http` properties without introducing any regressions.

You could also just move the shared methods to an `api.utils` module and import them here. That might be a little bit easier, but I think it's not quite as clean.

src/api/execution-environment.ts
```diff
@@ -0,0 +1,11 @@
...
+
+class API extends PulpAPI {
+  apiPath = 'repositories/container/container-push/';
+
+  constructor() {
+    super();
+  }
+}
+
+export const ExecutionEnvironmentAPI = new API();
```
newswangerd, line 11: I like the idea of just putting all EE related calls into a single module rather than presenting the different objects in the pulp API to the presentation layer. I'm curious to see how this will work in the final implementation,

newswangerd (changes requested):
Just need to DRY out the pulp api base.

src/api/execution-environment.ts
newswangerd: One problem you might run into is that page loading will be blocked until all the APIs resolve. This is probably fine for details, but I know on the list views we'll have to make a call for each container to get the description, and that could slow down the list page a lot. If the APIs are called separately you could use a skeleton to show that the description is still loading: https://patternfly-react.surge.sh/components/skeleton. If all the descriptions are grabbed asynchronously, latency might not be an issue though.

src/api/pulp.ts
```diff
@@ -0,0 +1,48 @@
...
+import { Constants } from '../constants';
+
+export class PulpAPI {
+  apiBaseURL = '/pulp/api/v3/';
+
+  apiPath: string;
+  http: any;
+
+  constructor() {
+    this.apiPath = this.apiBaseURL;
```
newswangerd, line 12: I'm surprised these calls work without the CSRF token getting set. I guess it probably gets set by the other API calls and reused here. At any rate, it might be worth adding the `authHandler` here.

src/api/base.ts
```diff
@@ -4,26 +4,26 @@ import { ParamHelper } from '../utilities';
...
-
-  apiBaseURL = API_HOST + API_BASE_PATH;
-
   apiPath: string;
   http: any;
 
   constructor() {
     this.http = axios.create({
-      baseURL: this.apiBaseURL,
+      baseURL: this.apiBaseURL(),
```
newswangerd, line 12: I would recommend passing this as a constructor argument instead of getting it from some function that you then have to override.

src/api/base.ts
```diff
@@ -4,26 +4,26 @@ import { ParamHelper } from '../utilities';
...
   }
 
+  apiBaseURL() {
+    return '';
+  }
+
   // Use this function to get paths in the _ui API. That will ensure the API version
   // gets updated when it changes
   getUIPath(url: string) {
-    return `_ui/${this.UI_API_VERSION}/${url}`;
```
newswangerd, line 26: This function isn't necessary for anything other than the hub endpoints, so you can remove it here and just provide it in the hub base class.

src/api/pulp.ts
```diff
@@ -0,0 +1,36 @@
...
+  apiBaseURL() {
+    return '/pulp/api/v3/';
+  }
+
+  public mapPageToOffset(p) {
+    // replace sort with ordering
+    const params = { ...p };
+    const sort = params['sort'];
+    params['ordering'] = sort;
+    delete params['sort'];
```
newswangerd, line 22: I would recommend moving `mapPageToOffset` to the base class and overriding the `list` method on the pulp class here to provide the mapping between ordering and sort.

Messing with `sort` in this function is kind of an unexpected side effect and this also makes it so that the actual code for the `mapPageToOffset` function is still duplicated in two places.

newswangerd (changes requested):

newswangerd (approved):

PR #293, opened 2021-01-22
Save null  when proxy_url is undefined(deleted)

ZitaNemeckova:
Fixes-Issue: AAH-174

Details:
If a proxy URL has been configured for a remote, it is not possible to remove it. Emptying the field and saving simply leaves the previous value in place. You can change the value, just not remove it.

### DIFF
--- a/src/api/remotes.ts
+++ b/src/api/remotes.ts
@@ -41,8 +41,9 @@ class API extends BaseAPI {
       'client_key',
       'client_cert',
       'ca_cert',
+      'proxy_url',
     ]) {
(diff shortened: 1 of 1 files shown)


newswangerd (approved):

hendersonreed:
I'm currently able to reproduce this issue, and I also identified an additional bug:

When editing the rh-certified repo, if I go in to add a proxy url, I need to put in a token. Then, if I go in to remove the proxy url, I can't save the newly edited repo settings because a red "When specifying 'auth_url' you must also specify 'token'." error appears at the bottom of the settings window.

PR #299, opened 2021-02-02
Add write only field

newswangerd:
Waiting for the following to merged/fixed:
- [x] https://github.com/ansible/galaxy_ng/pull/630
- [x] https://github.com/ansible/galaxy_ng/pull/626

Fixes: https://issues.redhat.com/browse/AAH-120

Adds a WriteOnlyField component and uses it on the remotes form to make it clearer when a user is resetting their password.

Screen Shot 2021-02-02 at 3 31 27 PM
Screen Shot 2021-02-02 at 3 31 35 PM

### DIFF
--- a/src/utilities/write-only-fields.ts
+++ b/src/utilities/write-only-fields.ts
@@ -1,0 +1,31 @@
+import { WriteOnlyFieldType } from '../api';
+
+export function isFieldSet(
+  name: string,
+  writeOnlyFields: WriteOnlyFieldType[],
+) {
+  const field = writeOnlyFields.find(el => el.name === name);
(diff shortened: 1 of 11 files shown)


src/utilities/write-only-fields.ts
```diff
@@ -0,0 +1,31 @@
+import { WriteOnlyFieldType } from '../api';
+
+export function isFieldSet(
```
newswangerd, line 3: Should these functions be part of the write only field component file, since they are meant to be used together?

ZitaNemeckova (approved):
Tested in UI. LGTM 👍

PR #301, opened 2021-02-11
Add test for AAH-160

hendersonreed:
The change in commands.js was simply a missing parenthesis, and the
change in package.json is just whitespace (seems like it was automated,
I don't remember making that change.)


ZitaNemeckova:
LGTM 👍 Thanks for catching missing paranthesis and whitespaces :)

PR #305, opened 2021-02-16
Add tests for super-user and self deletion, refactor user-dashboard.js

hendersonreed:
The tests are fairly straightforward and verify the fix for AAH-329. The refactor consisted of moving the user and group creation to a 'before' step, rather than creating and deleting the test user at the start and end of each test. This shortens the test run by a tiny bit and also makes it easier to understand what's going on where.


ZitaNemeckova (approved):
LGTM 👍 I really like your refactoring 👍

ironfroggy (approved):

PR #308, opened 2021-02-17
Add test for adding and removing remote repository credentials AAH-120.

hendersonreed:
I debated adding a command to handle setting and verifying the configuration options, but as it's currently just 1 test I think it's good the way it is. When we come back to this however, it's probably a good idea to clean up the test interface for setting remote repository config.

These tests pass currently, so feel free to merge as soon as you've reviewed.

### DIFF
--- a/test/cypress/integration/repo_management.js
+++ b/test/cypress/integration/repo_management.js
@@ -30,7 +30,7 @@ describe('Repo Management tests', () => {
         cy.get('input[id="proxy_password"]').type('test');
         cy.route('PUT', Cypress.env('prefix') + 'content/community/v3/sync/config/').as('saveConfig');
         cy.contains('Save').click();
-        cy.wait('@saveConfig');
+        cy.wait('@saveConfig').its('status').should('eq', 200);
 
         // verify values have been saved properly.
         cy.get('[aria-label="Actions"]:first').click(); // click the kebab menu on the 'community' repo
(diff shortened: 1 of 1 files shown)


test/cypress/integration/repo_management.js
```diff
@@ -15,4 +15,47 @@ describe('Repo Management tests', () => {
...
+
+        // first fill it up with data
+        cy.get('input[id="username"]').type('test');
+        cy.get('input[id="password"]').type('test');
+        cy.get('input[id="proxy_url"]').type('https://example.org');
+        cy.get('input[id="proxy_username"]').type('test');
+        cy.get('input[id="proxy_password"]').type('test');
+        cy.route('PUT', Cypress.env('prefix') + 'content/community/v3/sync/config/').as('saveConfig');
+        cy.contains('Save').click();
+        cy.wait('@saveConfig');
```
ZitaNemeckova, line 33: @hendersonreed What do you think about adding one extra check that response status is 200?

```diff
-        cy.wait('@saveConfig');
+        cy.wait('@saveConfig').its('status').should('eq', 200);
```

I don't think checking it in the form is enough because values there may be just cached and not the ones from server. So this test may pass even when `@saveConfig` returns 500.

ZitaNemeckova:
Except for one suggestion LGTM :+1:

test/cypress/integration/repo_management.js
hendersonreed: Sounds like a good idea to me, I'll add that check right now. Thanks!

hendersonreed: Added, let me know what you think.

ironfroggy (approved):

PR #312, opened 2021-03-02
Delete modal - create a shared delete modal, use for all deletions

himdel:
Fixes https://issues.redhat.com/browse/AAH-358

This unifies the confirmation modal for all deletions/removals in the app so that we're:

* using a confirmation modal,
* using patternfly title with `titleVariant='warning'` to fix the icon alignment in the header
* the title has the second word lowercased (Delete user => Delete group)
* there's a description, with the name of the deleted item in bold

It also adds a missing Delete button to the group detail screen (user detail has it),
tweaks `.gitignore` and `.prettierignore` to ignore vim swap files,
fixes a logo console warning,
and updates the Users tab in Group detail to work with 10+ users.
(I can split these off if needed :).)

**Before**:

Users > kebab Delete - icon alignment, missing bold

usersb

Users > detail > Delete - icon alignment, missing bold

userdb

Groups > Delete - icon alignment, missing description

groupsb

Groups > detail > Delete - no such button

groupdetb

Groups > detail > tab Users > kebab Remove - no confirmation modal, removed directly

---

**After**:

Users > kebab Delete

users

Users > detail > Delete

userd

Groups > Delete

groups

(see comments for a version with a list/count of users)

Groups > detail > Delete

groupdet
groupd

(see comments for a version with a list/count of users)

Groups > detail > tab Users > kebab Remove

groupuser

---

~~Question: right now, a group can be deleted even if it has users assigned to it. Should that be possible? Warning at least?~~
Resolved by showing a list of affected users.

### DIFF
--- a/src/containers/group-management/delete-group-modal.tsx
+++ b/src/containers/group-management/delete-group-modal.tsx
@@ -1,0 +1,51 @@
+import * as React from 'react';
+import { List, ListItem, Spinner } from '@patternfly/react-core';
+import { DeleteModal } from '../../components/delete-modal/delete-modal';
+
+interface IProps {
+  count?: number;
+  cancelAction: () => void;
+  deleteAction: () => void;
+  name: string;
+  users?: any[];
+}
+
+export class DeleteGroupModal extends React.Component<IProps> {
+  render() {
+    const { cancelAction, count, deleteAction, name, users } = this.props;
+
+    return (
+      <DeleteModal
+        cancelAction={cancelAction}
+        deleteAction={deleteAction}
+        title='Delete group?'
+      >
+        <b>{name}</b> will be permanently deleted.
+        <p>&nbsp;</p>
+        <div>
+          {users && count > 10 && (
+            <p>Deleting this group will affect {count} users.</p>
+          )}
+          {users && count > 0 && count <= 10 && (
+            <>
+              <p>These users will lose access to the group content:</p>
+              <List>
+                {users.map(u => (
+                  <ListItem key={u.username}>
+                    <b>{u.username}</b>
+                  </ListItem>
+                ))}
+              </List>
+            </>
+          )}
+          {users && !count && <p>No users will be affected.</p>}
+          {!users && (
+            <p>
+              Checking for affected users... <Spinner size='sm' />
+            </p>
+          )}
+        </div>
+      </DeleteModal>
+    );
+  }
+}
--- a/src/containers/group-management/delete-group-modal.tsx
+++ b/src/containers/group-management/delete-group-modal.tsx
@@ -1,13 +1,14 @@
 import * as React from 'react';
 import { List, ListItem, Spinner } from '@patternfly/react-core';
 import { DeleteModal } from '../../components/delete-modal/delete-modal';
+import { UserType } from '../../api';
 
 interface IProps {
   count?: number;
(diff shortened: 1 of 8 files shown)


himdel:
@sbuenafe-rh can you review please? :) This should be unifying all the galaxy UI removal deletion confirmation modals and aligning them with patternfly guidelines.

ZitaNemeckova (approved):
LGTM 👍 I'd like to get Susan's approval before merge.

sbuenafe-rh:
@himdel The delete user and remove user from group modals look good. 

Deleting a group's message needs to be reworked a little to mention that the users in this group will lose access. Not sure if we can list out the users...

himdel:
@sbuenafe-rh thanks :) would something like this make sense?

for a bit, if the users list is not already loaded (the blue bit is a small spinner, spinning):

spin

empty group:

none

nonempty group:

some

Any ideas how to deal with groups with 10+ users?

(WIP: right now this is only implemented on the group detail screen, and not limiting the number of users to display in the list, nor the list height)

sbuenafe-rh:
@himdel looks good! For 10+ users, we could just state that deleting this group will affect [X] users.

himdel:
Thanks, updated... :)

For up to 10 items...




Over 10...





I've also fixed the Users tab in group detail to show more than 10 items, sorted by username,
moved the group deletion view logic to a shared modal and used this from the group-list, so both group deletions should behave the same now :).

Ready for more review :)

src/containers/group-management/delete-group-modal.tsx
```diff
@@ -0,0 +1,51 @@
+import * as React from 'react';
+import { List, ListItem, Spinner } from '@patternfly/react-core';
+import { DeleteModal } from '../../components/delete-modal/delete-modal';
+
+interface IProps {
+  count?: number;
+  cancelAction: () => void;
+  deleteAction: () => void;
+  name: string;
+  users?: any[];
```
ZitaNemeckova, line 10: Use `UserType ` instead of `any`.

src/containers/group-management/group-list.tsx
```diff
@@ -48,6 +47,8 @@ interface IState {
   alerts: AlertType[];
   groups: any[];
   createModalVisible: boolean;
+  deleteModalCount?: number;
+  deleteModalUsers?: any[];
```
ZitaNemeckova, line 51: This should be `UserType` as well.

src/loaders/standalone/standalone-loader.tsx
```diff
@@ -163,15 +163,16 @@ class App extends React.Component<RouteComponentProps, IState> {
...
-        logo={
+        logo={<SmallLogo alt={APPLICATION_NAME}></SmallLogo>}
+        logoComponent={({ children }) => (
           <Link
             to={formatPath(Paths.searchByRepo, {
               repo: this.state.selectedRepo,
             })}
           >
-            <SmallLogo alt={APPLICATION_NAME}></SmallLogo>
+            {children}
```
ZitaNemeckova, line 173: 🎉

ZitaNemeckova (approved):
LGTM :+1:

sbuenafe-rh:
@himdel When checking for affected users...will the Delete button be disabled?

PR #313, opened 2021-03-03
Repository form

ZitaNemeckova:
Container Registry -> select a container -> click Edit in right top corner.

Screenshot 2021-04-13 at 16 27 46


src/api/base.ts
```diff
@@ -61,10 +61,12 @@ export class BaseAPI {
   }
 
   patch(id: string | number, data: any, apiPath?: string) {
+    console.log('PATCH');
```
newswangerd, line 64: Don't forget to remove these

src/components/execution-environment/repository-form.tsx
```diff
@@ -0,0 +1,98 @@
...
+              id='description'
+              value={description}
+              onChange={value => this.setState({ description: value })}
+              type='text'
+            />
+          </FormGroup>
+          <FormGroup key='groups' fieldId='groups' label='Groups with access'>
+            <ObjectPermissionField
+              groups={this.state.selectedGroups}
+              availablePermissions={[
```
himdel, line 78: would `Object.keys(GROUP_HUMAN_RESOURCES).filter((key) => key.startsWith('container.'))` make sense here, instead of copying the list?

himdel:
We should move the button from execution-environment-header to the top of the Detail tab, similar to https://github.com/ansible/ansible-hub-ui/pull/317.

Unless the idea is for the Edit button to work in all the tabs?

src/components/execution-environment-header/execution-environment-header.tsx
```diff
@@ -55,6 +68,15 @@ export class ExecutionEnvironmentHeader extends React.Component<
             ]}
           />
         }
+        pageControls={
+          <Button
+            variant='primary'
+            onClick={() => this.props.updateState({ editing: true })}
+          >
+            {' '}
```
himdel, line 76: the space before and after edit is probably not needed?

src/components/execution-environment/repository-form.tsx
```diff
@@ -0,0 +1,98 @@
...
+            label='Description'
+          >
+            <TextInput
+              id='description'
+              value={description}
+              onChange={value => this.setState({ description: value })}
+              type='text'
+            />
+          </FormGroup>
+          <FormGroup key='groups' fieldId='groups' label='Groups with access'>
```
himdel, line 75: expanding the select doesn't actually appear to do anything until you scroll down

it looks like setting a z-index doesn't work because the select options can't escape the grid made by the form (?) - is this solved anywhere?

(Or, `padding-bottom: 3em` works, if we don't have a handy info message to put there for padding :))

src/components/execution-environment/repository-form.tsx
```diff
@@ -0,0 +1,98 @@
...
+  availableGroups: GroupObjectPermissionType[];
+  selectedGroups: GroupObjectPermissionType[];
+}
+
+export class RepositoryForm extends React.Component<IProps, IState> {
+  constructor(props) {
+    super(props);
+    this.state = {
+      description: this.props.description,
+      availableGroups: [],
```
himdel, line 32: something is wrong here .. I don't see code to set availableGroups, or to use them
and the value in selectedGroups doesn't end up selected

Is the idea to just select a group to edit, without that selection meaning anything else?

Also, I can select the same group multiple times, but deleting one deletes all of them.

src/constants.tsx
```diff
@@ -125,6 +125,31 @@ export class Constants {
...
+    'container.namespace_push_containerdistribution':
+      'Push to existing execution environments',
+    'container.view_containernamespace': "View container's namespace",
+    'container.delete_containernamespace': "Delete container's namespace",
+    'container.namespace_delete_containerdistribution':
+      "Delete container's distribution",
+    'container.namespace_view_containerpushrepository':
+      "View container's repository",
+    'container.namespace_add_containerdistribution': 'Add distribution',
+    'container.change_containerdistribution': 'Change',
```
himdel, line 148: the last 5 buttons are probably missing "distribution" ?

src/components/execution-environment/repository-form.tsx
himdel: (also the select is a bit too wide for the trash icon to fit on the same line)

src/containers/execution-environment-detail/execution_environment_detail.tsx
```diff
@@ -76,6 +88,32 @@ class ExecutionEnvironmentDetail extends React.Component<
...
           tab='detail'
         />
+        {this.state.editing && (
+          <RepositoryForm
+            name={this.props.match.params['container']}
+            selectedGroups={this.state.groups}
+            description={this.state.description}
+            onSave={(description, selectedGroups) => {
+              let promises = [];
+              promises.push(
```
himdel, line 98: (you can just `const promises = [ ContainerDis..., Exec... ];`)

src/api/container-distribution.ts
```diff
@@ -0,0 +1,11 @@
+import { PulpAPI } from './pulp';
+
+export class API extends PulpAPI {
+  apiPath = 'distributions/container/container/';
+
+  constructor() {
```
himdel, line 6: (a constructor just calling super is exactly what you inherit if you don't mention a constructor)

src/components/execution-environment/repository-form.tsx
ZitaNemeckova: `ObjectPermissionField` conponent is handling the loading of available groups. The multiselection of same groups is known bug that we should fix.

Created https://issues.redhat.com/browse/AAH-501

src/components/execution-environment/repository-form.tsx
ZitaNemeckova: Moved to Constants.

himdel:
From the UX meeting...

* we can make the modal wider, that should help with the permission select styling and the trash icon
* there's a way for the select to expand over the modal .. https://github.com/ansible/ansible-hub-ui/blob/master/src/components/namespace-modal/namespace-modal.tsx#L151
* unify on "container" vs "execution envrionment" in the list of privileges (probably "container")
* remove some of the privileges

(A redesign of those components/dropping the modal can come later.)

---

Also, "Description" doesn't seem to get saved, and isn't displayed anywhere?

src/api/base.ts
```diff
@@ -65,6 +65,7 @@ export class BaseAPI {
   }
 
   getPath(apiPath: string) {
+    console.log(apiPath || this.apiPath);
```
newswangerd, line 68: console.log

src/components/execution-environment/repository-form.tsx
ZitaNemeckova: And you solved it :D

src/components/execution-environment/repository-form.tsx
```diff
@@ -0,0 +1,87 @@
...
+      <Modal
+        variant='large'
+        onClose={onCancel}
+        isOpen={true}
+        title={'Edit repository'}
+        actions={[
+          <Button
+            key='save'
+            variant='primary'
+            onClick={() => onSave(description, selectedGroups)}
```
newswangerd, line 50: onSave should reload the container that's displayed in the header after a successful save is complete.

src/constants.tsx
```diff
@@ -108,24 +108,44 @@ export class Constants {
...
+    delete_containerdistribution: 'Delete distribution',
+    push_containerdistribution: 'Push distribution',
+    pull_containerdistribution: 'Pull distribution',
+    view_containerdistribution: 'View distribution',
   };
+  static CONTAINER_NAMESPACE_PERMISSIONS = [
+    'namespace_delete_containerdistribution',
+    'namespace_view_containerpushrepository',
+    'change_containernamespace',
+    'namespace_view_containerdistribution',
```
newswangerd, line 144: Remove `namespace_view_containerdistribution`. That's for viewing private repos.

src/components/execution-environment/repository-form.tsx
```diff
@@ -0,0 +1,87 @@
...
+        <Form>
+          <FormGroup key='name' fieldId='name' label='Name'>
+            <TextInput id='name' value={name} isDisabled={true} type='text' />
+          </FormGroup>
+          <FormGroup
+            key='description'
+            fieldId='description'
+            label='Description'
+          >
+            <TextInput
```
newswangerd, line 68: disable this field if the user doesn't have `container.namespace_change_containerdistribution` permissions.

src/components/execution-environment/repository-form.tsx
```diff
@@ -0,0 +1,87 @@
...
+              id='description'
+              value={description}
+              onChange={value => this.setState({ description: value })}
+              type='text'
+            />
+          </FormGroup>
+          <FormGroup key='groups' fieldId='groups' label='Groups with access'>
+            <ObjectPermissionField
```
newswangerd, line 76: Disable this field if the user doesn't have `container.change_containernamespace` permissions.

src/containers/execution-environment-detail/base.tsx
```diff
@@ -0,0 +1,158 @@
...
+              </Button>
+            }
+          />
+          <Main>
+            {this.state.editing && (
+              <RepositoryForm
+                name={this.props.match.params['container']}
+                selectedGroups={this.state.selectedGroups}
+                description={this.state.repo.description}
+                onSave={(description, selectedGroups) => {
```
newswangerd, line 112: Don't send API request for fields the user hasn't changed. This will prevent things from breaking if the user has permissions to update the description but not the groups.

src/containers/execution-environment-detail/base.tsx
```diff
@@ -0,0 +1,158 @@
...
+                      },
+                    ),
+                  );
+                  promises.push(
+                    ExecutionEnvironmentNamespaceAPI.update(
+                      this.state.repo.namespace.name,
+                      { groups: selectedGroups },
+                    ),
+                  );
+                  Promise.all(promises).then(() =>
```
newswangerd, line 128: There's no error handling here. If the user submits an invalid request, the form just hangs.

newswangerd (changes requested):

src/components/execution-environment/repository-form.tsx
himdel: Yup, `availableGroups` is dead in that case :)

src/constants.tsx
```diff
@@ -125,6 +125,31 @@ export class Constants {
   static GROUP_HUMAN_PERMISSIONS = {
     change_namespace: 'Change namespace',
     upload_to_namespace: 'Upload to namespace',
+    'container.add_containernamespace': 'Create new execution environments',
```
himdel: these are still a mix of "execution environment" and "container"

src/containers/execution-environment-detail/base.tsx
```diff
@@ -72,20 +87,127 @@ export function withContainerRepo(WrappedComponent) {
...
+        });
+      }
+      return isEmpty(xorWith(original, newOne, isEqual)) && same;
+    }
+
+    private loadRepo() {
+      ExecutionEnvironmentAPI.get(this.props.match.params['container'])
+        .then(result => {
+          const repo = result;
+          ExecutionEnvironmentNamespaceAPI.get(result.data.namespace.name).then(
```
himdel, line 199: return, so that we catch errors with the catch on line 208?

src/containers/execution-environment-detail/base.tsx
```diff
@@ -98,5 +220,19 @@ export function withContainerRepo(WrappedComponent) {
 
       return 'detail';
     }
+
+    private waitForTask(task) {
+      return TaskAPI.get(task).then(async result => {
```
himdel, line 225: then + async function is a werid combination.. what about..?

```js
return TaskAPI.get(task).then((result) => {
  if (result.data.state !== 'completed') {
    return new Promise(r => setTimeout(r, 500).then(() => this.waitForTask(task));
  }
});
```

src/components/execution-environment/repository-form.tsx
```diff
@@ -0,0 +1,92 @@
...
+          </FormGroup>
+          <FormGroup
+            key='description'
+            fieldId='description'
+            label='Description'
+          >
+            <TextInput
+              id='description'
+              value={description}
+              isDisabled={this.props.permissions.includes(
```
himdel, line 70: missing `!` ?

src/components/execution-environment/repository-form.tsx
```diff
@@ -0,0 +1,92 @@
...
+              type='text'
+            />
+          </FormGroup>
+          <FormGroup key='groups' fieldId='groups' label='Groups with access'>
+            <ObjectPermissionField
+              groups={this.state.selectedGroups}
+              availablePermissions={Constants.CONTAINER_NAMESPACE_PERMISSIONS}
+              setGroups={g => this.setState({ selectedGroups: g })}
+              menuAppendTo='parent'
+              isDisabled={this.props.permissions.includes(
```
himdel, line 83: missing `!`?

src/containers/execution-environment-detail/base.tsx
```diff
@@ -72,20 +87,128 @@ export function withContainerRepo(WrappedComponent) {
...
+                      } else {
+                        this.loadRepo();
+                      }
+                    })
+                    .catch(() =>
+                      this.setState({
+                        editing: false,
+                        alerts: this.state.alerts.concat({
+                          variant: 'danger',
+                          title: "Error: changes weren't save",
```
himdel, line 156: *saved

src/containers/execution-environment-detail/execution_environment_detail.tsx
```diff
@@ -40,7 +49,7 @@ class ExecutionEnvironmentDetail extends React.Component<
   }
 
   render() {
-    return <Main>{this.renderDetail()}</Main>;
+    return this.renderDetail();
```
himdel, line 52: this change needs to happen for all the tabs, right?

Images and Activity now have..

```
Main
  ExecutionEnvironmentDetailImages
    Main
      Section
```

ZitaNemeckova: just fixing it :)

himdel (approved):
Seems like all outstanding issues are addressed :+1:

Tested in UI, I'm not seeing any new issues :+1:

newswangerd (approved):

PR #315, opened 2021-03-08
Show sync button when remote hosts are set to a third part server.

newswangerd:
Fixes: https://issues.redhat.com/browse/AAH-408


awcrosby (approved):
just from looking at syntax, looks alright

PR #317, opened 2021-03-10
Group detail - Permission tab UX changes

himdel:
Fixes https://issues.redhat.com/browse/AAH-364

This contains multiple UX changes to the Group detail screen, mostly in the Permissions tab:

* moved the Edit, Save and Cancel buttons from page header to tab header, as these are only related to the tab
* keeps the Delete button as is, but disables it when editing a group, and no longer hides it on the users tab
* makes the tabs appear disabled (cursor + no action on click) when editing Permissions
* sorts the list of available permissions in the Permission selects (note this makes the order unstable between translations once we have i18n)
* use plain text styling (remove gray blocks behind input text) unless in edit mode (this is from https://issues.redhat.com/browse/AAH-360 but consistency)

Cc @ZitaNemeckova, @sbuenafe-rh, this should be ready for review :)

---

Permissions tab in view mode (before and after):

perm false before
perm false after

Permissions tab in edit mode (before and after):

perm true before
perm true after

(the tab header also switches to the `not-allowed` cursor when hovering in edit mode)

List of group permissions when editing (before and after):

perm true2 before
perm true2 after

Users tab (before and after):

user before
user after

### DIFF
--- a/src/containers/group-management/group-detail.tsx
+++ b/src/containers/group-management/group-detail.tsx
@@ -301,7 +301,8 @@ class GroupDetail extends React.Component<RouteComponentProps, IState> {
                     )
                     .map(value =>
                       twoWayMapper(value, Constants.HUMAN_PERMISSIONS),
-                    )}
+                    )
+                    .sort()}
                   selectedPermissions={selectedPermissions
                     .filter(selected =>
                       group.object_permissions.find(perm => selected === perm),
--- a/src/containers/group-management/group-detail.tsx
+++ b/src/containers/group-management/group-detail.tsx
@@ -194,41 +194,23 @@ class GroupDetail extends React.Component<RouteComponentProps, IState> {
 
   private renderControls() {
     const { user } = this.context;
+    const { editPermissions } = this.state;
 
-    if (this.state.params.tab == 'users') {
+    if (!user || !user.model_permissions.delete_group) {
       return null;
     }
-    return this.state.editPermissions ? (
+
+    return (
       <ToolbarItem>
         <Button
-          onClick={this.actionSavePermissions}
+          isDisabled={editPermissions}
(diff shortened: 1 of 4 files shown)


src/containers/group-management/group-detail.tsx
```diff
@@ -299,7 +301,8 @@ class GroupDetail extends React.Component<RouteComponentProps, IState> {
                     )
                     .map(value =>
                       twoWayMapper(value, Constants.HUMAN_PERMISSIONS),
-                    )}
+                    )
+                    .sort()}
```
ZitaNemeckova, line 305: Or sort it at `src/constants.tsx`. WDYT?

ZitaNemeckova:
- Disabled tab has a text like "You must finish editing"
- Permissions in view mode should be labels
Screenshot 2021-03-10 at 10 55 54
@sbuenafe-rh WDYT?

LGTM code-wise 👍

sbuenafe-rh:
@himdel Looks good, but I would move the Save & Cancel to the bottom left (just to be consistent with the ReadMe Markdown editor)

src/containers/group-management/group-detail.tsx
himdel: Ah, good idea :+1: This would make the order stable accross translations, so I suppose it depends on that :)

himdel:
Updated to match the labels screenshot, thanks! :)
And added a tab title when disabled, except "you must" felt a bit strong, wdyt?..

IMG_20210311_014041

And moved the Save & Cancel buttons down, and changed to primary+secondary (from default+link) as in the editor (is that what we want to settle on?);  leaving the Edit button as is..

savecancel

ZitaNemeckova (approved):
LGTM :+1:

PR #319, opened 2021-03-10
Support repo-root-relative import paths

himdel:
This adds `resolve.alias` to the webpack config, and `compilerOptions.baseUrl` + `compilerOptions.paths` to tsconfig,
allowing for JS/TS imports relative to the repository root (as long as they live under `src/`).

(The second commit just updates most imports, I skipped the `./` ones and only changed the ones with `../`, can be trivially regenerated.)

---

Cc @newswangerd , @ZitaNemeckova WDYT? :)

Does it make sense to go with `src` as the alias? It fits the path so it seems like a reasonable default,
except there *is* a `src` npm package (we don't depend on it, and it's related to redis, so, unlikely in frontend JS), so we *may* want to consider something a bit more exotic like `^/src` or `@/src`?

### DIFF
--- a/config/webpack-ts-overrides.js
+++ b/config/webpack-ts-overrides.js
@@ -1,4 +1,5 @@
 const MiniCssExtractPlugin = require('mini-css-extract-plugin');
+const { resolve } = require('path');
 
 module.exports = {
   devtool: 'source-map',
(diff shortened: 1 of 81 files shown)


newswangerd (approved):
this is fantastic!

PR #322, opened 2021-03-12
Update to match group detail page changes from pull #317, AAH-364

hendersonreed:
Changes in the formatting for the group detail page required an update in the group management tests (changing a jquery selector). 

Also found that one of the tests was failing when adding permissions due to the dropdown covering up the 'save' button, so added a click on a nearby element to close the dropdown so we can then hit 'save'.

### DIFF
--- a/test/cypress/integration/group_management.js
+++ b/test/cypress/integration/group_management.js
@@ -34,10 +34,17 @@ describe('Hub Group Management Tests', () => {
         cy.login(adminUsername, adminPassword);
         cy.createGroup(name);
         cy.contains(name).should('exist');
+
         cy.addAllPermissions(name);
-        permissionTypes.forEach(permGroup => cy.get(`.pf-l-flex.pf-m-align-items-center.${permGroup}  [placeholder="No permission"]`).should('not.exist'));;
+        permissionTypes.forEach((permGroup) => {
+            cy.get(`.pf-l-flex.pf-m-align-items-center.${permGroup}`).contains('span', 'No permission').should('not.exist');
+        });
+
         cy.removeAllPermissions(name);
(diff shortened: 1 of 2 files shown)


himdel (approved):
:+1:

#317 changed the view mode from readonly form to Labels, which do end up as spans

himdel:
@hendersonreed feel free to merge if this works for you :) 

For me...

locally, via `npm run test` I'm getting `3 of 5 failed (60%)                      01:10       13        6        3        -        4  ` even on master (and it creates root-owned dirs in my test/cypress/)

while running via `npm ci && node_modules/.bin/cypress run`  gives `3 of 5 failed (60%)                      00:54       13        4        5        -        4  ` (so, 4 passing instead of 6)

and running with `cypress open` seems visually better, but often fails because it doesn't try to clean up preventively, and doesn't clean up on failure, so there are lot of failures like testGroup already exists, test user is taken, etc.

hendersonreed:
Hey! So Cypress tends to leave dangling state after a test fails (this is a deliberate design decision by the developers of the tool, since it makes it easier to diagnose what happened.) 

So after a test fails you may need to go back in and clean it up in order to rerun and verify that it works.

The other thing is that I develop the tests assuming a clean environment, so some of the jquery selectors that are used to locate elements might not work if there is unexpected data in the tests.

I'm going to go ahead and double-check that they're passing on my machine (pretty sure I wouldn't have pushed changes that didn't pass though.) I'd also love to talk at some point with you and @ZitaNemeckova and @ironfroggy about maybe revamping some of the test setup/teardown steps to make them a bit less fragile.

hendersonreed:
I just went ahead and double-checked and on a clean environment the tests are all passing, so I think it's just a feature of the way we have built our tests (not currently much preventative cleanup happening, only post-cleanup.)

Definitely an area for us to work on, but I'd still like to merge this in so at least we have passing tests on master.

ZitaNemeckova:
Bit late but it's passing in my clean env. as well.

PR #323, opened 2021-03-15
Fix dates to have Tooltip with exact date

ZitaNemeckova:
Dates should look like this if possible:

Screenshot 2021-05-03 at 13 02 32

### DIFF
--- a/src/components/date-component/date-component.tsx
+++ b/src/components/date-component/date-component.tsx
@@ -1,0 +1,23 @@
+import * as React from 'react';
+import { Tooltip } from '@patternfly/react-core';
+import * as moment from 'moment';
+
+interface IProps {
+  date: string;
+  isRow?: boolean;
+}
+
+export class DateComponent extends React.Component<IProps> {
+  render() {
+    const { date, isRow } = this.props;
+    return (
+      <Tooltip content={moment(date).format('MMMM Do YYYY')}>
+        {isRow ? (
+          <td>{moment(date).fromNow()}</td>
+        ) : (
+          <div>{moment(date).fromNow()}</div>
+        )}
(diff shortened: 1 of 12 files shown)


src/components/date-component/date-component.tsx
```diff
@@ -0,0 +1,23 @@
...
+  isRow?: boolean;
+}
+
+export class DateComponent extends React.Component<IProps> {
+  render() {
+    const { date, isRow } = this.props;
+    return (
+      <Tooltip content={moment(date).format('MMMM Do YYYY')}>
+        {isRow ? (
+          <td>{moment(date).fromNow()}</td>
```
himdel, line 16: I'm not sure DateComponent should know about tables, would it work to just `<td><DateComponent.../></td>`?

src/components/date-component/date-component.tsx
```diff
@@ -0,0 +1,23 @@
...
+interface IProps {
+  date: string;
+  isRow?: boolean;
+}
+
+export class DateComponent extends React.Component<IProps> {
+  render() {
+    const { date, isRow } = this.props;
+    return (
+      <Tooltip content={moment(date).format('MMMM Do YYYY')}>
```
himdel, line 14: Missing a time and timezone? (if the data is available :))

ZitaNemeckova: @sbuenafe-rh How should it looklike in tooltip? Thanks :)

sbuenafe-rh:
@ZitaNemeckova the format would be DD MMM YYYY, and if available the timestamp. For example, **07 Jan 2019, 23:33 UTC**

newswangerd (approved):

PR #324, opened 2021-03-15
User detail/form UX changes

himdel:
Fixes https://issues.redhat.com/browse/AAH-360

> Use plain text styling (remove gray blocks behind input text) unless in edit mode
> Move edit/delete buttons to header
> Reorder info to match table column order (see attachments for orders)
> > Order of info confusing. Previous page table shows username first + breadcrumb also shows username, but here username is hidden in the middle of the page -> reorder to match?

This moves the "render form fields based on definition" logic out of UserForm to a separate DataForm component,
changes the readonly view to simplify the formgroups, and only render the text value instead of an input,
removes the password & password confirmation fields in read only mode,
pushes the username field to the top,
and moves the Edit/Delete buttons to page header.

Cc @ZitaNemeckova , @sbuenafe-rh this should be ready for review :)

That said, 2 major questions..

* how *should* the view only mode actually look?, in Group we put the list of labels on the same line as the field name, so I'm not sure which way to go.
* the order of fields in list view is username, email, last name, first name, groups, created  
the order in detail/form now is username, first name, last name, email, passwords, groups  
I don't think it makes sense to go with "last name, first name" in edit, that would confuse everybody. And not sure if last name, first name in the list is intentional. So.. should we have the same order? Which one?

---

Affected screens:

Users > detail
Users > detail > Edit
Users > Create user
navbar > My Profile
navbar > My Profile > Edit

---

Screenshots...

Create user (before and after)
create-before
create-after

My profile - edit (before and after)
myedit-before
myedit-after

My profile - view (before and after)
myprof-before
myprof-after

User detail (view) (before and after)
uder-before
uder-after

User edit (before and after)
uedt-before
uedt-after

### DIFF
--- a/src/components/shared/data-form.tsx
+++ b/src/components/shared/data-form.tsx
@@ -39,21 +39,24 @@ export class DataForm extends React.Component<IProps> {
         <FormGroup
           fieldId={field.id}
           helperTextInvalid={errorMessages[field.id]}
-          isRequired={requiredFields.includes(field.id)}
+          isRequired={!isReadonly && requiredFields.includes(field.id)}
           key={field.id}
           label={field.title}
-          labelIcon={field.formGroupLabelIcon as any}
-          validated={validated}
+          labelIcon={!isReadonly && (field.formGroupLabelIcon as any)}
+          validated={isReadonly ? 'default' : validated}
         >
-          <TextInput
-            id={field.id}
-            isDisabled={isReadonly}
-            onChange={updateField}
-            placeholder={field.placeholder}
-            type={(field.type as any) || 'text'}
-            validated={validated}
-            value={model[field.id]}
-          />
+          {isReadonly ? (
+            model[field.id]
+          ) : (
+            <TextInput
+              id={field.id}
+              onChange={updateField}
+              placeholder={field.placeholder}
+              type={(field.type as any) || 'text'}
+              validated={validated}
+              value={model[field.id]}
+            />
+          )}
         </FormGroup>
       );
     });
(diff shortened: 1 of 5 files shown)


sbuenafe-rh:
@himdel for the view mode on the Groups, you can put the chips underneath the field label. When it's in edit mode, the dropdown select will be available with the chips that have been already selected in the field. 

For consistency:
Create user form: Username, First name, Last name, Email, Password, Groups
User list: Username, First name, Last name, Groups

ZitaNemeckova:
@himdel LGTM code-wise 👍 Feel free to merge once Susan's feedback is addressed.

ZitaNemeckova:
Something like this to fix the tests.
```diff
diff --git a/test/cypress/integration/profile.js b/test/cypress/integration/profile.js
index 6064c26..edf0b85 100644
--- a/test/cypress/integration/profile.js
+++ b/test/cypress/integration/profile.js
@@ -10,7 +10,8 @@ describe('My Profile Tests', () => {
         cy.get('[aria-label="user-dropdown"] button').click();
         // a little hacky, but basically
         // just click the one link that says 'My profile'.
-        cy.get('a').contains('My profile').click(); 
+        cy.get('a').contains('My profile').click();
+        cy.get(' button:contains("Edit")').click();
     });
 

```

himdel:
Thanks!

Updated the view mode to show Groups the same way as on the list screen..

udet

Updated the list view to match the Username, First name, Last name, Email, Groups ordering. (And added a missing space before the superuser tag.)

ulist

Aand updated the tests as suggested :)

ZitaNemeckova (approved):
LGTM 👍

PR #331, opened 2021-03-23
Collection-detail: add Requires Ansible

himdel:
fixes https://issues.redhat.com/browse/AAH-410

This adds a "Requires Ansible" row to collection detail, defaulting to "No" if not present.

When present.. (last line)

requires-yes

When absent.. 

not displayed

---

~~Question: should that be a "No" or more something like "Any"?~~ ~~(updated to "unspecified")~~ Updated to nothing :)

### DIFF
--- a/src/api/response-types/collection.ts
+++ b/src/api/response-types/collection.ts
@@ -34,6 +34,7 @@ export class CollectionVersionDetail extends CollectionVersion {
     issues: string;
     repository: string;
   };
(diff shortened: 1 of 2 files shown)


newswangerd (approved):

sbuenafe-rh:
@himdel If the author doesn't provide a "Requires Ansible" value then I would not show the Requires Ansible field.

PR #334, opened 2021-03-29
Add rate limit to remotes form.

newswangerd:
Requires: https://github.com/ansible/galaxy_ng/pull/707

### DIFF
--- a/src/components/repositories/remote-form.tsx
+++ b/src/components/repositories/remote-form.tsx
@@ -537,6 +537,28 @@ export class RemoteForm extends React.Component<IProps, IState> {
               }
             />
           </FormGroup>
+          <FormGroup
+            fieldId={'rate_limit'}
+            label={'Rate Limit'}
+            labelIcon={
+              <HelperText content='Limits total download rate in requests per second.' />
+            }
+            validated={
+              Number.isInteger(remote.rate_limit) || remote.rate_limit === null
+                ? 'default'
+                : 'error'
+            }
+            helperTextInvalid={'Must be an integer.'}
+          >
+            <TextInput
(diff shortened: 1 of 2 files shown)


src/components/repositories/remote-form.tsx
```diff
@@ -537,6 +537,28 @@ export class RemoteForm extends React.Component<IProps, IState> {
...
+                : 'error'
+            }
+            helperTextInvalid={'Must be an integer.'}
+          >
+            <TextInput
+              id='rate_limit'
+              type='number'
+              value={remote.rate_limit}
+              onChange={value =>
+                this.updateRemote(parseInt(value), 'rate_limit')
```
himdel, line 558: `parseInt(value, 10)`? Unless we want to support 0x, etc..

src/components/repositories/remote-form.tsx
```diff
@@ -537,6 +537,28 @@ export class RemoteForm extends React.Component<IProps, IState> {
...
             />
           </FormGroup>
+          <FormGroup
+            fieldId={'rate_limit'}
+            label={'Rate Limit'}
+            labelIcon={
+              <HelperText content='Limits total download rate in requests per second.' />
+            }
+            validated={
+              Number.isInteger(remote.rate_limit) || remote.rate_limit === null
```
himdel, line 547: Does this work? `parseInt` returns a number of NaN, never null, and I don't see any NaN -> null conversion in `updateRemote`?

newswangerd: I just removed the `parseInt`. The API doesn't care if the the value is a string or a number, just that it can be converted into a number.

PR #346, opened 2021-04-14
Handle errors when adding user to group

himdel:
Fixes https://issues.redhat.com/browse/AAH-209

Added an `addAlert` method, used for all existing alerts, and added more request error handling to `group-detail`,
Also unsetting the selected value in the add modal on close,
and add a permission check for Add on the empty state screen.

---

```
109:    GroupAPI.get(this.state.params.id)
115:    GroupAPI.getPermissions(this.state.params.id)
```

if loading data fails on the intial load, replaced infinispinner with an empty screen with alerts..
badgroup

---

```
404:            UserAPI.list({ username__contains: name, page_size: 5 })
529:    UserAPI.list()
```

failure when autosuggesting groups

----

```
518:        return UserAPI.update(id.toString(), {
738:    UserAPI.list({
```

when saving from add modal

(modal now closes on failure, maybe it shouldn't?)

---

```
756:    UserAPI.update(user.id, user)
```

when removing user from group

(modal *doesn't* close on failure, maybe it should?)

### DIFF
--- a/src/containers/group-management/group-detail.tsx
+++ b/src/containers/group-management/group-detail.tsx
@@ -562,12 +562,15 @@ class GroupDetail extends React.Component<RouteComponentProps, IState> {
           title={'No users yet'}
           description={'Users will appear once added to this group'}
           button={
-            <Button
-              variant='primary'
-              onClick={() => this.setState({ addModalVisible: true })}
-            >
-              Add
-            </Button>
+            !!user &&
+            user.model_permissions.change_group && (
+              <Button
+                variant='primary'
+                onClick={() => this.setState({ addModalVisible: true })}
+              >
+                Add
+              </Button>
+            )
           }
         />
       );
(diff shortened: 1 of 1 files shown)


himdel:
@ZitaNemeckova, @newswangerd  can you review please?

Do we have any whole screen error screen that could be used when loading the group fails?
Have we dealt with the modals vs alerts issue? Should I just try to raise the alerts over the backdrop?

And we may want to consider making tasks to go through every file and find these API calls without catch if there's going to be more of us :)

himdel:
Also, I've been testing with network error (no webpack-dev-server), 504 (no api), 404 (no record), but not with any errors API could return, is there a convention? (As long as those end up in `exception.message`, they'll get displayed, but maybe that needs a bit lof logic to put it there?)

ZitaNemeckova:
@himdel 404 should redirect to NotFound page. Especially GET requests.

newswangerd (approved):

himdel:
Merged for now, @ZitaNemeckova please ping me with the details of https://github.com/ansible/ansible-hub-ui/pull/346#issuecomment-823129449 once you're back from PTO :).

I see *some* specific 404 handling (edit namespace, delete user modal), but that doesn't redirect anywhere.
And not sure what you meant by the loading bit (this is just removing an unused `loading` variable) :).

PR #349, opened 2021-04-15
Load feature flags from API.

newswangerd:
Disable container registry views if the container registry api is disabled.

### DIFF
--- a/src/loaders/standalone/routes.tsx
+++ b/src/loaders/standalone/routes.tsx
@@ -31,33 +31,66 @@ import {
   ExecutionEnvironmentDetailImages,
   ExecutionEnvironmentManifest,
 } from 'src/containers';
-import { ActiveUserAPI } from 'src/api';
+import {
+  ActiveUserAPI,
+  FeatureFlagsAPI,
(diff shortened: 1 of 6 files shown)


src/loaders/standalone/routes.tsx
```diff
@@ -31,33 +31,66 @@ import {
...
+    // This component is mounted on every route change, so it's a good place
+    // to check for an active user.
+    const { user } = this.context;
     if (!user) {
       ActiveUserAPI.getUser()
-        .then(result => {
-          setUser(result);
-          this.setState({ isLoading: false });
+        .then(userResponse => {
+          FeatureFlagsAPI.get().then(flagsResponse => {
```
himdel, line 89: would it make sense to return the promise, so that the parent catch catches also featureflagsapi errors?

himdel (approved):
Tested, all the container urls 404 and the Container Registry menu item disappears when I change `GALAXY_FEATURE_FLAGS` in `galaxy_ng/app/settings.py` to say `'execution_environments': False,` :).

And not seeing any change when True :+1:

PR #356, opened 2021-04-21
Remove dead code from import-list

ZitaNemeckova:
Delete dead code. No issue.


himdel (approved):
:+1:, doesn't look they were ever used.

PR #357, opened 2021-04-21
Add spinner to Container list view

ZitaNemeckova:
Fixes https://issues.redhat.com/browse/AAH-506

Go to Container Registry and check what you see before loading is done.

Before:
Empty state before containers are loaded
After:
Spinner before containers are loaded


himdel (approved):
LGTM :)

PR #359, opened 2021-04-21
Add spinner to Repo Management

ZitaNemeckova:
Go to Repo Management.

Before:
There's an empty statepage before data loads.
After:
There's a spinner before data loads.

### DIFF
--- a/src/containers/repositories/repository-list.tsx
+++ b/src/containers/repositories/repository-list.tsx
@@ -152,7 +152,8 @@ class RepositoryList extends React.Component<RouteComponentProps, IState> {
           />
         )}
         <BaseHeader title='Repo Management'>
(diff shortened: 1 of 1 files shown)


himdel:
`renderContent` already does do the spinner in both the local & remote branches .. but the empty check was added *before* the loading check in the Remote tab.

So, better fix it there? Or remove the spinner from `renderContent` and keep your change around it?

ZitaNemeckova:
@himdel Thanks :) I removed it from render to make it more consistent with other pages.

himdel:
Makes sense :) But unless github cache is deceiving me, maybe you forgot to push the update? ;)

ZitaNemeckova:
@himdel I didn't even commit it locally 😮  Fixed 😸

himdel (approved):
LGTM :+1: 

(`itemCount` is only assigned to now, so that could be removed complately, but looks good :))

PR #362, opened 2021-04-21
Update tests to handle recent changes to UI

hendersonreed:
Just a few updates to track some of the changes that have been merged into the UI over the last little while.

Had to disable one of the repo management tests that verifies we can save and delete info from the proxy configuration - now that we no longer bundle in a requirements.yml, we need to figure out some way to upload one in the future.

I looked around a little bit and it's not totally obvious that it's supported by cypress (might require the addition of a plugin). So if we elect to do that we should make an issue and track it. Until then, this fixes the other two failures I saw.


ZitaNemeckova (approved):
Tests are running. LGTM 👍

PR #395, opened 2021-05-04
Container detail fixes

ZitaNemeckova:
Fixes https://issues.redhat.com/browse/AAH-250?focusedCommentId=16124143&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16124143

- [ ] On the Read Me edit view, the edit field is very short. It only shows 3 or 4 lines and cannot be resized.

- [ ] On the Read Me edit view, the preview can become too wide and will squish the edit field even smaller.

- [ ] On the Read Me edit view, the markdown textarea's label is not a proper label (accessibility issue)

- [ ] Activity tab is cut off at 10 events without pagination, a way to load more, or any indication that this is not a complete list.

- [ ] In the image tab, the tags for a given image have no order.

- [ ] In the image tab, the tags column can become too large (especially when showing all tags) and shrinks the remaining columns in the table.

- [ ] In the image tag edit dialog, if you try to add a tag that already exists on an image, you get an error saying "
This tag already exists on another image. Do you want to move it to this image?"** which is both incorrect and misleading.**


src/components/markdown-editor/markdown-editor.scss
```diff
@@ -22,10 +22,11 @@
...
 
   .editor {
-    flex-shrink: 0;
+    min-width: 500px;
+    max-width: 500px;
   }
 }
 .resources-editor {
-  height: 500px;
+  height: 500px !important;
```
newswangerd, line 30: is there any way to do this without using `!important`?

ZitaNemeckova: used `id` instead. Fixed.

src/components/markdown-editor/markdown-editor.tsx
```diff
@@ -24,7 +24,7 @@ export class MarkdownEditor extends React.Component<IProps, {}> {
               <FormGroup fieldId='resources' helperText={helperText}>
                 Raw Markdown
                 <TextArea
-                  className='resources-editor'
+                  aria-label={'Markdown editor'}
```
ironfroggy, line 27: If we have a visible label, "Raw Markdown", the label should be marked with a <label> tag instead of an aria-label attribute.

ironfroggy (changes requested):
The repository description field has no length limit,  but the display of it is limited to too lines. Either the description needs a limit or the display of it needs to be flexible, or probably both.

ZitaNemeckova:
@ironfroggy I cannot limit a length of a description as it may be already set before it's pulled and pushed. I changed the form to have resizable input for description to make it more user friendly. In the view it's done according to UX mocks (limited to two lines with tooltip that shows the whole description).

ironfroggy (approved):

PR #396, opened 2021-05-05
[WIP] Namespace, my namespaces merge

himdel:
Fixes https://issues.redhat.com/browse/AAH-487

..todo remove rest of routes/move?


himdel:
One more update..



the list screen now makes the link clickable,
and uses the footer to show a label for ediitable namespaces

newswangerd:
I'm thinking we should get rid of the `My namespace` label until we have a better way to check permissions on objects under the namespaces endpoint. There isn't an efficient way to correlate items on the my namespaces list when the number of items mismatches. 
Screen Recording 2021-05-11 at 10 39 11 AM mov

himdel:
Ah, you're right, the two api approach fails with paging. Which makes the Edit also impossible now.

I just pushed the version without the last commit, so this no longer changes the cards on the list screen, and only queries both APIs  on the detail screen.

src/components/patternfly-wrappers/link-tabs.tsx
```diff
@@ -0,0 +1,42 @@
+import * as React from 'react';
+import cx from 'classnames';
+
+import { Link } from 'react-router-dom';
+
+interface IProps {
+  /** List of tabs */
+  tabs: {
```
newswangerd, line 8: This interface should be exported so that other components can use it correctly for type checking.

newswangerd (approved):

himdel:
The new state of empty state:

PR #422, opened 2021-05-11
Add horizontal scroll to documentation

ZitaNemeckova:
Fixes https://issues.redhat.com/browse/AAH-599

For https://cloud.redhat.com/ansible/automation-hub/repo/published/ibm/power_ibmi/content/module/ibmi_display_fix
Note: it looks really bad in `standalone`. In c.r.com you cannot read everything but left nav is visible.

Steps to reproduce:

- Create ibm namespace
- Upload https://cloud.redhat.com/ansible/automation-hub/repo/published/ibm/power_ibmi
- Go to that collection 
- Select Documentation
- Select Modules
- Select `ibmi_display_fix`

Before:
Screenshot 2021-05-11 at 8 58 57

After:
Screenshot 2021-05-11 at 9 19 55
Screenshot 2021-05-11 at 9 20 39

### DIFF
--- a/src/containers/collection-detail/collection-detail.scss
+++ b/src/containers/collection-detail/collection-detail.scss
@@ -29,6 +29,7 @@
     flex-grow: 1;
     padding: 24px;
     padding-top: 0px;
+    overflow-x: auto;
 
     // make the height close to the height ofthe page so short docs don't
     // look weird


himdel:
If the idea is to just make sure the user can scroll to see the whole table, :+1:, adding approve.

That said, I think the bug is really elsewhere.. galaxy_ng is the only place that generates a table that's too big to show.
The same content can be seen in the original [documentation](https://ibm.github.io/ansible-for-i/modules/ibmi_display_fix.html) and on [github](https://github.com/IBM/ansible-for-i/blob/devel/docs/source/modules/ibmi_display_fix.rst), neither of which needs wrapping.

himdel (approved):
The code is right if the bug is really considered to be in the UI.
EDIT: it is :+1: 

(We could also set max-width and rely on wrapping instead of scrollbars, except that may break horribly on phones.)

PR #428, opened 2021-05-11
Add backported-* label on stable-* merges

himdel:
runs on push to a `stable-\*` branch,
takes the first line of the topmost commit message:

> Original PR title (#original PR number) (#backport PR number)

extracts the first parenthesised PR number, as long as there are 2
and adds a `backported-${branch number}` label to the original PR

Cc @ZitaNemeckova , @newswangerd 

Also relevant: https://github.com/sanitizers/patchback-github-app/issues/12

### DIFF
--- a/.github/workflows/backported-labels.yml
+++ b/.github/workflows/backported-labels.yml
@@ -5,8 +5,6 @@ on:
   workflow_dispatch:
   push:
     branches: [ 'stable-*' ]
-  pull_request: # debug
-    branches: [ 'master' ] # debug
 
 jobs:
   labels:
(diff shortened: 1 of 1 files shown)


newswangerd:
@himdel does this work with the backports created by patchback?

himdel:
Yup, it should just watch the stable-X.Y branch for new pushes, and when something is pushed, it parses the PR number from that. So as long as patchback pushes things there, it should work :)

himdel:
So, this is *probably* ready...

Running with debugging overrides (to assume branch other than master) in https://github.com/ansible/ansible-hub-ui/runs/3042394137, it succeeded in everything apart from actually adding the label (as in both the LABEL and the PR values make sense given the overrides).

Both `gh` (CLI) and [action-add-labels](https://github.com/actions-ecosystem/action-add-labels) fail the same way, `Resource not accessible by integration`.

I'm *hoping* this is because as a PR, it's still running in the context of my repo, and won't have these problems when triggered by a stable push, but it may still fail, we'll see.

newswangerd (approved):
I guess we just have to merge this and see if it works.

himdel:
Looks like it works :)

4.2 https://github.com/ansible/ansible-hub-ui/runs/3058396169?check_suite_focus=true
4.3 https://github.com/ansible/ansible-hub-ui/runs/3058476222?check_suite_focus=true

:+1:

PR #452, opened 2021-05-18
Cypress - add debugging info to workflow

himdel:
Another attempt at addressing sporadic github actions cypress failures.

This is either a waiting problem, or a caching problem, but one that doesn't seem to happen outside github actions (from what I can tell).

Trying to add a bit more output to tell what's happening,
and busting the cache, by renaming it.


ZitaNemeckova (approved):
LGTM :+1:

himdel:
Merging together with https://github.com/ansible/ansible-hub-ui/pull/453.

#453 fixes the rq vs click version issue,
#452 busts the old cache so that the change applies to master as well, and adds more debug info

PR #454, opened 2021-05-18
Cypress - fix mkdir bug

himdel:
Introduced in #452, the tests fail on a 500, caused by moving the UI to the wrong place.
The extra mkdir changes the meaning of the mv following it, undoing that part, not needed now that the container doesn't fail setup.


ZitaNemeckova (approved):
LGTM 👍

PR #456, opened 2021-05-19
Fix tag failed message

newswangerd:
Don't know what I was thinking with the existing message.


himdel (approved):

himdel:
(restarting cypress, it should pass now)

PR #463, opened 2021-05-20
Remove padding & reorder install version-installation

ShaiahWren:
Screen Shot 2021-05-22 at 10 26 19 AM

HI Zita:)

Tackling a couple more issues here:
https://issues.redhat.com/browse/AAH-550
https://issues.redhat.com/browse/AAH-555

The tag.scss file with the margin is now restored. Also added back the className 'tag' from the tag.tsx file. I then went into the namespace-list.scss and removed the padding-Left on --pf-c-toolbar__content.  Hopefully it looks right now!

The next issue was a simple reordering of the Installation - Install Version, GridItems.

Screen Shot 2021-05-20 at 3 31 09 PM


ZitaNemeckova (changes requested):
@ShaiahWren Reordering looks good! 

But I don't think we should remove `tags` class to fix AAH-555. We need to go from this
Screenshot 2021-05-21 at 10 44 58
to this
Screenshot 2021-05-21 at 10 44 47

It seems that class `.pf-c-toolbar__content` is what setting that extra padding.

ZitaNemeckova (approved):
LGTM 👍

ShaiahWren:
pr closed.

PR #464, opened 2021-05-20
Test the namespace page modes with galaxykit test data

ironfroggy:
**NOTE: Since this is the first PR using galaxykit, I'm asking for more feedback from everyone. Please let me know what you think about the integration. I do plan to add better helpers, but you can always call out to the galaxykit command directly like these tests do now.**

Also NOTE: This is WIP, I'm adding more test cases ASAP.

Also Also NOTE:  I need to fix the GitHub Action stuff so I can use galaxykit and the tests can pass here and not just locally. If you want to run it locally, you need to install galaxykit from this branch until the next release: https://github.com/hendersonreed/galaxykit/pull/7

Test the namespace page modes after setting up test data via the new galaxykit helper.

Thank you for a new Pull Request!

Please, make reviewers work easier by checking, if you are not introducing any technical debt:
https://github.com/ansible/ansible-hub-ui/blob/master/developer_guidelines.md

Describe the rationale and use case for this pull request.  Provide any background, examples, and images that provide further information to accurately describe what it is that you are adding to the repo.  Add subsections as necessary to organize and feel free to link and reference other PRs as necessary, but also include them in the links section below as a quick reference.

Guidelines:
* Keep Pull Request titles short and to the point, ideally under 72 characters
* Provide as much context/info in the description as necessary to get the reviewer up to the same domain knowledge level as yourself
* Keep code changes as short as possible and implementing a single feature/fix/refactoring, when possible

Links [Optional]
----------------

* http://documentation.for/library/that/I/am/adding
* [PR relevant issue or pull_request](#123)

If there are any manual steps that you would like the reviewer(s) to take to verify your changes, please describe in detail the steps to reproduce the features added by the pull request, or the bug before and after the change.

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -22,7 +22,7 @@ jobs:
 
     - name: "Install galaxykit dependency"
       run: |
-        pip install git+https://github.com/hendersonreed/galaxykit.git@aah-487-namespace-and-cli
+        pip install galaxykit.git==0.1.0
 
     - name: "Set env.SHORT_BRANCH, env.GALAXY_NG_COMMIT"
       run: |
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -20,6 +20,10 @@ jobs:
 
     steps:
 
+    - name: "Install galaxykit dependency"
+      run: |
+        pip install git+https://github.com/hendersonreed/galaxykit.git@aah-487-namespace-and-cli
+
     - name: "Set env.SHORT_BRANCH, env.GALAXY_NG_COMMIT"
       run: |
         SHORT_BRANCH=`sed 's/^refs\/heads\///' <<< $BRANCH`
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -22,7 +22,7 @@ jobs:
 
     - name: "Install galaxykit dependency"
       run: |
-        pip install galaxykit.git==0.1.0
+        pip install galaxykit==0.1.0
 
     - name: "Set env.SHORT_BRANCH, env.GALAXY_NG_COMMIT"
       run: |
--- a/test/cypress/support/commands.js
+++ b/test/cypress/support/commands.js
@@ -24,6 +24,7 @@
 // -- This will overwrite an existing command --
 // Cypress.Commands.overwrite('visit', (originalFn, url, options) => { ... })
 
+import shell from 'shell-escape-tag'
 var urljoin = require('url-join');
 
 Cypress.Commands.add('findnear', {prevSubject: true}, (subject, selector) => {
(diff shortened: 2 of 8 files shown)


test/cypress/support/commands.js
```diff
@@ -234,3 +234,19 @@ Cypress.Commands.add('deleteGroup', {}, (name) => {
...
 });
+
+// GalaxyKit Integration
+
+Cypress.Commands.add('galaxykit', {}, (args) => {
+    var adminUsername = Cypress.env('username');
+    var adminPassword = Cypress.env('password');
+    var server = Cypress.config().baseUrl + Cypress.env('prefix');
+
+    cy.exec(`galaxykit -s ${server} -u ${adminUsername} -p ${adminPassword} ${args}`, (error, stdout) => {
```
himdel, line 245: this probably needs quoting around everything but `args`? (technically it's only trusted data, but passwords can still be tricky)

test/cypress/support/commands.js
```diff
@@ -234,3 +234,19 @@ Cypress.Commands.add('deleteGroup', {}, (name) => {
...
+        if (error) {
+            throw new Error(`Galaxykit failed: ${error}`)
+        } else {
+            console.log(stdout);
```
himdel, line 249: (would it make sense to log the output together with corresponding `args`?)

himdel:
I like the approach, the commands seem straightfoward enough  (I'm *not* saying we wouldn't need docs though :))

Re helpers, I don't think we'd need for every single command, but it might make sense to provide some kind of ...templating helper? Something to turn `magic('user create :username :password', { username: 'testUser2', password: 'p@ssword1' })` into `cy.galaxykit("user create 'testUser2' 'p@ssword1'")` that deals with the whole escaping of single quotes and backslashes inside those strings. Otherwise we'd have to constantly think about whether our example data is valid while being passed through bash. 

[shell-escape-tag](https://www.npmjs.com/package/shell-escape-tag) might do the trick, letting us write

```
cy.galaxykit(shell`user create ${username} ${password}`)
```

and taking care of the rest.

test/cypress/support/commands.js
ironfroggy: Yeah good idea

test/cypress/support/commands.js
ironfroggy: Good catch, thanks

hendersonreed (approved):
This looks great! And good to hear that it's running properly, at least at the moment.

I do agree generally with @himdel - finding some way to ensure that the strings passed to `cy.galaxykit` are actually _valid_ galaxykit commands would be huge for the test developing experience.

I don't know that figuring that out though is worth delaying the merging of this? I don't know - I'm biased because I want galaxykit to start delivering real value soon :smile:

.github/workflows/cypress.yml
```diff
@@ -20,6 +20,10 @@ jobs:
 
     steps:
 
+    - name: "Install galaxykit dependency"
+      run: |
+        pip install galaxykit.git==0.1.0
```
himdel, line 25: `ERROR: No matching distribution found for galaxykit.git==0.1.0`

happening both on github actions and locally

himdel (approved):
LGTM, we can iron these things out later, I agree this is good to go :+1: 

Except for https://github.com/ansible/ansible-hub-ui/pull/464/files#r639097900 which would break cypress on github actions, do fix that one please :)

ZitaNemeckova (approved):

himdel:
Manually backporting to 4.3 in https://github.com/ansible/ansible-hub-ui/pull/487

PR #455, opened 2021-05-19
Change source to children in ReactMarkdown to be according to 6.0.0

ZitaNemeckova:
Go to Container Registries -> select one -> see README section 

Before:
image
image

After:
image
image

Changelog: https://github.com/remarkjs/react-markdown/blob/main/changelog.md#change-source-to-children

NOTE: NOT TO BE BACKPORTED TO 4.3 AS IT STILL HAS OLD VERSION OF `react-markdown` PACKAGE.

PR #471, opened 2021-05-21
[stable-4.2] Change source to children in ReactMarkdown to be according to 6.0.0

ZitaNemeckova:
`stable-4.2` of #455


himdel (approved):
Works in 4.2

PR #482, opened 2021-05-24
Match text style

ShaiahWren:
Hi Zita,
Here is a pr for issue:
https://issues.redhat.com/browse/AAH-553

I added className? to the <NumericLabel/> component and then split the number and label properties into 2 spans in the JSX. In collection-list-item.tsx I added a className to the label span. Then in the list-item.scss added the following properties:
.type-label {
  font-size: var(--pf-global--FontSize--xs);
  color: var(--pf-global--color--200);
  text-transform: capitalize;
}
that should now match what is in the collections list in card view. 
Lmk if I should try another approach. :)
Screen Shot 2021-05-24 at 5 14 07 PM
Screen Shot 2021-05-24 at 1 15 44 PM

### DIFF
--- a/src/components/collection-list/list-item.scss
+++ b/src/components/collection-list/list-item.scss
@@ -10,7 +10,7 @@
   text-transform: capitalize;
 }
 
-.type-label{
+.type-label {
   font-size: var(--pf-global--FontSize--xs);
   color: var(--pf-global--color--200);
   text-transform: capitalize;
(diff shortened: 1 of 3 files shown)


src/components/collection-list/list-item.scss
```diff
@@ -9,3 +9,9 @@
 .content {
   text-transform: capitalize;
 }
+
+.type-label {
```
himdel, line 13: this is a global css class, so we should go for a more descriptive name I think

what about numeric-label-capitalize-text? (or something like that)

himdel (approved):
LGTM, I'm seeing size 16px numbers and size 12px text in both card view and list view now :+1:

ShaiahWren:
pr closed.

PR #484, opened 2021-05-25
add tooltip component

ShaiahWren:
Hi, Zita && Martin :)
Here is a pr for issue:
https://issues.redhat.com/browse/AAH-554

Just adding a tooltip for the descriptions. I did not differentiate between those visibly overflowing and those not. 
Screen Shot 2021-05-25 at 2 29 04 PM
Screen Shot 2021-05-25 at 2 06 49 PM

### DIFF
--- a/src/components/cards/collection-card.tsx
+++ b/src/components/cards/collection-card.tsx
@@ -9,6 +9,7 @@ import {
   Text,
   TextVariants,
   Badge,
+  Tooltip,
 } from '@patternfly/react-core';
 
 import { Link } from 'react-router-dom';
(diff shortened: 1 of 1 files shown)


src/components/cards/collection-card.tsx
```diff
@@ -72,9 +73,17 @@ export class CollectionCard extends React.Component<IProps> {
           </div>
         </CardHeader>
         <CardBody>
-          <div className='description'>
-            {this.getDescription(latest_version.metadata.description)}
-          </div>
+          <Tooltip
+            content={
+              <div>
+                {this.getDescription(latest_version.metadata.description)}
```
himdel, line 79: `getDescription` is the function that's truncating the contents ... so right now you get the same shortened description in both the tooltip and the card ... this one should just be `latest_version.metadata.description` without the method call.

(Also no need for the `<div>` once #472 gets merged, but it's not merged yet.)

himdel (approved):
LGTM, I can see the full tooltip now :+1: :)

ShaiahWren:
pr closed.

PR #485, opened 2021-05-25
Collections/Detail: move master version dropdown under header; add last updated

ShaiahWren:
HI:)
Issue https://issues.redhat.com/browse/AAH-549

I moved the install version dropdown onto its own line surrounded by two spans with  text. And then formatted accordingly.

Let me know if you'd like me to change classNames, etc., format or reorganize it in some way. 

Have a great evening! :)

Screen Shot 2021-05-25 at 4 27 19 PM
Screen Shot 2021-05-25 at 4 30 54 PM

### DIFF
--- a/src/components/headers/base-header.tsx
+++ b/src/components/headers/base-header.tsx
@@ -31,7 +31,6 @@ export class BaseHeader extends React.Component<IProps, {}> {
       contextSelector,
       versionControl,
     } = this.props;
-    console.log(latestVersion);
     return (
       <div className={cx('background', className)}>
         {contextSelector && (
@@ -64,7 +63,9 @@ export class BaseHeader extends React.Component<IProps, {}> {
           <div className='install-version-column'>
             <span>Version</span>
             <div className='install-version-dropdown'>{versionControl}</div>
-            <span className='last-updated'>Last updated {latestVersion}</span>
+            {latestVersion ? (
+              <span className='last-updated'>Last updated {latestVersion}</span>
+            ) : null}
           </div>
         ) : null}
 
--- a/src/components/headers/base-header.tsx
+++ b/src/components/headers/base-header.tsx
@@ -44,7 +44,7 @@ export class BaseHeader extends React.Component<IProps, {}> {
             <div className='install-version-column'>
               <span>Version</span>
               <div className='install-version-dropdown'>{pageControls}</div>
-              <span className="last-updated">Last updated [exact time]</span>
+              <span className='last-updated'>Last updated [exact time]</span>
             </div>
           ) : null}
         </div>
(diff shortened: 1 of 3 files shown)


himdel:
Hey, I think we have 2 issues here :)...

This is currently moving the `pageControls` bit in `BaseHeader` to the left side, and adding things there .. but that's the wrong place. But `pageContols` (and `BaseHeader`) are used from group detail, execution evnironments, namespace detail, user form, etc. and this change doesn't make sense on all those screens. You will need to do the changes in the component that's *using* `BaseHeader`, not in baseheader itself.

The other part is...

> Move master version dropdown under header and remove install version dropdown; add “last updated exact time” to the right in gray (if data is present).

We shouldn't be showing `[exact time]`, we should be showing the actual time there. And not showing Last updated if there is no such time in the data.

Also, there's still the "install version" dropdown that should be removed, right?

src/components/headers/base-header.tsx
```diff
@@ -51,11 +55,19 @@ export class BaseHeader extends React.Component<IProps, {}> {
...
             </div>
           </div>
-
           {pageControls ? (
             <div className='header-right'>{pageControls}</div>
           ) : null}
         </div>
+        {versionControl ? (
+          <div className='install-version-column'>
+            <span>Version</span>
```
himdel, line 64: This is still introducing bits that are specific to Collections in the base header.

Can the BaseHeader change be just...

```
{versionControl ? <div className='install-version-column'>{versionControl}</div> : null}
```

and have the rest of logic (including latestVersion) passed inside that param?

src/components/headers/base-header.tsx
```diff
@@ -51,11 +55,19 @@ export class BaseHeader extends React.Component<IProps, {}> {
...
+            <div className='install-version-dropdown'>{versionControl}</div>
+            {latestVersion ? (
+              <span className='last-updated'>Last updated {latestVersion}</span>
```
himdel, line 67: `Last updated <DateComponent date={latestVersion} />`

we use DateComponent to show dates, for consistent date formats

src/components/headers/collection-header.tsx
```diff
@@ -79,23 +82,29 @@ export class CollectionHeader extends React.Component<IProps> {
...
+                  <FormSelectOption
+                    key={v.version}
+                    value={v.version}
+                    label={'v' + v.version}
+                  />
+                ))}
+              </FormSelect>
+            </div>
+            {latestVersion ? (
+              <span className='last-updated'>Last updated {latestVersion}</span>
```
himdel, line 106: `<span className='last-updated'>Last updated <DateComponent date={latestVersion} /></span>`

we use DateComponent to show dates, for consistent date formats

himdel (approved):
LGTM :)

himdel:
@ShaiahWren the tests should be fixed on master, all you need to do is remove the last commit and rebase :)

himdel (approved):

ShaiahWren:
Pr closed.

PR #498, opened 2021-06-01
[master-stable] Set maintenance window.

himdel:
cherry picked from commit 8571a2a74a217569c1308681bd7fbf3dc6ac2e2a, #122

And updated the times of the maintenance window to match the current one.

### DIFF
--- a/src/loaders/insights/Routes.js
+++ b/src/loaders/insights/Routes.js
@@ -22,7 +22,7 @@ import { Paths } from 'src/paths';
 const MaintenancePage = asyncComponent(() =>
   import(
     /* webpackChunkName: "namespace_detail" */
-    './containers/maintenance/maintenance-page'
+    '../../containers/maintenance/maintenance-page'
   ),
 );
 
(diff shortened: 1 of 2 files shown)


src/loaders/insights/Routes.js
```diff
@@ -19,6 +19,13 @@ import { Paths } from 'src/paths';
  *         see the difference with DashboardMap and InventoryDeployments.
  *
  */
+const MaintenancePage = asyncComponent(() =>
+  import(
+    /* webpackChunkName: "namespace_detail" */
+    './containers/maintenance/maintenance-page'
```
ZitaNemeckova, line 25: This should be `../../containers/maintenance/maintenance-page`

ZitaNemeckova: Fixing only because you are unavailable and it's pressing.

ZitaNemeckova (approved):

ZitaNemeckova:
Testing everything works :)

PR #507, opened 2021-06-02
Add filter to documentation

ShaiahWren:
HI Martin && Zita!!

See issue:
https://issues.redhat.com/browse/AAH-242
A new searchbar containing the same filter as Contents tab has been implemented for the documentation.
"X" button clears the input field.
Count in parentheses reflects the number of filtered results. 

The scroll feature was removed, at the direction of @himdel, until a new stateless component can be created for the searchbar itself.

Before:
Screen Shot 2021-06-02 at 3 34 11 PM
After:
Screen Shot 2021-06-30 at 4 07 21 PM

### DIFF
--- a/src/components/collection-detail/table-of-contents.tsx
+++ b/src/components/collection-detail/table-of-contents.tsx
@@ -209,7 +209,7 @@ export class TableOfContents extends React.Component<IProps, IState> {
     return (
       <NavExpandable
         key={title}
-        title={capitalize(`${title} (${links.length})`)}
+        title={capitalize(`${title} (${filteredLinks.length})`)}
         isExpanded={isExpanded}
         isActive={this.getSelectedCategory() === title}
       >
--- a/src/components/collection-detail/table-of-contents.tsx
+++ b/src/components/collection-detail/table-of-contents.tsx
@@ -189,8 +189,8 @@ export class TableOfContents extends React.Component<IProps, IState> {
 
   private renderLinks(links: DocsEntry[], title, filterString: string) {
     const isExpanded = !this.state.collapsedCategories.includes(title);
-    const filteredLinks = links.filter(
-      link => link.display.includes(filterString)
+    const filteredLinks = links.filter(link =>
+      link.display.includes(filterString),
     );
     return (
       <NavExpandable
--- a/src/components/collection-detail/table-of-contents.tsx
+++ b/src/components/collection-detail/table-of-contents.tsx
@@ -190,7 +190,7 @@ export class TableOfContents extends React.Component<IProps, IState> {
   private renderLinks(links: DocsEntry[], title, filterString: string) {
     const isExpanded = !this.state.collapsedCategories.includes(title);
     const filteredLinks = links.filter(
-      link => link.display.indexOf(filterString) > -1,
+      link => link.display.includes(filterString)
     );
     return (
       <NavExpandable
--- a/src/components/collection-detail/table-of-contents.tsx
+++ b/src/components/collection-detail/table-of-contents.tsx
@@ -47,6 +47,7 @@ interface IProps {
   selectedType?: string;
   className?: string;
   updateParams: (p) => void;
+  searchBarRef?: React.Ref<HTMLInputElement>;
 }
 
 export class TableOfContents extends React.Component<IProps, IState> {
@@ -86,6 +87,7 @@ export class TableOfContents extends React.Component<IProps, IState> {
           <ToolbarGroup>
             <ToolbarItem>
               <TextInput
+              ref={this.props.searchBarRef}
                 value={params.keywords}
                 onChange={val => {
                   updateParams(ParamHelper.setParam(params, 'keywords', val));
--- a/src/components/collection-detail/table-of-contents.tsx
+++ b/src/components/collection-detail/table-of-contents.tsx
@@ -88,7 +88,7 @@ export class TableOfContents extends React.Component<IProps, IState> {
           <ToolbarGroup>
             <ToolbarItem>
               <SearchInput
-                // ref={this.props.searchBarRef}
+                ref={this.props.searchBarRef}
                 value={params.keywords}
                 onChange={val => {
                   updateParams(ParamHelper.setParam(params, 'keywords', val));
--- a/src/components/collection-detail/table-of-contents.tsx
+++ b/src/components/collection-detail/table-of-contents.tsx
@@ -9,7 +9,6 @@ import {
   NavItem,
   NavList,
   SearchInput,
-  TextInput,
   Toolbar,
   ToolbarGroup,
   ToolbarItem,
--- a/src/components/collection-detail/table-of-contents.tsx
+++ b/src/components/collection-detail/table-of-contents.tsx
@@ -8,6 +8,7 @@ import {
   NavExpandable,
   NavItem,
   NavList,
+  SearchInput,
   TextInput,
   Toolbar,
   ToolbarGroup,
@@ -86,12 +87,13 @@ export class TableOfContents extends React.Component<IProps, IState> {
         <Toolbar>
           <ToolbarGroup>
             <ToolbarItem>
-              <TextInput
+              <SearchInput
                 ref={this.props.searchBarRef}
                 value={params.keywords}
                 onChange={val => {
                   updateParams(ParamHelper.setParam(params, 'keywords', val));
                 }}
+                onClear={ () => updateParams(ParamHelper.setParam(params, 'keywords', ''))}
                 aria-label='find-content'
                 placeholder='Find content'
               />
--- a/src/components/collection-detail/table-of-contents.tsx
+++ b/src/components/collection-detail/table-of-contents.tsx
@@ -36,7 +36,6 @@ class Table {
 
 interface IState {
   collapsedCategories: string[];
-  searchBarValue: string;
 }
 
 interface IProps {
@@ -60,7 +59,6 @@ export class TableOfContents extends React.Component<IProps, IState> {
 
     this.state = {
       collapsedCategories: [],
-      searchBarValue: this.props.params.keywords || '',
     };
   }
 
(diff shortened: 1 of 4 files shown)


src/components/collection-detail/table-of-contents.tsx
```diff
@@ -165,17 +187,19 @@ export class TableOfContents extends React.Component<IProps, IState> {
     return table;
   }
 
-  private renderLinks(links, title) {
+  private renderLinks(links: DocsEntry[], title, filterString: string) {
     const isExpanded = !this.state.collapsedCategories.includes(title);
-
+    const filteredLinks = links.filter(
+      link => link.display.indexOf(filterString) > -1,
```
himdel, line 193: would `link => link.display.contains(filterString)` work?

src/components/collection-detail/table-of-contents.tsx
```diff
@@ -165,17 +187,19 @@ export class TableOfContents extends React.Component<IProps, IState> {
...
+  private renderLinks(links: DocsEntry[], title, filterString: string) {
     const isExpanded = !this.state.collapsedCategories.includes(title);
-
+    const filteredLinks = links.filter(link =>
+      link.display.includes(filterString),
+    );
     return (
       <NavExpandable
         key={title}
         title={capitalize(`${title} (${links.length})`)}
```
ZitaNemeckova, line 198: I think number of items should depend on filter's results like so
Screenshot 2021-06-03 at 12 57 56
But I'm not sure because it's not the case for Contents tab. @sbuenafe-rh WDYT?

src/components/collection-detail/table-of-contents.tsx
```diff
@@ -47,7 +57,7 @@ export class TableOfContents extends React.Component<IProps, IState> {
   constructor(props) {
     super(props);
 
-    this.state = { collapsedCategories: [] };
+    this.state = { collapsedCategories: [], searchBarValue: '' };
```
ZitaNemeckova, line 60: @ShaiahWren You should read from `this.props.params.keywords` and if any value is present set `searchBarValue` to it. Otherwise the search value is lost on reload. Also you'll need to change `params` from `object` to `{keywords?: string}` in `interface IProps`.

ZitaNemeckova:
Closing and reopening to see if Cypress goes green.

ZitaNemeckova:
When I go to Documentation for first time there's no data in left navbar.
Screenshot 2021-06-11 at 15 03 42

ShaiahWren:
There should be data now, lmk if you can see it. I also made the searchbar case insensitive.

src/components/collection-detail/table-of-contents.tsx
```diff
@@ -165,17 +197,21 @@ export class TableOfContents extends React.Component<IProps, IState> {
...
   }
 
-  private renderLinks(links, title) {
+  private renderLinks(links: DocsEntry[], title, filterString: string) {
     const isExpanded = !this.state.collapsedCategories.includes(title);
-
+    const filteredLinks = links.filter(link =>
+      link.display.toLowerCase().includes(filterString.toLowerCase()),
+    );
+    console.log('filterstring: ', filterString);
```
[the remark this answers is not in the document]
ShaiahWren: What you mean here...  adding an 'X' to the end of a search bar to clear the search on click? 
Or literally implement the chip keywords functionality?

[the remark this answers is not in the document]
newswangerd: Something like this 
Screen Shot 2021-06-16 at 2 32 21 PM

There might be a way to do this with the patternfly text input field.

ShaiahWren:
@himdel The filter looks okay... but the scrolling issue remains even after I added the 'clear' button that David suggested. You mentioned that using the window/focus  might not be the right approach. What would your strategy be to stop the scrolling? Thanks :)

himdel:
> What would your strategy be to stop the scrolling? 

If I understand correctly, the intent behind the scrolling was to scroll up to the actual content when the user clicks a link in the documentation.. so it happens in `render()` mostly by coincidence, because that would have been the only re-render that happened (outside of switching tabs, which drop the content anyway).

So I'd say we can just move the code outside of render, to the onclick .. make it happen explicitly when clicking a documentation link, instead of for all renders. (That would be a link insinde the doc body, or from the Documentation/Modules/Plugins links on the left side.) Does that help? :)

ShaiahWren:
@himdel This is very helpful... I'm not sure if I am implementing it correctly thought. See screenshot, noting lines 200-208 and 233
Screen Shot 2021-06-22 at 4 06 17 PM

himdel:
Aaah, you're on the right track I think.

Does it work?

I think you may not need the `if` at all now, but otherwise if it works, LGTM :)

And if it *doesn't* work, it's probably because the `onclick` happens before the href causes a navigation event, which triggers react router to re-render the whole thing, and either the ref changes to a new value, or the scroll gets cancelled by navigation, in which case delaying the callback might do the trick. (Or there could be a way to conditionally render on navigation, though I'm not 100% sure there.)

ShaiahWren:
@himdel Hmm! I added a setTimeout 5000 to the execute scroll function. It doesn't scroll at all now. Is it in the wrong place?

himdel:
I'll have to take a look, most likely it's really trying to scroll to an element  that no longer exists.

But maybe all we need to do is to scroll before navigating. What about changing that Link to not have a href at all, only do the onclick, and have it scroll up first, and then go to the link?

Did the original version without setTimeout work at all?

ShaiahWren:
@himdel The original version didn't work.
I tried several things, none of which worked... and now I'm feeling a bit stuck. 
Given how long I've been working on this particular issue I think it'd be good for you to take a closer look :)

himdel:
OK, here's what I've found...

All of the plugins I have seem to use the `displayHTML` branch of `render()`, which means we're displaying raw html (line 155), and not `RenderPluginDoc` which is the only user of `renderDocLink`.

So, we're stuck with the links just assuming the url, without us having a good way to do anything with that.
Which rather limits our options :).
I'm also having trouble finding any cross-docs links that actually work and don't end up 404ing or going to another domain.

Maybe we should just drop the scroll code entirely in this PR, and deal with that separately, seems like a proper problem by itself.

---

As far as actual solutions...

* can't just do it in `render` because both search and navigation change state and there's no nice way of distinguishing between the two
  * (except for hacks like also saving the last search, and comparing, except that would also be saved in state, and it'd be very easy to accidentally break it into an infinite render)
  * your original solution of checking the active element is definitely one of the better available solutions, but now there's a clear button you'd also have to check if *that* wasn't active (and we'd have to add any other state change to the logic if there is a new one) .. still, might be the least problematic version for now :)

* can't do it before clicking the link, as the links are not under our control
  * (except for parsing the generated HTML, or iterating through any `<a` elements in the content and adjusting there; again might be hard to not accidentally do this multiple times to the same elements)

* we might be able to use the browser navigation event itself, though I'm not excited about that either
  * we could scroll to top in `onbeforeunload` event handler, we'd have to write a wrapper component so that it gets unloaded properly when changing tabs, etc., but I'm not sure it would work reliably

* we might be able to use react router events (`useLocation` or `history.listen`) to detect route change and compare to previous state, as long as the component state survives navigation (I need to find a working link to verify that :)) (this seems to show the options https://stackoverflow.com/questions/41911309/how-to-listen-to-route-changes-in-react-router-v4
  * if that works well, any route navigation from a route matching `Paths.collectionDocsPageByRepo` to a route matching the same pattern but different values could do the scrolling
  * the catch is that `useLocation` is simple to use .. in functional components, but not necesarily class components, while history listeners need careful logic to register and unregister when needed

himdel:
Hm, there may be one more solution...

State changes trigger a re-render, but the only reason we'd actually have to re-render the docs segment itself is if the content changes, not the search.

Looking at it that way, it's just that this component tries to do too much.
If we can split the part that renders the docs into a separate stateless component, then that component should only be re-rendered on prop change, so we *can* scroll to top in *that* `render()`, I think... testing..

ShaiahWren:
@himdel thanks for all you work and for looking into this!!
At this point, what do you think I should do with this pr. You mentioned moving the search state to a different component, which sounds like a good idea.
I wonder if tackling that should be moved to another issue/pr - would that be okay?
If so, what could I do to make this pr pass muster for now?
I'm just thinking this might be a good idea, given that it's been in the works for an entire sprint...
I'm happy to keep working on the new issue and help build the new component :)

himdel:
Agreed, I think the easiest way forward is either to comment out the "scroll to top" logic, or keep your first version with the `activeElement` logic, I'm fine with either :).

It absolutely makes sense to tackle the scrolling in a separate PR, this is pretty much ready otherwise :).

(In general, there's almost no situation where splitting up a PR would be a bad thing, smaller PRs a easier review.
If you're doing the same change in a 1000 files, theres' no point in splitting into a 1000 PRs, but except for things like that, feel free to split whenever it makes sense to you :).)

ShaiahWren:
I removed (commented) the scrolling feature. 
Lmk if it's okay to merge as it is now :)

himdel (approved):
LGTM, works in the UI :+1: :)

src/components/collection-detail/table-of-contents.tsx
```diff
@@ -165,17 +200,19 @@ export class TableOfContents extends React.Component<IProps, IState> {
...
     const isExpanded = !this.state.collapsedCategories.includes(title);
-
+    const filteredLinks = links.filter(link =>
+      link.display.toLowerCase().includes(filterString.toLowerCase()),
+    );
     return (
       <NavExpandable
         key={title}
-        title={capitalize(`${title} (${links.length})`)}
+        title={capitalize(`${title} (${filteredLinks.length})`)}
```
himdel, line 211: this only fixes the count on the documentation tab,

for Contents tab, the logic is in `CollectionContentList`, the counting loop doesn't take filtering into account

Are you planning to address that one in a separate PR?

ShaiahWren:
@himdel  Lmk if this looks ok to merge... I fixed the contents tab to match the count. :)

himdel:
The new changes LGTM :+1:, tested, they work the same now :)

ShaiahWren:
pr closed. :) :)

PR #490, opened 2021-05-27
Various fixes

newswangerd:
- Fix repo selector on collection detail: https://issues.redhat.com/browse/AAH-663
- Remove some unused permissions from the UI
- Fix a 404 on collection details page when there is no content in a collection: https://issues.redhat.com/browse/AAH-666

PR #508, opened 2021-06-02
Fix group_permissions spec to ignore private container permissions

himdel:
#490 disables 2 container permissions and renames the group,
 this updates the tests to do the same.

This should get cypress green on master,
the same change was also backported as #501 => adding `backport-4.3` as well.


ZitaNemeckova (approved):
Tested. LGTM 👍

PR #509, opened 2021-06-03
Namespace details, Collections tab - use the same filter as Collections screen

himdel:
Fixes https://issues.redhat.com/browse/AAH-245, https://issues.redhat.com/browse/AAH-544, https://issues.redhat.com/browse/AAH-555

This unifies filtering on both collection list screens:

**Collections > Collections**

(no change)

**Collections > Namespaces > View Collections > tab Collections**

before:

after:

---

This also allows EmptyStateFilter to reset the filters, enabled in Collections and in Namespace detail:

---

Two remaining differences remain:
* no card view on namespace detail page, only list view
* clearing filters doesn't clear the search input (also present in users, but not groups, with a similar issue in containers)

### DIFF
--- a/src/components/collection-list/collection-filter.scss
+++ b/src/components/collection-list/collection-filter.scss
@@ -1,0 +1,12 @@
+.toolbar-wrapper {
+  .toolbar {
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+
+    .pagination-container {
+      display: flex;
+      align-items: center;
+    }
+  }
+}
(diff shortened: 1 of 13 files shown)


himdel:
Cc @ZitaNemeckova , @sbuenafe-rh this should be ready :)

sbuenafe-rh:
@himdel On the namespace page, the list view of all the collections, the toolbar should be connected to the list.

ZitaNemeckova (approved):
LGTM 👍

himdel:
Adding backport-4.3, as all 3 issues are slated for 4.4cloud*

PR #512, opened 2021-06-03
resize logo to 40px

ShaiahWren:
Hello! :)

See issue:
https://issues.redhat.com/browse/AAH-552
I wasn't able to get a screenshot because I am not running in insights mode, but I think this fixes the issue. Just let me know if not.

### DIFF
--- a/src/components/cards/collection-card.tsx
+++ b/src/components/cards/collection-card.tsx
@@ -51,6 +51,7 @@ export class CollectionCard extends React.Component<IProps> {
             image={namespace.avatar_url}
             alt={company + ' logo'}
             size='40px'
+            unlockWidth={true}
           />
           <TextContent>{this.getCertification(repo)}</TextContent>
         </CardHeader>
--- a/src/components/cards/namespace-card.tsx
+++ b/src/components/cards/namespace-card.tsx
@@ -33,7 +33,7 @@ export class NamespaceCard extends React.Component<IProps, {}> {
               unlockWidth
               image={avatar_url}
               alt={company + ' logo'}
-              size='50px'
+              size='40px'
             />
           </CardHeaderMain>
         </CardHeader>
--- a/src/components/collection-list/collection-list-item.tsx
+++ b/src/components/collection-list/collection-list-item.tsx
@@ -54,7 +54,8 @@ export class CollectionListItem extends React.Component<IProps, {}> {
           <Logo
             alt={company + ' logo'}
             image={namespace.avatar_url}
-            size='50px'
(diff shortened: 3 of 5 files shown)


ZitaNemeckova:
Screenshot 2021-06-04 at 14 23 26

It seems this needs more CSS magic.

You can try it out in standalone as well under Namespaces -> click View Collections. You can add an image on that page. Click the kebab next to `Upload Collections` button. Click `Edit namespace`. Add Url to any image to `Logo URL`.

ShaiahWren:
I need a bit more clarification on setting the size :) Should the image height + width be set to 50px? My only concern is that setting fixed height/width might affect the aspect ratio of the logo in wonky ways

himdel:
The problem is in header.scss .. `.image` has `padding-right: 16px` but padding gets counted as part of `width` and `height` .. so either Logo has to add 16 to the width, or it can't use `.image`

ShaiahWren:
GIve this another try... I added 16 px to the size attribute on logo in the base-header.tsx. I completely understand if this is still the wrong approach, just let me know

himdel:
It's *probably* the wrong approach... if we pass actual numerical pixel size to Logo, it should mean the exact max-dimension size of the logo.

That said, given this css difficulty, I'm not sure it makes sense for `Logo` to have a `size` prop in the first place.
We may prefer to achieve the same result by doing the sizing entirely in css.

On the other hand, we can as well drop the padding and make sure the top logo (or any other logos) stays working as it should,
or we can switch from `<img` to `background-image` with `background-size: contain` and the padding becomes irrelevant.

ZitaNemeckova:
@ShaiahWren @himdel We should bring this up with Susan tomorrow and ask her to show us how it should looklike. She said it's same as Catalog but I have no data there...

ShaiahWren:
Take a look... the logo on CollectionCard, CollectionListItem, NamespaceCard, and BaseHeader are all now 40px (height) and the minimum-width has been removed using unlockWidth={true}. The logos are now... very big, but they ARE now 40px in height, which was initially requested.

ZitaNemeckova:
@sbuenafe-rh Which one is better? Limiting `width` or not? 
Screenshot 2021-06-10 at 11 25 07
Screenshot 2021-06-10 at 11 25 57

ZitaNemeckova (approved):
LGTM 👍 Just need to double check `width` with Susan.

sbuenafe-rh:
@ZitaNemeckova yep, the list view looks good when we limit the width.

PR #513, opened 2021-06-03
wrap namespace name in card container

ShaiahWren:
Hi David!!
Issue:
https://issues.redhat.com/browse/AAH-165
I simply put flex-wrap on the .ns-card-container class. See cards.scss & namespace-card.tsx. 
Screen Shot 2021-06-03 at 2 33 27 PM
Screen Shot 2021-06-03 at 2 29 26 PM

Seems to have done the trick.

### DIFF
--- a/src/components/cards/namespace-card.tsx
+++ b/src/components/cards/namespace-card.tsx
@@ -40,8 +40,8 @@ export class NamespaceCard extends React.Component<IProps, {}> {
           </CardHeaderMain>
         </CardHeader>
         <Tooltip content={<CardTitle>{company || name}</CardTitle>}>
-        <CardTitle>{this.getDescription(company || name)}</CardTitle>
-          </Tooltip>
+          <CardTitle>{this.getDescription(company || name)}</CardTitle>
+        </Tooltip>
         <CardBody className={'wrap'}>{name}</CardBody>
         {namespaceURL && (
           <CardFooter className={'wrap'}>
--- a/src/components/cards/namespace-card.tsx
+++ b/src/components/cards/namespace-card.tsx
@@ -39,15 +39,15 @@ export class NamespaceCard extends React.Component<IProps, {}> {
             />
           </CardHeaderMain>
         </CardHeader>
-        <Tooltip content={<CardTitle>{company || name}</CardTitle>}>
+        <Tooltip content={company || name}>
           <CardTitle>{this.getDescription(company || name)}</CardTitle>
         </Tooltip>
-        <Tooltip content={<CardBody>{name}</CardBody>}>
+        <Tooltip content={name}>
           <CardBody>{this.getDescription(name)}</CardBody>
         </Tooltip>
 
         {namespaceURL && (
-          <CardFooter className={'wrap'}>
+          <CardFooter>
             <Link to={namespaceURL}>View collections</Link>
           </CardFooter>
         )}
--- a/src/components/cards/namespace-card.tsx
+++ b/src/components/cards/namespace-card.tsx
@@ -37,10 +37,10 @@ export class NamespaceCard extends React.Component<IProps, {}> {
             />
           </CardHeaderMain>
         </CardHeader>
-        <CardTitle>{company || name}</CardTitle>
-        <CardBody>{name}</CardBody>
+        <CardTitle className={"wrap"}>{company || name}</CardTitle>
+        <CardBody className={"wrap"}>{name}</CardBody>
         {namespaceURL && (
-          <CardFooter>
+          <CardFooter className={"wrap"}>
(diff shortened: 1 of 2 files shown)


ZitaNemeckova:
@ShaiahWren Not sure why it worked for you but it didn't for me in Chrome. 

This worked for me:
```diff
diff --git a/src/components/cards/cards.scss b/src/components/cards/cards.scss
index 65b03e3..fb356be 100644
--- a/src/components/cards/cards.scss
+++ b/src/components/cards/cards.scss
@@ -1,6 +1,6 @@
 .ns-card-container {
   width: 300px;
-  flex-wrap: wrap;
+  height: 280px;
 
   .title {
     font-weight: bold;
@@ -13,7 +13,10 @@
     width: 180px;
   }
 }
-
+.wrap {
+  width: 300px;
+  word-break: break-word;
+}
 .collection-card-container {
   width: 280px;
   // height: 340px;
diff --git a/src/components/cards/namespace-card.tsx b/src/components/cards/namespace-card.tsx
index b5914cc..66962cf 100644
--- a/src/components/cards/namespace-card.tsx
+++ b/src/components/cards/namespace-card.tsx
@@ -37,8 +37,8 @@ export class NamespaceCard extends React.Component<IProps, {}> {
             />
           </CardHeaderMain>
         </CardHeader>
-        <CardTitle>{company || name}</CardTitle>
-        <CardBody>{name}</CardBody>
+        <CardTitle className={"wrap"}>{company || name}</CardTitle>
+        <CardBody className={"wrap"}>{name}</CardBody>
         {namespaceURL && (
           <CardFooter>
             <Link to={namespaceURL}>View collections</Link>
```

ShaiahWren:
Hi Zita!!
I implemented your changes and here is what I have... it addresses the overflow for the title but not the CardBody of CardFooter. Is this a concern? 

David mentioned using a Tooltip + truncating method for the title. What do you think of that? 

See photo:
Screen Shot 2021-06-04 at 10 24 03 AM

ShaiahWren:
And here is an edge case where a long long title completely overflows the card. Not a likely scenario, but the might have been what David was thinking when he thought of the Tooltip...
Screen Shot 2021-06-04 at 10 33 08 AM

ZitaNemeckova:
@ShaiahWren You can create a namespace with name that contains spaces, commas and dots? It didn't let me make the name longer than 64 chars.
Screenshot 2021-06-04 at 16 39 05
Screenshot 2021-06-04 at 16 39 30

ZitaNemeckova:
hmm `CardFooter` should always be `View Collections` so it shouldn't be a problem. As for `CardBody` it should work with `clasName={'wrap'}`.

But truncating and having a Tooltip sounds like good idea too. Do you like that idea better?

ShaiahWren:
I think Tooltip sounds like a good option.
Shall I do that for both the title and CardBody, or just one?

ZitaNemeckova:
@ShaiahWren both :)

ShaiahWren:
I'm using the same truncating method on both the CardTitle and CardBody, but of course the "length" of the two will appear differently on bolded vs plain size text, see screenshot. Is this and issue? If so I'll make two separate truncating methods. Thanks! 
Screen Shot 2021-06-07 at 10 59 51 AM

newswangerd (approved):

himdel:
So.. I don't want to block this PR on this, since we're already using this elsewhere.

**But** truncating the text to N characters will never really work... both of these are exactly 32 characters:




Ideally, this should be fixed in css, using `text-overflow: ellipsis`

src/components/cards/namespace-card.tsx
```diff
@@ -37,14 +39,30 @@ export class NamespaceCard extends React.Component<IProps, {}> {
             />
           </CardHeaderMain>
         </CardHeader>
-        <CardTitle>{company || name}</CardTitle>
-        <CardBody>{name}</CardBody>
+        <Tooltip content={<CardTitle>{company || name}</CardTitle>}>
```
ZitaNemeckova, line 42: No need to add `<CardTitle>` or `<CardBody>` to the `content`.

ShaiahWren:
@ZitaNemeckova , I cleaned up the Tooltip so CardBody and CardTitle aren't in the 'content' attribute anymore. Also removed the className='wrap' from the footer. Not sure why I put that there in the first place, must have been from an earlier draft. 
@himdel you have a great point. In the case that we put text-overflow: ellipsis on CardTitle & CardBody, the full name will simply not be visible on the card view until clicking on 'view collections'. 
I'm happy to make the change. Lmk

ZitaNemeckova (approved):

ShaiahWren:
Pr closed

PR #514, opened 2021-06-03
minor edit to approval status message

ShaiahWren:
Hi Henderson! :)

Issue: https://issues.redhat.com/browse/AAH-169

A minor change to import-console.tsx. Approval status now reads "could not be determined yet".

### DIFF
--- a/src/components/my-imports/import-console.tsx
+++ b/src/components/my-imports/import-console.tsx
@@ -143,7 +143,7 @@ export class ImportConsole extends React.Component<IProps, {}> {
       } else if (rlist.includes(Constants.PUBLISHED)) {
         approvalStatus = 'approved';
       } else {
-        approvalStatus = 'could not be determined';
+        approvalStatus = 'could not be determined yet';
       }
 
       collectionHead = (


hendersonreed (changes requested):
@ShaiahWren this is great - looks like @awcrosby  suggested wording along the lines of 'approval status could not be determined yet', so I see why you went with that.

Is there a good reason for using that wording instead of something that's a little more accurate, like "awaiting review" or something like that? 'not determined' implies that galaxy_ng the product isn't sure about the approval status, when really galaxy_ng _does_ know the status of the collection - it's in limbo and needs to be approved or rejected.

awcrosby:
This message will only show until the `CollectionVersion` is moved from the `inbound-<ns>` repo to either `staging` or `published` - so it is a temporary statement until the task to move the `CollectionVersion` completes.

"awaiting review" would be wrong here since it may not require review, it may go directly to `published`. We suggested just adding "yet" after talking with david and calvin today.

awcrosby (approved):

ZitaNemeckova:
@hendersonreed are you ok with Andrew's explanation?

hendersonreed (approved):
Just went in and double-checked this, was unable to reproduce the original bug - it now says "waiting for approval" when a collection has been neither approved nor rejected. 

Either it was my error when I reported the bug, or it has been fixed during some other PR. I'd say 'could not be determined yet' is ideal. Thanks!

ShaiahWren:
pr closed.

PR #522, opened 2021-06-09
Maintenance window test in prod-beta

ZitaNemeckova:
This is test PR to check that Maintenance window doesn't need Automation Hub API to show up.


src/loaders/insights/Routes.js
```diff
@@ -149,6 +156,8 @@ InsightsRoute.propTypes = {
 export const Routes = props => {
   const path = props.childProps.location.pathname;
 
+  return <MaintenancePage></MaintenancePage>;
```
himdel, line 159: this is no longer needed (both changes in this file)

ZitaNemeckova: Removed :)

PR #523, opened 2021-06-10
EmptyStateFilter: replace empty message

himdel:
Updating the No results match the filter criteria message.

Before:

After:


himdel:
Cc @sbuenafe-rh :)

sbuenafe-rh:
@himdel LGTM!

PR #525, opened 2021-06-10
Namespace form cypress

ShaiahWren:
Hi!:)
This pr addresses issue:
https://issues.redhat.com/browse/AAH-616

I believe I will need some extra input on this. 
The first test, which creates a new namespace, only passes correctly the first time it is run-- the next time there is a duplicate error message and the 'create' button is disabled.

There is a similar, inverse problem when testing the duplicate error message itself. 

Any input would be greatly appreciated!

### DIFF
--- a/src/loaders/standalone/standalone-loader.tsx
+++ b/src/loaders/standalone/standalone-loader.tsx
@@ -321,7 +321,7 @@ class App extends React.Component<RouteComponentProps, IState> {
           url: formatPath(Paths.searchByRepo, {
             repo: this.state.selectedRepo,
           }),
-          'data-cy':'collections',
+          'data-cy': 'collections',
         }),
         menuItem('Namespaces', {
           url: Paths[NAMESPACE_TERM],
--- a/src/loaders/standalone/standalone-loader.tsx
+++ b/src/loaders/standalone/standalone-loader.tsx
@@ -321,6 +321,7 @@ class App extends React.Component<RouteComponentProps, IState> {
           url: formatPath(Paths.searchByRepo, {
             repo: this.state.selectedRepo,
           }),
+          'data-cy':'collections',
         }),
         menuItem('Namespaces', {
           url: Paths[NAMESPACE_TERM],
--- a/src/loaders/standalone/standalone-loader.tsx
+++ b/src/loaders/standalone/standalone-loader.tsx
@@ -321,7 +321,6 @@ class App extends React.Component<RouteComponentProps, IState> {
           url: formatPath(Paths.searchByRepo, {
             repo: this.state.selectedRepo,
           }),
-          'data-cy': 'collections',
         }),
         menuItem('Namespaces', {
           url: Paths[NAMESPACE_TERM],
--- a/test/cypress/integration/menu.js
+++ b/test/cypress/integration/menu.js
@@ -87,6 +87,8 @@ describe('Hub Menu Tests', () => {
 
         it('creates new namespace', () => {
           cy.login(username, password);
+          const createNamespaceButton = cy.get('[data-cy=create-namespace]')
+          createNamespaceButton.click()
         })
     });
 
--- a/test/cypress/integration/menu.js
+++ b/test/cypress/integration/menu.js
@@ -67,7 +67,3 @@ describe('Hub Menu Tests', () => {
     });
   });
 });
-
-
-
-
--- a/test/cypress/integration/menu.js
+++ b/test/cypress/integration/menu.js
@@ -53,14 +53,19 @@ describe('Hub Menu Tests', () => {
 
     it('creates new namespace', () => {
       cy.login(adminUsername, adminPassword);
-    //   cy.on('uncaught:exception', (err, runnable) => {
-    //     return false
-    // })
-    const collectionTab = cy.get('[data-cy=collections]')
-    collectionTab.click()
+      cy.on('uncaught:exception', (err, runnable) => {
+        return false
+    })
+    const collectionTab = cy.menuPresent('Namespaces')
+
+    collectionTab.click({ force: true})
       const createNamespaceButton = cy.get('[data-cy=create-namespace]')
       createNamespaceButton.click()
+      const create = cy.get('[data-cy=create]')
+      create.click()
     })
+
+    
 
     describe('user without permissions', () => {
         let username = 'nopermission';
--- a/test/cypress/integration/menu.js
+++ b/test/cypress/integration/menu.js
@@ -51,56 +51,10 @@ describe('Hub Menu Tests', () => {
       cy.logout();
     });
 
-    it('creates new namespace', () => {
-      cy.login(adminUsername, adminPassword);
-      cy.on('uncaught:exception', (err, runnable) => {
-        return false
-    })
-    const collectionTab = cy.menuPresent('Namespaces')
-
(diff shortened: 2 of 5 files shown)


test/cypress/integration/menu.js
```diff
@@ -1,60 +1,139 @@
...
+    before(() => {
+    cy.login(adminUsername, adminPassword);
+    cy.on('uncaught:exception', (err, runnable) => {
+      return false;
+    });
+    const collectionTab = cy.menuPresent('Namespaces');
+    collectionTab.click({ force: true });
+    const createNamespaceButton = cy.get('[data-cy=create-namespace]');
+    createNamespaceButton.click();
+    const create = cy.get('[data-cy=create]');
```
ZitaNemeckova, line 37: Maybe `createButton` to make it easier to read.

ZitaNemeckova:
@ShaiahWren `data-cy` use is great!

- [ ] Test should be in `namespaces.js` or in a new file
- [ ] `should.(...` test should be inside `it`
- [ ] created namespace should be deleted in `after` if possible (@hendersonreed Is it possible to delete namespace via `galaxykit`?)

ZitaNemeckova:
@ShaiahWren What do you think about something like this for testing all possible error messages? It's your code with a bit of "keep it simple" approach (mainly just `get...` functions)
```js
describe('A namespace form', () => {
  let adminUsername = Cypress.env('username');
  let adminPassword = Cypress.env('password');
  let getMessage = () => {return cy.get('.pf-c-form__helper-text');};
  let getCreateButton = () => {return cy.get('[data-cy=create]');};
  let getInputBox = () => {return cy.get('#pf-modal-part-2 #newNamespaceName');};
  before(() => {
    cy.logout();
    cy.login(adminUsername, adminPassword);

    cy.menuGo('Collections > Namespaces');
    const createNamespaceButton = cy.get('[data-cy=create-namespace]');
    createNamespaceButton.click();
  });
  it('should give message before typing', () => {
    getMessage().should('have.text', 'Please, provide the namespace name');
    getCreateButton().should('be.disabled');
  });
  it('should give message if input is empty', () => {
    getInputBox().type(' ');
    cy.get('[data-cy=create]').should('be.disabled');
    getMessage().should('have.text', 'Name can only contain [A-Za-z0-9_]');
    getCreateButton().should('be.disabled');
  });
// here goes other error messages checks
});
```

ZitaNemeckova:
For testing duplicate error I'd use `cy.galaxykit("-i namespace create", "testns1")` command to create a namespace and then wrote the test that tries to create `testns1` via a form.

As for the problem for testing a creating new namespace in multiple runs assuming it's not possible to delete a namespace via `galaxykit`. You either have to change the name of that namespace or reset DB before you run the test again.

Reseting DB via CLI (from `galaxy_ng` dir):
```
./compose down --volume
./compose up -d postgres redis
./compose run --rm api manage migrate
./compose run --rm -e PULP_FIXTURE_DIRS='["/src/galaxy_ng/dev/automation-hub"]' api manage loaddata initial_data.json
```

hendersonreed:
@ZitaNemeckova I don't know that namespace deletion via the API is possible yet even. I've been poking at it, and haven't been able to (getting HTTP 403 permission errors.) I've asked @awcrosby about it, but my suspicion is that we'll get namespace deletion support when we get support for content deletion.

hendersonreed:
Just got confirmation that it is indeed disabled for all users at the moment.

test/cypress/integration/menu.js
ShaiahWren: Agreed!

test/cypress/integration/namespace_form.js
```diff
@@ -0,0 +1,93 @@
...
+    getCreateButton().should('be.disabled');
+    clearInput();
+  });
+  it('should give message if input is shorter than 3 characters', () => {
+    getInputBox().type('na');
+    getMessage().should('have.text', 'Name must be longer than 2 characters');
+    getCreateButton().should('be.disabled');
+    clearInput();
+  });
+  if (
```
ZitaNemeckova, line 61: Typo? `if` -> `it`?

test/cypress/integration/namespace_form.js
```diff
@@ -0,0 +1,90 @@
...
+  });
+
+  // Note: the next test should work the first time, but not on rerun. to reset DB via CLI (from galaxy_ng dir):
+
+  // ./compose down --volume
+  // ./compose up -d postgres redis
+  // ./compose run --rm api manage migrate
+  // ./compose run --rm -e PULP_FIXTURE_DIRS='["/src/galaxy_ng/dev/automation-hub"]' api manage loaddata initial_data.json
+
+  // it('creates a new namespace with no error messages', () => {
```
ZitaNemeckova, line 85: Please uncomment :)

ShaiahWren: The last test on line 85?
And the reset DB steps as well?

ZitaNemeckova: Just the test. Reset DB steps has to stay as a comment for a developer or you can remove them. Up to you :)

ZitaNemeckova:
@ShaiahWren So it's definely GH Actions building prod UI without `data-cy`. So you'll need to use `className` instead :(

As for `cy.wait(5000)` you can either remove it or replace it witth waiting on a request like [here]( https://github.com/ansible/ansible-hub-ui/blob/master/test/cypress/support/commands.js#L144-L151)

ShaiahWren:
I removed all the data-cy attributes... it passes locally but not here as you can see. So sad

Also, I tried changing beforeEach() hook to before() and the last two tests failed (locally). Do you have any ideas why?

src/loaders/standalone/standalone-loader.tsx
```diff
@@ -321,6 +321,7 @@ class App extends React.Component<RouteComponentProps, IState> {
           url: formatPath(Paths.searchByRepo, {
             repo: this.state.selectedRepo,
           }),
+          'data-cy': 'collections',
```
ZitaNemeckova, line 324: Remaining `data-cy` :)

ZitaNemeckova:
@ShaiahWren Keep `beforeEach`. It seems that if it's not there than request returns `403` and it will need more investigation into why. And I don't want to block this PR any longer.

test/cypress/integration/namespace_form.js
```diff
@@ -0,0 +1,90 @@
...
+  let adminPassword = Cypress.env('password');
+
+  let getCreateNamespace = () => {
+    return cy.get('.pf-c-button.pf-m-primary');
+  };
+  let getMessage = () => {
+    return cy.get('.pf-c-form__helper-text');
+  };
+  let getCreateButton = () => {
+    return cy.get('.pf-c-button.pf-m-primary.submit-button');
```
ZitaNemeckova, line 13: Use this `'.pf-c-modal-box__footer .pf-m-primary'` instead. I tested it and it worked in GH :) then remove `className='submit-button'`

ShaiahWren:
@hendersonreed would you be able to take a look at why my pr continues to fail? Zita gets it to work on her end.

hendersonreed:
@ShaiahWren looking now :)

ShaiahWren:
Finally merging :)

PR #527, opened 2021-06-11
Wrap pre element so it doesn't overflow

ZitaNemeckova:
Fixes https://issues.redhat.com/browse/AAH-599

Collection detail documentation page of: `ibm` -> `power_ibm` -> `module > ibmi_install_product_from_savf`

Before:
Screenshot 2021-06-11 at 10 30 49

After:
Screenshot 2021-06-11 at 10 42 45

### DIFF
--- a/src/containers/collection-detail/collection-detail.scss
+++ b/src/containers/collection-detail/collection-detail.scss
@@ -47,6 +47,6 @@
     }
   }
 }
-pre {
+.docs-container pre {
   white-space: pre-wrap;
 }
--- a/src/containers/collection-detail/collection-detail.scss
+++ b/src/containers/collection-detail/collection-detail.scss
@@ -47,3 +47,6 @@
     }
   }
 }
+pre {
+  white-space: pre-wrap;
+}


ShaiahWren (commented):
LGTM 
white-space selectors are new for me! It looks like it works well for the task at hand.

ShaiahWren (approved):
Nicely done.

src/containers/collection-detail/collection-detail.scss
```diff
@@ -47,3 +47,6 @@
     }
   }
 }
+pre {
```
himdel, line 50: global style for every single `pre`? That doesn't sound safe, use a class?

ZitaNemeckova: Cannot use the class. It's generated in `@ansible/galaxy-doc-builder`

himdel: You can still use the known parent class, 
`.docs-container pre` is still better than global `pre` :)

ZitaNemeckova:
@himdel Updated. Please have a look. Thanks :)

himdel (approved):
Works for me :)

PR #539, opened 2021-06-15
[prod-stable] Maintenance window

ZitaNemeckova:


.github/workflows/deploy-cloud.yml
```diff
@@ -0,0 +1,58 @@
+name: "Deploy cloud.redhat.com"
```
himdel, line 1: `name: "Deploy cloud.redhat.com from prod-stable"` please? (so that we can tell where an action comes from)

himdel (approved):

PR #540, opened 2021-06-15
Edit namespace fields Cypress testing

ShaiahWren:
Hi Zita!! :)
https://issues.redhat.com/browse/AAH-61

This pr checks the following fields:
- name: not editable in the form
- company name: must not exceed 64 characters

I will do more prs for the other fields, but I wanted to first make sure the organization of the test was to your liking. :) Please advise with any feedback. 

In general, it is my preference to use data-cy selectors, as there are often  id's or classnames inherited from patternfly that apply to multiple elements on the browser. But I didn't use them this time because I know we don't use Travis anymore. The other pr 525 had this issue of duplicate selectors, so I'll have to wait for Martin's bugfix...


src/components/cards/namespace-card.tsx
```diff
@@ -28,7 +28,7 @@ export class NamespaceCard extends React.Component<IProps, {}> {
   render() {
     const { avatar_url, name, company, namespaceURL } = this.props;
     return (
-      <Card className='ns-card-container'>
+      <Card data-cy='ns-card' className='ns-card-container'>
```
ZitaNemeckova, line 31: Unfortunately GH Actions is ignoring `data-cy`. Please use `className` instead.

test/cypress/integration/namespace-edit.js
```diff
@@ -0,0 +1,84 @@
+describe('Edit a namespace', () => {
+  let baseUrl = Cypress.config().baseUrl;
+  let adminUsername = Cypress.env('username');
+  let adminPassword = Cypress.env('password');
+
+  let createNamespace = () => {
+    return cy.galaxykit('-i namespace create', 'testns1');
+  };
+  let viewNamespaceDetail = () => {
+    let link = cy.get('a[href*="ui/repo/published/testns1"]').click();
```
ZitaNemeckova, line 10: Nitpicking. Replacing`let link =` with `return` would make the code cleaner :)

test/cypress/integration/namespace-edit.js
```diff
@@ -0,0 +1,84 @@
...
+    return link;
+  };
+  let kebabToggle = () => {
+    let kebab = cy.get('#pf-dropdown-toggle-id-11');
```
ZitaNemeckova, line 14: Same as above :)

test/cypress/integration/namespace-edit.js
```diff
@@ -0,0 +1,84 @@
...
+      return false;
+    });
+    createNamespace();
+    cy.menuGo('Collections > Namespaces');
+    viewNamespaceDetail();
+    kebabToggle();
+    editNamespace();
+  });
+
+  describe('the name field', () => {
```
ZitaNemeckova, line 67: This seems to be extra and breaks the readabilitty of the error message a bit.

test/cypress/integration/namespace-edit.js
```diff
@@ -0,0 +1,84 @@
...
+    it('tests that the name field is disabled from editing', () => {
+      checkName();
+    });
+  });
+
+  describe('the company name field', () => {
```
ZitaNemeckova, line 73: Same as above.

test/cypress/integration/namespace-edit.js
```diff
@@ -0,0 +1,84 @@
...
+    it('tests the company name for errors', () => {
+      checkCompanyName();
+    });
+    it('saves a new company name', () => {
+      saveCompanyName();
+    });
+    it('checks field in namespace detail', () => {
```
ZitaNemeckova, line 80: This should be a part of previous it.

test/cypress/integration/namespace-edit.js
```diff
@@ -0,0 +1,84 @@
...
+      cy.get('.pf-c-title').contains('Company name');
```
ZitaNemeckova, line 81: This should be a check/test`cy.get('.pf-c-title').should('contain','Company name');`

test/cypress/integration/namespace-edit.js
```diff
@@ -0,0 +1,84 @@
...
+  };
+
+  let saveCompanyName = () => {
+    getCompanyName()
+      .clear()
+      .type('Company name');
+    saveButton();
+    getUrl().should('eq', 'http://localhost:8002/ui/my-namespaces/testns1');
+  };
+  beforeEach(() => {
```
ZitaNemeckova, line 54: Please try to write the test the way you can use `before` instead of `beforeEach` to save a bit of a testing time :)

test/cypress/support/commands.js
```diff
@@ -27,6 +27,10 @@
 import shell from 'shell-escape-tag';
 var urljoin = require('url-join');
 
+Cypress.Commands.add('dataCy', value => {
```
ZitaNemeckova, line 30: This is never used. Plus we cannot use `data-cy` anyway :(

src/components/patternfly-wrappers/stateful-dropdown.tsx
```diff
@@ -69,6 +69,7 @@ export class StatefulDropdown extends React.Component<IProps, IState> {
         position={position || DropdownPosition.right}
         autoFocus={false}
         aria-label={ariaLabel}
+        // data-cy='edit-button'
```
ZitaNemeckova, line 72: Please erase commented code.

src/components/cards/namespace-card.tsx
```diff
@@ -48,7 +48,9 @@ export class NamespaceCard extends React.Component<IProps, {}> {
 
         {namespaceURL && (
           <CardFooter>
-            <Link to={namespaceURL}>View collections</Link>
+            <Link data-cy='view-collections' to={namespaceURL}>
```
ZitaNemeckova, line 51: This is never used and `data-cy` is not working on GH Actions :(

ShaiahWren:
@ZitaNemeckova Very pleased to see that my test has passed :) There are a few more steps before this test is complete, but I thought I'd submit this for a code review. There are many functions ~ let me know if I can reorganize them better. Thank you! :)

test/cypress/integration/namespace-edit.js
```diff
@@ -0,0 +1,196 @@
...
+    checkName();
+  });
+  it('tests the company name for errors', () => {
+    checkCompanyName();
+  });
+  it('saves a new company name', () => {
+    saveCompanyName();
+    cy.get('.pf-c-title').should('contain', 'Company name');
+  });
+  it.only('tests the namespace owners field', () => {
```
ZitaNemeckova, line 183: Please remove `only`.

test/cypress/integration/namespace-edit.js
```diff
@@ -0,0 +1,196 @@
...
+      .click();
+  };
+  let editNamespace = () => {
+    return cy.contains('Edit namespace').click();
+  };
+
+  let saveButton = () => {
+    return cy.contains('Save');
+  };
+  let getUrl = () => {
```
ZitaNemeckova, line 25: This doesn't need to be a function.

ZitaNemeckova:
@ShaiahWren In general for using variables it's good when you call it twice or more times.

test/cypress/integration/namespace-edit.js
```diff
@@ -0,0 +1,196 @@
...
+  let getUrlField = () => {
+    return cy.get('div.useful-links div.link-url #url').click();
+  };
+  let linksHelper = () => {
+    return cy.get('#links-helper');
+  };
+  let checkLink = () => {
+    cy.get('div.link > a')
+      .should('contain', "Link to Shaiah's website")
+      .and('have.attr', 'href', 'https://shaiahwren.com/');
```
ZitaNemeckova, line 129: A neutral `https://example.com` url would be better. It's usually used when a random url link is needed. And it says that the use is allowed for everyone.

src/loaders/standalone/standalone-loader.tsx
```diff
@@ -344,7 +344,6 @@ class App extends React.Component<RouteComponentProps, IState> {
         url:
           'https://access.redhat.com/documentation/en-us/red_hat_ansible_automation_platform/',
         external: true,
-        'data-cy': 'documentation-tab',
```
ZitaNemeckova, line 347: 👍 for cleaning.

ShaiahWren:
The test should be complete :) Lmk any other edits you'd like me to make before I merge.

test/cypress/integration/namespace-edit.js
```diff
@@ -0,0 +1,240 @@
...
+    });
+    createNamespace();
+    cy.menuGo('Collections > Namespaces');
+    viewNamespaceDetail();
+    kebabToggle();
+    editNamespace();
+  });
+
+  it('tests that the name field is disabled from editing', () => {
+    checkName();
```
himdel, line 213: these make the tests harder to read..

```js
const checkName = () => { body of the test };

...lots of code...

it('tests that the name field is disabled from editing', () => {
    checkName();
});
```

compare with simple

```js
it('tests that the name field is disabled from editing', () => {
  body of the test
});
```

where you immediately see what the test actually does, without having to look up other functions :).


Can you please only use functions like this for shared functionality used by multiple tests?

(Second, related point, if the function is a one liner, it may be better to have a copy than to have to look it up as well, unless it encodes some complex shared logic.)

test/cypress/integration/namespace-edit.js
```diff
@@ -0,0 +1,240 @@
...
+  let getEditTab = () => {
+    return cy
+      .get(
+        'ul.pf-c-tabs__list > li.pf-c-tabs__item > button > span.pf-c-tabs__item-text',
+      )
+      .contains('Edit resources')
+      .click();
+  };
+
+  getTextField = () => {
```
himdel, line 178: global function? (missing let or `const`)

test/cypress/integration/namespace-edit.js
```diff
@@ -0,0 +1,240 @@
...
+    kebabToggle();
+    editNamespace();
+    getEditTab();
+    getTextField().should('contain', 'Editing the readme file');
+  };
+
+  beforeEach(() => {
+    cy.visit(baseUrl);
+    cy.login(adminUsername, adminPassword);
+    cy.on('uncaught:exception', (err, runnable) => {
```
himdel, line 202: this one is surprising, why is it needed?

ShaiahWren:
@himdel Thanks for the great advice! :) 
I removed most of the one-liners from functions. I kept some functions that are uses multiple times in one test, or across tests. Let me know if you feel that the remaining ones are appropriate.

test/cypress/integration/namespace-edit.js
```diff
@@ -0,0 +1,167 @@
...
+    cy.menuGo('Collections > Namespaces');
+    cy.get('a[href*="ui/repo/published/testns1"]').click();
+    kebabToggle();
+    cy.contains('Edit namespace').click();
+  });
+
+  it('tests that the name field is disabled from editing', () => {
+    return cy.get('#name').should('be.disabled');
+  });
+  it('tests the company name for errors', () => {
```
himdel, line 56: one last nitpick: can you add newlines between `it`?

It makes it easier to visually tell where one test ends and another begins :)

```js
it('...', () => {
   ...
});

it('...', () => {
   ...
});
```

test/cypress/integration/namespace-edit.js
```diff
@@ -0,0 +1,167 @@
...
+      )
+      .contains('Edit resources')
+      .click();
+  };
+
+  let getTextField = () => {
+    return cy.get('div.pf-c-form__group-control > textarea.pf-c-form-control');
+  };
+
+  beforeEach(() => {
```
himdel, line 43: https://github.com/ansible/ansible-hub-ui/pull/540#discussion_r653491283 still seems unresolved...

is it possible to write the tests in a way that uses `before` only? Unless we need to separately login and create a namespace for each test?

himdel:
Thanks, this looks about right :) Added two more comments, but we're close :)

test/cypress/integration/namespace-edit.js
ShaiahWren: No problem! My last push will hopefully solve this.

test/cypress/integration/namespace-edit.js
ShaiahWren: Point well taken... I remember talking to Zita about before() vs beforeEach() and we couldn't get it working with before() so she told me to just merge. As it's written now I 100% agree that it's not an efficient test, and there is no reason we need separate logins. 
I can do a bit more research and try to make it work!

himdel: Feel free to merge as is :) We can change to `before` once we're doing login by setting cookies.

The issue is that cypress clears all cookies before every test ([docs](https://docs.cypress.io/api/cypress-api/cookies)).
And the test needs to PUT on Save, to get a response with the error message.

ShaiahWren:
pr. merged :)

PR #543, opened 2021-06-16
[stable-prod] Take down MaintenancePage

ZitaNemeckova:


himdel (approved):
Whitespace is off, otherwise LGTM

PR #534, opened 2021-06-14
Github actions follow-up: try to fix deploy-cloud on master, dev-release

himdel:
Follow-up to #526 (https://github.com/ansible/ansible-hub-ui/pull/526#issuecomment-860753116)

dev-release: errors on `GitHub Releases requires a tag`
because we're just adding the tag in a previous step, it can't figure out which tag to use,
providing via `input_tag_name`

deploy-cloud: errors on `npm ERR! cb() never called!` (on master, but works on prod-beta)
this is a generic "we don't know what's wrong" npm error,
the hive mind suggests a local caching issue, or a http issue, trying to add `npm cache verify` and see what happens.

Cc @newswangerd @ZitaNemeckova

PR #532, opened 2021-06-14
[prod-beta] Test deploy workflow

himdel:
This is a part of #526, for testing deploy on prod-beta.

Details there, this only contains the `deploy-cloud` workflow, updated ssh key, and removes .travis.yml.

PR #526, opened 2021-06-11
Migrate from travis to github actions

himdel:
This creates 4 new github action workflows, to replace our .travis.yml:

* "Dev release"
  * `.github/workflows/dev-release.yml`
  * runs on pushes to `master` only
  * updates the `dev` tag
  * deploys a github release artifact, built with `BRANCH=master` => `betaBranch = true`

* "Deploy cloud.redhat.com"
  * `.github/workflows/deploy-cloud.yml`
  * runs on pushes to `master`, `master-stable`, `prod-beta`, `prod-stable`
  * runs our [deploy](https://github.com/ansible/ansible-hub-ui/blob/master/package.json#L90-L93) build task, and [RedHatInsights/insights-frontend-builder-common/src/bootstrap.sh](https://raw.githubusercontent.com/RedHatInsights/insights-frontend-builder-common/master/src/bootstrap.sh)
  * provides these travis env vars (in addition to `COMMIT_AUTHOR_{EMAIL,USERNAME}` and `REPO` required by the script, and `NODE_OPTIONS`)
```
      TRAVIS_BRANCH=`sed 's/^refs\/heads\///' <<< $GITHUB_REF`
      TRAVIS_BUILD_NUMBER: ${{ env.GITHUB_RUN_NUMBER }}
      TRAVIS_BUILD_WEB_URL: ${{ env.GITHUB_SERVER_URL }}/${{ env.GITHUB_REPOSITORY }}/actions/runs/${{ env.GITHUB_RUN_ID }}
      TRAVIS_COMMIT_MESSAGE=`git log -1 --format=%B`
      TRAVIS_EVENT_TYPE: "push"
      TRAVIS_PULL_REQUEST: "false"
      TRAVIS_TAG: ""
```

* "Stable release"
  * `.github/workflows/stable-release.yml`
  * runs on version-like **tag** pushes, explicitly excluding the `dev` tag and tags matching `**cloud**`
  * deploys a github release artifact, built with empty `BRANCH`
  * reuses `TRAVIS_COMMIT_MESSAGE` to set the release body (release name being the tag)

* "Generate package manifest"
  * `.github/workflows/update-manifest.yml`
  * runs on pushes to `master` and `stable-prod` (TODO `stable-prod` or `stable` don't exist, different branch? remove?)
  * runs [.travis/update_manifest.sh](https://github.com/ansible/ansible-hub-ui/blob/master/.travis/update_manifest.sh) (#176, #372)
[...]

PR #546, opened 2021-06-16
[prod-beta] update deploy-cloud workflow

himdel:
Follow-up to #526 (+ #534, #538) .. copying workflows to the relevant branches
Parent (+links to other branches): #544

This just updates the existing workflow added to prod-beta in #532 to match master.

NOTE: merging this will cause a build (`prod-beta`)


himdel:
Landed in prod-beta via a force-push from master-stable (4.4cloud02 release).

PR #555, opened 2021-06-21
Fix namespace create form test to be idempotent.

ironfroggy:
AAH-616 added the `namespace_form.js` test file. The last test, which creates a namespace, only works the first time the test is run. It requires resetting the database before another successful run.

I believe all tests should be idempotent.

This small change randomizes the name of the namespace created so the test can pass every time.


ZitaNemeckova (approved):
I like it 👍

PR #566, opened 2021-06-29
stable-release workflow: expect release to exist, only add assets

himdel:
The `stable-release` workflow is triggered by pushing a tag to one of the branches which contain the workflow (master, stable-4.*).

Right now, it expcets only the tag to exists, and tries to create a release and upload assets.

But, what we're actually doing is creating the tag by creating a github release manually.

Before: the workflow creates a new draft github release with the right assets, and fails because it can't undraft it as the release already exists.
After: the workflow expects a release to exist, and only uploads the assets

(Merging after discussing on https://github.com/ansible/ansible-hub-ui/pull/565,
testing a release on master, the removing it, and backporting to the other branches if it works.)

### DIFF
--- a/.github/workflows/stable-release.yml
+++ b/.github/workflows/stable-release.yml
@@ -1,5 +1,6 @@
 name: "Stable release"
 
+# creating a release creates a tag as well; this expects the release to exist
 on:
   push:
     tags:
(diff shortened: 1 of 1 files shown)


himdel:
Testing steps...

created a new release on the github release page
it triggered https://github.com/ansible/ansible-hub-ui/runs/2942298039?check_suite_focus=true
successfully uploaded the asset to the existing release :+1:

PR #574, opened 2021-06-29
Update Cypress to 7.6.0

himdel:
4.11.0 was released on 2020-07-21
7.6.0 was released on 2021-06-23
[changelog](https://docs.cypress.io/guides/references/changelog#7-6-0)

Updating :)

(This also removes code that causes cypress to ignore uncaught exceptions in namespace form.)

### DIFF
--- a/test/cypress/integration/namespace_form.js
+++ b/test/cypress/integration/namespace_form.js
@@ -28,9 +28,6 @@ describe('A namespace form', () => {
   beforeEach(() => {
     cy.visit(baseUrl);
     cy.login(adminUsername, adminPassword);
-    cy.on('uncaught:exception', (err, runnable) => {
-      return false;
-    });
     createNamespace();
     cy.menuGo('Collections > Namespaces');
     getCreateNamespace().click();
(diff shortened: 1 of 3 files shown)


himdel:
(
Originally, the problem we've hit seemed to point at https://stackoverflow.com/questions/23892547/what-is-the-best-way-to-trigger-onchange-event-in-react-js/46012210#46012210...

react 16 changes the way how change is triggered on inputs,
manually it can be done via

```
// company is the actual input element

var nativeInputValueSetter = Object.getOwnPropertyDescriptor(window.HTMLInputElement.prototype, "value").set;
nativeInputValueSetter.call(company, 'magicFxoo')

var ev2 = new Event('input', { bubbles: true });
company.dispatchEvent(ev2);
```

except this is fixed in https://github.com/cypress-io/cypress/commit/4a56ca83b3b6e19ec57d10a8160f54f513e8f8ec
which is already in 4.11

Still, makes sense to upgrade anyway.
)

himdel:
(Closing and reopening, to see if the issue from the first commit reappears or not.)

newswangerd (approved):

PR #587, opened 2021-07-04
Changed URL in Container Registry empty state

jerabekjiri:
Issue: https://issues.redhat.com/browse/AAH-744

Hello,
The new URL is still 404, but my guess is it's early access? :)

### DIFF
--- a/src/containers/execution-environment-list/execution_environment_list.tsx
+++ b/src/containers/execution-environment-list/execution_environment_list.tsx
@@ -80,7 +80,7 @@ class ExecutionEnvironmentList extends React.Component<
         variant='link'
         onClick={() =>
           window.open(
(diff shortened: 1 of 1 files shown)


himdel:
LGTM :+1:

Cc @cbudz the URL still 404s, do we have any timeline/tracking for that?

cbudz:
@himdel that URL will 404 until we publish with the release of the platform. Production URL for docs follows the convention: https://access.redhat.com/documentation/en-us/red_hat_ansible_automation_platform/1.2/html/<doc-title>/index so its the best we can do for providing a URL in advance of publication.

himdel:
@cbudz Thanks! The problem is this is also getting into the 4.3 release, I'm not sure that's synced with the platform?

himdel (approved):

cbudz:
@himdel 4.3 is part of 2.0 so we should be good.

PR #608, opened 2021-07-07
package-lock: update after #325

himdel:
Follow-up to #325, I forgot to include `package-lock.json` :)


newswangerd (approved):

himdel:
Hm, merging despite cypress error, to get dependabot unstuck :)

The error is the same as in https://github.com/ansible/ansible-hub-ui/pull/607/checks?check_run_id=3011797965
Fixing separately...

PR #611, opened 2021-07-07
User list fixes

jerabekjiri:
Hello,
I resolved some issues, but wasn't sure about a few of them, so I left comments.
Issue: https://issues.redhat.com/browse/AAH-350

[AAH-351](https://issues.redhat.com/browse/AAH-351): align kebab button to the right in table, removed right padding and aligned button right.
before:
subtask1_before
after:
subtask1_after

[AAH-352](https://issues.redhat.com/browse/AAH-352): align filter input + button (compound filter) with top pagination, created the user-management.scss file and added .toolbar class.
before:
subtask2_before
after:
subtask2_after

[AAH-353](https://issues.redhat.com/browse/AAH-353): removed bottom and top padding + removed the div element, because it's pointless now.
before:
subtask3_before
after:
subtask3_after

[AAH-354](https://issues.redhat.com/browse/AAH-354): 
Done this with ref pointing to CompoundFilter component 
`compoundFilterRef = React.createRef<CompoundFilter>();` and then invoking it onClear function here 
`  
private clearTextInput = () => {
    this.compoundFilterRef.current.onClear();
  };
` Function clearTextInput is passed as a callback prop and gets called inside clearAllFilters in CompoundFilter component once the user clicks on clear all filters [see commit here](https://github.com/ansible/ansible-hub-ui/pull/611/commits/7a6cef97742bcb7af537b2ac8113879e66cbb08e)

[AAH-355](https://issues.redhat.com/browse/AAH-355)
As mentioned this is an API issue, not sure what to do here?

[AAH-356](https://issues.redhat.com/browse/AAH-356): table header column first_name wraps instead truncating on smaller screens, done this by adding className prop  to SortTable component and passing patternfly wrap class.
before:
subtask6_before
after:
subtask6_after

[AAH-357](https://issues.redhat.com/browse/AAH-357): Already in order, but columns "Groups" and "Created" are missing in the filter dropdown?
subtask_7_already_done


src/containers/user-management/user-list.tsx
```diff
@@ -398,6 +403,10 @@ class UserList extends React.Component<RouteComponentProps, IState> {
   private get closeAlert() {
     return closeAlertMixin('alerts');
   }
+
+  private clearTextInput = () => {
+    this.compoundFilterRef.current.onClear();
```
himdel, line 408: this one looks problematic, abusing a ref to call a callback on a sibling component doesn't sound right

Maybe we should just exclude this from the PR for now? I'm working on moving this kind of state to the parent component.

src/containers/user-management/user-management.scss
```diff
@@ -0,0 +1,5 @@
+.toolbar {
```
himdel, line 1: This seems rather generic ... such a class will affect all toolbars on all the screens, not just Users

src/containers/user-management/user-list.tsx
```diff
@@ -250,6 +254,7 @@ class UserList extends React.Component<RouteComponentProps, IState> {
           title: 'First name',
           type: 'alpha',
           id: 'first_name',
+          className: 'pf-m-wrap'
```
himdel, line 257: (nitpick: prefer to add trailing commas, fewer problems for later :)) (also, doesn't `npm run prettier` fix this?)

src/containers/user-management/user-list.tsx
jerabekjiri: Using the ref was kind of a workaround, that sounds like a better solution to control state from the user-list and pass it as a prop

src/containers/user-management/user-list.tsx
jerabekjiri: fixed, forgot about prettier :)

src/containers/user-management/user-management.scss
jerabekjiri: changed to `.user-list-toolbar`

jerabekjiri:
@himdel pls check
Fixed the name of the toolbar class and removed the onClear ref functionality. 
Also, should I create a new task in Jira for clearing out the text field for later?

himdel:
> Fixed the name of the toolbar class and removed the onClear ref functionality.

Thanks, the changes look good :)

> Also, should I create a new task in Jira for clearing out the text field for later?

No need, we can just move [AAH-354](https://issues.redhat.com/browse/AAH-354) back to TODO / Backlog :).

---

Testing...

* [x] AAH-351 (align kebab) - LGTM :+1: 
* [x] AAH-352 (filter & button) - LGTM :+1: (container registry & groups should get the same treatment, there may be an issue already?)
* [x] AAH-353 (removed padding) - LGTM :+1: 
* AAH-354 (clear all filters) - skipping :+1:  https://github.com/ansible/ansible-hub-ui/pull/611#discussion_r667529405
* AAH-355 (API issue) - nothing to do here indeed
* [x] AAH-356 (wrap headers) - LGTM :+1: 
* AAH-357 (order of columns) - not addressed in this PR

LGTM :+1: :)

---

> [AAH-355](https://issues.redhat.com/browse/AAH-355) As mentioned this is an API issue, not sure what to do here?

Agreed, you can't fix that one in this PR, the API repo is [ansible/galaxy_ng](https://github.com/ansible/galaxy_ng), so it would be a separate PR there :).
(You may want to move the issue back from "Code review" for now?)

> [AAH-357](https://issues.redhat.com/browse/AAH-357): Already in order, but columns "Groups" and "Created" are missing in the filter dropdown?

Either this was wrong before and fixed already (then we can just close that issue), or maybe this is referring to adding the other columns .. @ZitaNemeckova do you know if we have more info there?

himdel (approved):
LGTM :+1: 

(Just remember this doesn't close AAH-354, AAH-355, AAH-357.)

ZitaNemeckova:
@jerabekjiri @himdel AAH-357 is already fixed. No need to add more columns. Please move that issue to Obsolete :) 

AAH-355 is API issue (updated with link to PR that fixed same issue for namespaces). I think it should be easy to fix. If you feel like it you can go for it or ask Shaiah if they can do it.

ZitaNemeckova (approved):

PR #619, opened 2021-07-08
cloud.redhat.com -> console.redhat.com

himdel:
Related to https://issues.redhat.com/browse/AAH-749

This updates all cloud.redhat.com references to point to console.redhat.com instead.

Keeping the cloud logic for now, as the transition will take a while (hence the FIXME for later).

Note: actually test deploy once possible, if `@redhat-cloud-services/frontend-components-config` has to be updated, we need webpack 5 first

(Dev mode `ci.foo.redhat.com` is not affected by this.)

### DIFF
--- a/src/constants.tsx
+++ b/src/constants.tsx
@@ -155,10 +155,16 @@ export class Constants {
     'galaxy-qa.ansible.com',
   ];
   static DOWNSTREAM_HOSTS = [
+    // FIXME 2021-09: remove obsolete cloud* references
     'cloud.redhat.com',
     'cloud.stage.redhat.com',
     'ci.cloud.redhat.com',
     'qa.cloud.redhat.com',
+
+    'console.redhat.com',
+    'console.stage.redhat.com',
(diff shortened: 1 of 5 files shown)


ZitaNemeckova (approved):
LGTM 👍 We need to keep in mind `prod-beta` testing time.

src/constants.tsx
```diff
@@ -155,10 +155,15 @@ export class Constants {
     'galaxy-qa.ansible.com',
   ];
   static DOWNSTREAM_HOSTS = [
+    // FIXME 2020-09: remove obsolete cloud* references
```
himdel, line 158: *2021

ZitaNemeckova:
So, the DOWNSTREAM_HOSTS change doesn't actually change anything for insights mode, as it only affects the remote repositories screen.

*But*, in order to see the sync button for remotes from console.redhat.com, standalone instances need the updated list => adding backport labels.

PR #620, opened 2021-07-11
Fix cypress group_management and namespace-edit failures

himdel:
Currently we have 2 sporadic test failures:

[This run](https://github.com/ansible/ansible-hub-ui/runs/3021216894?check_suite_focus=true) shows both, but the recent red runs in https://github.com/ansible/ansible-hub-ui/actions all fail with one or the other.

---

> `group_management.js`: Hub Group Management Tests "before each" hook for "admin user can add/remove permissions to/from a group"
> The following error originated from your application code, not from Cypress. It was caused by an unhandled promise rejection.
> Request failed with status code 403

So, the 3rd spec is failing in beforeEach, because a request 403s.
That's because the *2nd* test ends in a `deleteUser` call, which only waits for the API response, but doesn't wait for the redirect to the users list screen. (So the redirect happens during the next test's `beforeEach`, when the cookie is gone, hence the 403.)

=> Adding a wait for the users list screen to load to the end of `deleteUser` - by waiting for the list users API request (since the modal is still on the list screen, detecting the screen by title didn't quite work :)).

And doing the same thing for `createUser` - locally I'm also seeing a failure where `addUserToGroup` calls `menuGo` to Groups *before* the previous `createUser` handler navigates to Users, causing the test to expect Groups, while ending up on the Users page => adding a wait after adding a user, to only return after navigation.

---

> `namespace-edit.js`: Edit a namespace tests the namespace owners field
> Timed out retrying after 4050ms: `cy.click()` failed because this element is detached from the DOM.
> `&lt;li id=&quot;select-option-16257619708947hw15daewze&quot; role=&quot;presentation&quot; class=&quot;pf-c-select__menu-wrapper&quot;&gt;...&lt;/li&gt;`

When filling out the owner group field, making sure that:

* a group always exists that can be autocompleted
* we're waiting for the API autocomplete requests
* using the group name to autocomplete, ensuring the right wrapper to click is found that way
 
---

More fixes:

`cy.visit(baseUrl) + cy.login(...)` can fail: when already logged in, baseUrl is Containers - adding an explicit `cy.visit('/ui/login')`.

Server error responses fail in `mapErrorMessages` because they only contain the error as string, not a json object with an array of errors - adding support for that.

Update `cy.server()` & `cy.route()` calls, both are deprecated, use `cy.intercept()` instead. ([docs](https://docs.cypress.io/api/commands/intercept))

Remove any `after` and `afterAll`, move cleanup to `before`.

Simplify `deleteTestUsers` and `deleteTestGroups` helpers by reusing the `cy.galaxykit` helper, and having it yield a list of nonempty lines.

### DIFF
--- a/src/utilities/map-error-messages.ts
+++ b/src/utilities/map-error-messages.ts
@@ -3,6 +3,13 @@
 
 export function mapErrorMessages(err) {
   const messages: any = {};
+
+  // 500 errors only have err.response.data string
+  if (typeof err.response.data === 'string') {
+    messages['__nofield'] = err.response.data;
+    return messages;
+  }
+
   for (const e of err.response.data.errors) {
     if (e.source) {
       messages[e.source.parameter] = e.detail;
@@ -12,5 +19,6 @@ export function mapErrorMessages(err) {
       messages['__nofield'] = e.detail || e.title;
     }
   }
+
   return messages;
 }
--- a/test/cypress/integration/menu.js
+++ b/test/cypress/integration/menu.js
@@ -14,10 +14,6 @@ describe('Hub Menu Tests', () => {
     'User Access > Groups',
   ];
 
-  beforeEach(() => {
-    cy.visit(baseUrl);
-  });
-
   it('admin user sees complete menu', () => {
     cy.login(adminUsername, adminPassword);
     menuItems.forEach(item => cy.menuPresent(item));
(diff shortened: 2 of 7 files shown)


himdel:
Cc @ZitaNemeckova , this should fix the test failures on master :)

ZitaNemeckova (approved):
LGTM 👍

PR #626, opened 2021-07-12
Submitting create user request twice fix

jerabekjiri:
Issue: [AAH-654](https://issues.redhat.com/browse/AAH-654 )

Fixed the part of the issue where the form submits the POST request twice. :)


himdel (approved):
:+1: LGTM, the change makes sense, type=submit button does submit the form

Tested in the UI :)

PR #635, opened 2021-07-13
Warning message for inactive token

jerabekjiri:
Issue: [AAH-727](https://issues.redhat.com/browse/AAH-727)

Added a warning message after the 'Load token' button is clicked. 
Note: the warning message _The token will expire after 30 days of inactivity. To renew the token, run the command below._ will be most likely be changed when I get the right text from Chris. 

Can you please check if this is the correct design approach with the code block example inside the alert?

after token is loaded:
warning_message

expanded codeblock:
warning_message_expanded

### DIFF
--- a/src/containers/token/token-insights.tsx
+++ b/src/containers/token/token-insights.tsx
@@ -94,7 +94,8 @@ class TokenPage extends React.Component<RouteComponentProps, IState> {
               style={{ paddingTop: 'var(--pf-global--spacer--md)' }}
             >
               <span>
-                Run the command below periodically to prevent your token from expiring.
+                The token will expire after 30 days of inactivity. Run the
+                command below periodically to prevent your token from expiring.
               </span>
               <ClipboardCopy
                 isCode
--- a/src/containers/token/token-insights.tsx
+++ b/src/containers/token/token-insights.tsx
@@ -94,8 +94,7 @@ class TokenPage extends React.Component<RouteComponentProps, IState> {
               style={{ paddingTop: 'var(--pf-global--spacer--md)' }}
             >
               <span>
-                The token will expire after 30 days of inactivity. To renew the
-                token, run the command below.
+                Run the command below periodically to prevent your token from expiring.
               </span>
               <ClipboardCopy
                 isCode
--- a/src/containers/token/token-insights.tsx
+++ b/src/containers/token/token-insights.tsx
@@ -1,11 +1,16 @@
 import * as React from 'react';
 
 import { withRouter, RouteComponentProps, Link } from 'react-router-dom';
-import { ClipboardCopy, Button } from '@patternfly/react-core';
+import {
+  ClipboardCopy,
+  ClipboardCopyVariant,
+  Button,
+} from '@patternfly/react-core';
 
 import { Paths } from 'src/paths';
 import { BaseHeader, Main } from 'src/components';
 import { getRepoUrl } from 'src/utilities';
+import { AppContext } from 'src/loaders/app-context';
 
 interface IState {
   tokenData: {
@@ -37,7 +42,9 @@ class TokenPage extends React.Component<RouteComponentProps, IState> {
   }
 
   render() {
+    const { user } = this.context;
     const { tokenData } = this.state;
+    const renewTokenCmd = `curl https://sso.redhat.com/auth/realms/redhat-external/protocol/openid-connect/token -d grant_type=refresh_token -d client_id="${user.username}" -d refresh_token=\"${tokenData?.refresh_token ?? '{{ user_token }}'}\" --fail --silent --show-error --output /dev/null`;
 
     return (
       <React.Fragment>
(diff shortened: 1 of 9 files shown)


jerabekjiri:
Just noticed, in the token renew command
`curl https://sso.redhat.com/auth/realms/redhat-external/protocol/openid-connect/token -d grant_type=refresh_token -d client_id="{{ user_name }}" -d refresh_token="{{ user_token }}" --fail --silent --show-error --output /dev/null` 
should `{{ user_token }}` and `{{ user_token }}` be changed for real values?

ZitaNemeckova:
@sbuenafe-rh can you please have a look, thanks :)

@jerabekjiri I would say depends because you may not have the token generated when you read the message. So maybe just the `user_name`. WDYT @newswangerd ?

jerabekjiri:
> @jerabekjiri I would say depends because you may not have the token generated when you read the message. So maybe just the `user_name`.

@ZitaNemeckova Well the warning message is displayed after the primary 'Load token' button is clicked, so the token is always generated? Or should we just show the message whenever a user visits the API token page?

ZitaNemeckova:
@jerabekjiri You are right! I missed that it's displayed after clicking the button. In that case it makes sense to populate both with real value.

ZitaNemeckova:
@jerabekjiri please add a screenshot of whole page, please.

jerabekjiri:
@ZitaNemeckova Here you go
warning_message_page

I was looking into that test failure, not really sure what to do here.

ZitaNemeckova:
@sbuenafe-rh Can you please have a look at the added warning at https://github.com/ansible/ansible-hub-ui/pull/635#issuecomment-880578071 from UX perspective. Thank you :)

jerabekjiri:
According to Chris Budzilowicz, the text in UI is good for now.

ZitaNemeckova:
Susan give it 👍 too.

himdel (commented):
Let's wait for Susan's feedback, I'd put the message *under* the API token section, closer to where the token is shown.

sbuenafe-rh:
@ZitaNemeckova @jerabekjiri After re-reading this PR and speaking with @himdel I understand that we cannot track the token, therefore we should not be using the inline alert. The inline alert causes me to think that the user needs to take immediate action. I would place the token inactivity expiration message in the API token card below the copy to clipboard component. :)

jerabekjiri:
- Moved the message below the token copy to clipboard component. 
- Changed it to an all-time visible message and once the user loads the token, the placeholder value (user_token) in the command is updated to the real token value. 

token_updated

src/containers/token/token-standalone.tsx
```diff
@@ -1,10 +1,17 @@
 import * as React from 'react';
 
 import { withRouter, RouteComponentProps } from 'react-router-dom';
-import { ClipboardCopy, Button } from '@patternfly/react-core';
+import {
+  ClipboardCopy,
+  ClipboardCopyVariant,
+  Button,
+  Alert,
```
ZitaNemeckova, line 8: `Alert` and `AlertActionCloseButton ` aren't used. Please remove them.

jerabekjiri: Removed :)

himdel:
We also need to change the url to use the actual current URL.
Most instances won't be running on sso.redhat.com

newswangerd (changes requested):
The renew token message only applies to insights mode, not standalone mode. These changes should be put on `token-insights.tsx`.

The changes look good tho

jerabekjiri:
Moved the warning message to the insights mode and removed it in standalone. 
Screenshot from 2021-07-20 17-54-56

jerabekjiri:
@newswangerd @himdel @ZitaNemeckova check changes pls :)

newswangerd:
Waiting on confirmation from the content team about how the renew script functions. Will update my review soon.

src/containers/token/token-insights.tsx
```diff
@@ -82,6 +89,22 @@ class TokenPage extends React.Component<RouteComponentProps, IState> {
             ) : (
               <Button onClick={() => this.loadToken()}>Load token</Button>
             )}
+            <div
+              className='pf-c-content'
+              style={{ paddingTop: 'var(--pf-global--spacer--md)' }}
+            >
+              <span>
+                The token will expire after 30 days of inactivity. To renew the
```
newswangerd, line 97: So, this script only works before the token expires. Lets change the wording to something like "Run the command below periodically to prevent your token from expiring."

newswangerd (changes requested):

cbudz:
@newswangerd Yes, I think that language is fine.

newswangerd:
@jerabekjiri Oh, I didn't mean for you to get rid of the first part. The full text should be "The token will expire after 30 days of inactivity. Run the command below periodically to prevent your token from expiring."

Sorry for the confusion

jerabekjiri:
@newswangerd My fault, I didn't even think about it and changed it in hurry :facepalm: Should be fixed now :)

newswangerd (approved):
Looks good!

himdel (approved):
Works for me, LGTM :)

himdel:
(when doing a manual backport, also add #711 to make prettier happy :))

PR #637, opened 2021-07-14
Approval Dashboard issues 

jerabekjiri:
Issue [AAH-359](https://issues.redhat.com/browse/AAH-359)

- changed dropdown option "repository" to "status":
Found a little problem here, the `repository` attribute is used as a param in API, changing it to `status` breaks the API filtering. Fixed it by passing a new object `updatedCollectionParams`, which doesn't contain `status` attr, and also added `repository` attribute that is expected. 
`
  const updatedCollectionParams =
      'status' in this.state.params
        ? _.omit(
            {
              repository: this.state.params['status'],
              ...this.state.params,
            },
            ['status'],
          )
        : this.state.params;
`
Maybe a cleaner solution would be to solve it on the backend and rename it API attribute to status?

- Staging should be a dropdown option if it's a chip as it is currently (needs review goes into staging, approved goes into published): 
I think this is already solved, but with this change `niceNames={{ staging: 'Needs review' }}` It's now "Needs review" chip.  

before: 
filter_before

after:
staging_after

According to the issue, the dropdown option `status` should be the second option, but it's third item in the dropdown filter? Should I change this? 
expanded_dropdown

### DIFF
--- a/src/components/patternfly-wrappers/applied-filters.tsx
+++ b/src/components/patternfly-wrappers/applied-filters.tsx
@@ -69,7 +69,7 @@ export class AppliedFilters extends React.Component<IProps, {}> {
                 updateParams(ParamHelper.deleteParam(params, key, v))
               }
             >
-              {v}
+              {niceNames[v] || v}
             </Chip>
           ))}
         </ChipGroup>
--- a/src/components/patternfly-wrappers/applied-filters.tsx
+++ b/src/components/patternfly-wrappers/applied-filters.tsx
@@ -19,6 +19,7 @@ interface IProps {
    * when displaying the param field name
    */
   niceNames?: object;
+  niceValues?: object;
   style?: React.CSSProperties;
   className?: string;
 }
@@ -49,7 +50,7 @@ export class AppliedFilters extends React.Component<IProps, {}> {
   }
 
   private renderGroup(key: string) {
-    const { niceNames, params, updateParams } = this.props;
+    const { niceNames, niceValues, params, updateParams } = this.props;
 
     let chips;
 
@@ -69,7 +70,7 @@ export class AppliedFilters extends React.Component<IProps, {}> {
                 updateParams(ParamHelper.deleteParam(params, key, v))
               }
             >
-              {niceNames[v] || v}
+              {niceValues?.[key]?.[v] || v}
             </Chip>
           ))}
         </ChipGroup>
--- a/src/containers/certification-dashboard/certification-dashboard.tsx
+++ b/src/containers/certification-dashboard/certification-dashboard.tsx
@@ -183,7 +183,13 @@ class CertificationDashboard extends React.Component<
                   }
                   params={params}
                   ignoredParams={['page_size', 'page', 'sort']}
-                  niceNames={{ staging: 'Needs review' }}
+                  niceValues={{
+                    status: {
+                      [Constants.PUBLISHED]: 'Approved',
+                      [Constants.NEEDSREVIEW]: 'Needs Review',
+                      [Constants.NOTCERTIFIED]: 'Rejected',
+                    },
+                  }}
                 />
               </div>
               {loading ? (
--- a/src/containers/certification-dashboard/certification-dashboard.tsx
+++ b/src/containers/certification-dashboard/certification-dashboard.tsx
@@ -23,7 +23,7 @@ import {
   ExclamationCircleIcon,
   CheckCircleIcon,
 } from '@patternfly/react-icons';
-import _ from 'lodash';
+import { omit } from 'lodash';
 
 import { CollectionVersionAPI, CollectionVersion, TaskAPI } from 'src/api';
 import { filterIsSet, ParamHelper } from 'src/utilities';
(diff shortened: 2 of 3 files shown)


src/components/patternfly-wrappers/applied-filters.tsx
```diff
@@ -69,7 +69,7 @@ export class AppliedFilters extends React.Component<IProps, {}> {
                 updateParams(ParamHelper.deleteParam(params, key, v))
               }
             >
-              {v}
+              {niceNames[v] || v}
```
himdel, line 72: hm, this would mix the translations of param field names with translations of param field values

what about adding a separate `niceValues`?

src/containers/certification-dashboard/certification-dashboard.tsx
```diff
@@ -22,6 +23,7 @@ import {
   ExclamationCircleIcon,
   CheckCircleIcon,
 } from '@patternfly/react-icons';
+import _ from 'lodash';
```
himdel, line 26: `import { omit } from 'lodash'` would be more consistent with our other uses of lodash (but both work :))

src/components/patternfly-wrappers/applied-filters.tsx
himdel: Actually, this may be a bit more complicated than that, because the set of values can be different for each name, right?

I'm thinking.. `{ niceNames?.[key]?.[v] || v }`
(And passing the prop as `niceValues={{ status: { staging: 'Needs review' } }}`)

jerabekjiri: Cool, I like it, thanks :)

src/containers/certification-dashboard/certification-dashboard.tsx
```diff
@@ -141,8 +143,8 @@ class CertificationDashboard extends React.Component<
                             title: 'Collection Name',
                           },
                           {
-                            id: 'repository',
-                            title: 'Repository',
+                            id: 'status',
```
ZitaNemeckova, line 146: Instead of `omit` logic keep `id: status`. `id` doesn't need to change when `title` does.

src/containers/certification-dashboard/certification-dashboard.tsx
```diff
@@ -207,7 +216,7 @@ class CertificationDashboard extends React.Component<
 
   private renderTable(versions, params) {
     if (versions.length === 0) {
-      return filterIsSet(params, ['namespace', 'name', 'repository']) ? (
+      return filterIsSet(params, ['namespace', 'name', 'status']) ? (
```
ZitaNemeckova, line 219: Keep `repository`.

src/containers/certification-dashboard/certification-dashboard.tsx
```diff
@@ -518,8 +527,19 @@ class CertificationDashboard extends React.Component<
...
+      'status' in this.state.params
+        ? omit(
+            {
+              repository: this.state.params['status'],
+              ...this.state.params,
+            },
+            ['status'],
+          )
+        : this.state.params;
+
```
ZitaNemeckova, line 540: Remove in favor of https://github.com/ansible/ansible-hub-ui/pull/637/files#r672151414 .

src/containers/certification-dashboard/certification-dashboard.tsx
```diff
@@ -22,6 +23,7 @@ import {
   ExclamationCircleIcon,
   CheckCircleIcon,
 } from '@patternfly/react-icons';
+import { omit } from 'lodash';
```
ZitaNemeckova, line 26: Unused. Please remove.

ZitaNemeckova (approved):
One unused import. Otherwise LGTM 👍

newswangerd:
The status in the dropdown is still showing up as the repository name.

Screen Shot 2021-07-22 at 1 34 39 PM

jerabekjiri:
@newswangerd Thanks for pointing that out, I thought we can't set "nice name" to patternfly select after reading this
`// patternfly does not allow for us to set a display name aside from the ID...`
but didn't notice this applies only to multiple select. 
Should be fixed now
nicename_dropdown

newswangerd (approved):
Looks like it's working. Nice job!

src/containers/certification-dashboard/certification-dashboard.tsx
```diff
@@ -181,6 +182,16 @@ class CertificationDashboard extends React.Component<
...
                   ignoredParams={['page_size', 'page', 'sort']}
+                  niceValues={{
+                    repository: {
+                      [Constants.PUBLISHED]: 'Approved',
+                      [Constants.NEEDSREVIEW]: 'Needs Review',
+                      [Constants.NOTCERTIFIED]: 'Rejected',
+                    },
+                  }}
+                  niceNames={{
+                    repository: 'status',
```
himdel, line 193: this should match the option title.. `status` -> `Status`

himdel (approved):
Apart from the capital S, LGTM, tested :)

PR #618, opened 2021-07-08
[WIP] Cypress - cookie storage based login helper

MilanPospisil:
Thank you for a new Pull Request!

Please, make reviewers work easier by checking, if you are not introducing any technical debt:
https://github.com/ansible/ansible-hub-ui/blob/master/developer_guidelines.md

Describe the rationale and use case for this pull request.  Provide any background, examples, and images that provide further information to accurately describe what it is that you are adding to the repo.  Add subsections as necessary to organize and feel free to link and reference other PRs as necessary, but also include them in the links section below as a quick reference.

Guidelines:
* Keep Pull Request titles short and to the point, ideally under 72 characters
* Provide as much context/info in the description as necessary to get the reviewer up to the same domain knowledge level as yourself
* Keep code changes as short as possible and implementing a single feature/fix/refactoring, when possible

Links [Optional]
----------------

* http://documentation.for/library/that/I/am/adding
* [PR relevant issue or pull_request](#123)

Steps for Testing/QA [Optional]
-------------------------------

If there are any manual steps that you would like the reviewer(s) to take to verify your changes, please describe in detail the steps to reproduce the features added by the pull request, or the bug before and after the change.

PR #643, opened 2021-07-16
Reworking of ui login using cookie storage

MilanPospisil:
New commands cookieLogin and cookieLogout were created. Two tests were converted to use these new commands. CookieLogin stores sessionId and csrftoken in local storage and adds them when cookieLogin is called. It stores cookies per user.

### DIFF
--- a/test/cypress/integration/login.js
+++ b/test/cypress/integration/login.js
@@ -17,8 +17,6 @@ describe('Test cookieLogin for cookie storage', () => {
     cy.login(username, password);
     cy.contains(username);
     cy.logout();
-    // the wait is important here, otherwise it will creat occasional test fails
-    cy.wait(2000);
     cy.login(adminUsername, adminPassword);
     cy.contains(adminUsername);
   });
--- a/test/cypress/integration/login.js
+++ b/test/cypress/integration/login.js
@@ -1,6 +1,4 @@
 describe('Test cookieLogin for cookie storage', () => {
-  //let baseUrl = Cypress.config().baseUrl;
-  let baseUrl = '/';
   let adminUsername = Cypress.env('username');
   let adminPassword = Cypress.env('password');
   let username = 'nopermission';
@@ -8,7 +6,6 @@ describe('Test cookieLogin for cookie storage', () => {
 
   before(() => {
     cy.login(adminUsername, adminPassword);
-    cy.visit('');
     cy.deleteTestUsers();
     cy.galaxykit('user create', username, password);
     cy.logout();
@@ -54,7 +51,22 @@ describe('Test cookieLogin for cookie storage', () => {
     cy.cookieLogin(username, password);
     cy.contains(username);
     cy.cookieLogout();
+	cy.getCookies().then(cookies => {
+		let sessionid = null;
+		let csrftoken = null;
 
+		cookies.forEach(cookie => {
+			if (cookie.name == 'sessionid') sessionid = cookie.value;
+			if (cookie.name == 'csrftoken') csrftoken = cookie.value;
+		});
+		
+		cy.expect(sessionid).to.be.null;
+		cy.expect(csrftoken).to.be.null;
+	});
+	cy.getUserTokens((user_tokens) =>
+	{
+		cy.expect(user_tokens).to.eql({});
+	});
     cy.cookieLogin(adminUsername, adminPassword);
     cy.contains(adminUsername);
   });
--- a/test/cypress/integration/login.js
+++ b/test/cypress/integration/login.js
@@ -48,23 +48,9 @@ describe('Test cookieLogin for cookie storage', () => {
 
     cy.cookieLogin(adminUsername, adminPassword);
     cy.contains(adminUsername);
-
-    cy.cookieLogin(username, password);
-    cy.contains(username);
-
-    cy.cookieLogin(adminUsername, adminPassword);
-    cy.contains(adminUsername);
   });
 
   it('can cookieLogin with logout as admin or different user - this will force to login manualy every time', () => {
-    cy.cookieLogin(username, password);
-    cy.contains(username);
-    cy.cookieLogout();
-
-    cy.cookieLogin(adminUsername, adminPassword);
-    cy.contains(adminUsername);
-    cy.cookieLogout();
-
     cy.cookieLogin(username, password);
     cy.contains(username);
     cy.cookieLogout();
--- a/test/cypress/integration/login.js
+++ b/test/cypress/integration/login.js
@@ -10,26 +10,24 @@ describe('Test cookieLogin for cookie storage', () => {
     cy.login(adminUsername, adminPassword);
     cy.deleteTestUsers();
     cy.galaxykit('user create', username, password);
-	cy.logout();
+    cy.logout();
   });
 
   it('can login manually and logout as admin or different user', () => {
-	cy.login(username, password);
+    cy.login(username, password);
     cy.contains(username);
     cy.logout();
-	// the wait is important here, otherwise it will creat occasional test fails
-	cy.wait(2000);
+    // the wait is important here, otherwise it will creat occasional test fails
+    cy.wait(2000);
     cy.login(adminUsername, adminPassword);
     cy.contains(adminUsername);
   });
 
-  
   it('can login as user and store cookie', () => {
     cy.cookieLogin(username, password);
     cy.contains(username);
   });
 
-  
   it('can login as admin and store cookie', () => {
     cy.cookieLogin(adminUsername, adminUsername);
     cy.contains(adminUsername);
@@ -46,7 +44,7 @@ describe('Test cookieLogin for cookie storage', () => {
   });
 
   it('can cookieLogin without cookie removal between its as admin or different user', () => {
-	cy.cookieLogin(username, password);
+    cy.cookieLogin(username, password);
     cy.contains(username);
 
     cy.cookieLogin(adminUsername, adminPassword);
--- a/test/cypress/integration/login.js
+++ b/test/cypress/integration/login.js
@@ -51,22 +51,21 @@ describe('Test cookieLogin for cookie storage', () => {
     cy.cookieLogin(username, password);
     cy.contains(username);
     cy.cookieLogout();
-	cy.getCookies().then(cookies => {
-		let sessionid = null;
-		let csrftoken = null;
+    cy.getCookies().then(cookies => {
+      let sessionid = null;
+      let csrftoken = null;
 
-		cookies.forEach(cookie => {
-			if (cookie.name == 'sessionid') sessionid = cookie.value;
-			if (cookie.name == 'csrftoken') csrftoken = cookie.value;
-		});
-		
-		cy.expect(sessionid).to.be.null;
-		cy.expect(csrftoken).to.be.null;
-	});
-	cy.getUserTokens((user_tokens) =>
-	{
-		cy.expect(user_tokens).to.eql({});
-	});
+      cookies.forEach(cookie => {
+        if (cookie.name == 'sessionid') sessionid = cookie.value;
+        if (cookie.name == 'csrftoken') csrftoken = cookie.value;
+      });
+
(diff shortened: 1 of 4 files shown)


test/cypress/integration/login.js
```diff
@@ -0,0 +1,91 @@
+describe('Test cookieLogin for cookie storage', () => {
+  let baseUrl = Cypress.config().baseUrl;
+  let adminUsername = Cypress.env('username');
+  let adminPassword = Cypress.env('password');
+  let username = 'nopermission';
+  let password = 'n0permissi0n';
+
+  before(() => {
+    cy.visit(baseUrl);
+    cy.login(adminUsername, adminPassword);
```
himdel, line 10: don't `visit` `login` or `logout` here, no need

test/cypress/integration/login.js
```diff
@@ -0,0 +1,91 @@
...
+    cy.deleteTestUsers();
+    cy.galaxykit('user create', username, password);
+    cy.logout();
+  });
+
+  it('Test manual login and logout as admin or different user', () => {
+    cy.visit(baseUrl);
+    cy.login(username, password);
+    cy.visit(baseUrl);
```
himdel, line 19: login already does visit

test/cypress/integration/login.js
```diff
@@ -0,0 +1,91 @@
...
+    cy.contains(username);
+  });
+
+  it('Test login2 admin', () => {
+    cy.cookieLogin(adminUsername, adminUsername);
+    cy.visit(baseUrl);
+    cy.contains(adminUsername);
+  });
+
+  it('Test login3 user', () => {
```
himdel, line 40: how is this different from login1?

test/cypress/integration/login.js
```diff
@@ -0,0 +1,91 @@
...
+    cy.cookieLogin(username, password);
+    cy.visit(baseUrl);
+    cy.contains(username);
+  });
+
+  it('Test login4 admin', () => {
```
himdel, line 46: how is this different from login2?

test/cypress/integration/login.js
```diff
@@ -0,0 +1,91 @@
+describe('Test cookieLogin for cookie storage', () => {
```
himdel, line 1: Also make sure to keep the naming consistent.. `describe` should say what it's describing, `it` should continue a sentence

What about something like...

```diff
-describe('Test cookieLogin for cookie storage', ...
+describe('Login helpers', ...

-  it('Test manual login and logout as admin or different user', ...
+  it('can log in and out manually as admin or a different user', ...
```

test/cypress/integration/login.js
```diff
@@ -0,0 +1,91 @@
...
+  let baseUrl = Cypress.config().baseUrl;
```
himdel, line 2: no need for `baseUrl`, we can use `'/'`

test/cypress/integration/login.js
```diff
@@ -0,0 +1,91 @@
...
+  it('Test manual login and logout as admin or different user', () => {
+    cy.visit(baseUrl);
+    cy.login(username, password);
+    cy.visit(baseUrl);
+    cy.contains(username);
+    cy.logout();
+    cy.login(adminUsername, adminPassword);
+    cy.visit(baseUrl);
+    cy.contains(adminUsername);
+    cy.logout();
```
himdel, line 25: no need for the last `logout`

test/cypress/integration/login.js
```diff
@@ -0,0 +1,91 @@
...
+  it('Test cookieLogin without logout and cookie removal as admin or different user', () => {
+    cy.cookieLogin(username, password);
+    cy.visit(baseUrl);
+    cy.contains(username);
+
+    cy.cookieLogin(adminUsername, adminPassword);
+    cy.visit(baseUrl);
+    cy.contains(adminUsername);
+
+    cy.cookieLogin(username, password);
```
himdel, line 61: why twice?

test/cypress/integration/login.js
```diff
@@ -0,0 +1,91 @@
...
+    cy.visit(baseUrl);
+    cy.contains(username);
+
+    cy.cookieLogin(adminUsername, adminPassword);
+    cy.visit(baseUrl);
+    cy.contains(adminUsername);
+  });
+
+  it('Test cookieLogin with logout and cookie removal as admin or different user', () => {
```
himdel, line 70: cookieLogout+cookieLogin is always the same as cookieLogin, do we need this test at all?

test/cypress/support/commands.js
```diff
@@ -55,6 +55,37 @@ Cypress.Commands.add('menuGo', {}, name => {
...
+
+      user_tokens[username] = { sessionid, csrftoken };
+    });
+  } else {
+    let csrftoken = user_tokens[username].csrftoken;
+    let sessionid = user_tokens[username].sessionid;
+
+    cy.setCookie('csrftoken', csrftoken);
+    cy.setCookie('sessionid', sessionid);
+    cy.visit(Cypress.config().baseUrl);
```
himdel, line 85: this should actually just be `cy.visit('/')`, since `baseUrl` is set to `Cypress.config().baseUrl` (heh :)), `cy.visit('/')` does the same thing (https://github.com/ansible/ansible-hub-ui/pull/618#discussion_r667530926)

test/cypress/support/commands.js
```diff
@@ -55,6 +55,37 @@ Cypress.Commands.add('menuGo', {}, name => {
...
+
+Cypress.Commands.add('cookieLogin', {}, (username, password) => {
+  if (!user_tokens[username]) {
+    cy.login(username, password);
+    cy.getCookies().then(cookies => {
+      let sessionid;
+      let csrftoken;
+
+      cookies.forEach(cookie => {
+        if (cookie.name == 'sessionid') sessionid = cookie.value;
```
himdel, line 73: `npm run prettier` before merging, should fix the formatting (add newlines and blocks)

test/cypress/integration/login.js
```diff
@@ -0,0 +1,91 @@
...
+    cy.logout();
+    cy.login(adminUsername, adminPassword);
+    cy.visit(baseUrl);
+    cy.contains(adminUsername);
+    cy.logout();
+  });
+
+  it('Test login1 user', () => {
+    cy.cookieLogin(username, password);
+    cy.visit(baseUrl);
```
himdel, line 30: `cookieLogin` also already does `visit`, no need to do it twice

test/cypress/integration/login.js
MilanPospisil: You are testing contextual thing, so you need to check if login switching works correctly, so I log into user, then into admin, then back to user. 4th login back to admin may not be needed indeed.

test/cypress/integration/login.js
MilanPospisil: But without visit, it is not working, I cant find username in the right top screen and I dont know why.

test/cypress/integration/login.js
MilanPospisil: This is test if it is also working when cookies are not deleted, but rather set to new value.

test/cypress/integration/login.js
MilanPospisil: I though it is necessary to test if cookieLogout is really deleting cookies and working properly.

test/cypress/integration/login.js
himdel: There is no context between `it` calls.

MilanPospisil: There is context - cookie. One test is in one big it with logins with context, other is test with separate its for testing with no context. Both things should be tested. But if you think we not need to test it, I can delete it.

MilanPospisil:
> 
> 
> Don't. No waits with a random number please.
> 
> Do you need to wait for a request? For URL change?

test/cypress/integration/login.js
himdel: Ah then feel free to keep it, `visit` will wait for the page to load.

test/cypress/integration/login.js
```diff
@@ -0,0 +1,75 @@
+describe('Test cookieLogin for cookie storage', () => {
+  //let baseUrl = Cypress.config().baseUrl;
+  let baseUrl = '/';
```
himdel, line 3: dead code now

test/cypress/integration/login.js
himdel: I do see the point of the previous 4 `it`s,

But line 46-47 tests exactly the same thing as the `it` on line 35 - cookieLogin works with no cookie and cached user_tokens. Still makes sense as setup.

Then line 49-50 tests that cookieLogin works with preexisting cookie and cached user_tokens, that's a good thing to test, I agree.

But then 52-53 again tests that cookieLogin works with preexisting cookie and cached user_tokens, which is a dup of the previous.

And then 55-56 does it yet again.

So that's the part I don't understand :).

test/cypress/integration/login.js
```diff
@@ -0,0 +1,75 @@
...
+    cy.contains(adminUsername);
+
+    cy.cookieLogin(username, password);
+    cy.contains(username);
+
+    cy.cookieLogin(adminUsername, adminPassword);
+    cy.contains(adminUsername);
+  });
+
+  it('can cookieLogin with logout as admin or different user - this will force to login manualy every time', () => {
```
himdel, line 59: > this will force to login manualy every time

It won't? `user_tokens` still has all the cookies.

Additionally, if you change `cookieLogout` to not do anything, this test will still pass, so.. is it testing anything?

test/cypress/integration/login.js
himdel: So.. should we just test that cookieLogout + visit gives you the login screen?

test/cypress/integration/login.js
MilanPospisil: I deleted some of test where you got the point, rest must be discussed between us, there is clearly some misunderstanding :-)

test/cypress/integration/login.js
MilanPospisil: Solved.

test/cypress/integration/login.js
MilanPospisil: This works now, probably because of rebase? You did repaired the errors as you told me?

test/cypress/integration/login.js
MilanPospisil: Ok

test/cypress/integration/login.js
MilanPospisil: i dont know how to easily test it besides some additions to login command. However I removed some of the tests as you wanted, so probably we dont need to test it.

test/cypress/support/commands.js
MilanPospisil: Ok.

test/cypress/support/commands.js
MilanPospisil: ok.

test/cypress/integration/login.js
MilanPospisil: ok, so you want to test if cookies are deleted.

MilanPospisil: You are right, I forgot to clear user_tokens storage as well :-) Now Im forcing it to relogin manualy and Im testing that cookies and user_tokens are both clear.

test/cypress/integration/login.js
MilanPospisil: ok.

test/cypress/support/commands.js
himdel: Ah, sorry I though prettier did this automatically, apparently not.

Please use..

```js
if (cookie.name === 'sessionid') {
  sessionid = cookie.value;
}
if (cookie.name === 'csrftoken') {
  csrftoken = cookie.value;
}
```

(I'll try to fix prettier to deal with it as well.)

himdel (approved):
LGTM, I think this is ready :)

(you may want to rebase once #684 is merged)
(and don't forget to remove the WIP tag)

PR #668, opened 2021-07-20
[master-stable] Monkey patch automation prefix

ZitaNemeckova:
This needs to be fixed in nicer way.

### DIFF
--- a/src/loaders/insights/Routes.js
+++ b/src/loaders/insights/Routes.js
@@ -152,7 +152,7 @@ export const Routes = props => {
   return (
     <Switch>
       <InsightsRoute
-        path={Paths.repositories}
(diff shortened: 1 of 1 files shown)


src/loaders/insights/Routes.js
```diff
@@ -152,7 +152,7 @@ export const Routes = props => {
   return (
     <Switch>
       <InsightsRoute
-        path={Paths.repositories}
+        path={'automation-' + Paths.repositories}
```
himdel, line 155: You need to add new routes.

The whole UI still assumes /repositories exists, just the menu doesn't.

ZitaNemeckova:
Closing in favor of ...

PR #679, opened 2021-07-22
Test that token is generated and has correct form (length and characters)

MilanPospisil:
In Collections -> API Token, token is generated after click on Load Token. We check with regexp if the token meets requirements (thus it is not something else or some error).

### DIFF
--- a/.directory
+++ b/.directory
@@ -1,6 +1,0 @@
-[Dolphin]
-Timestamp=2021,6,28,12,29,21
-Version=4
-
-[Settings]
-HiddenFilesShown=true
--- a/.directory
+++ b/.directory
@@ -1,0 +1,6 @@
+[Dolphin]
+Timestamp=2021,6,28,12,29,21
+Version=4
+
+[Settings]
+HiddenFilesShown=true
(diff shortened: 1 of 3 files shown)


.directory
```diff
@@ -0,0 +1,6 @@
+[Dolphin]
```
himdel, line 1: You'll probably want to add `.directory` to you global gitignore (https://gist.github.com/subfuzion/db7f57fff2fb6998a16c)

test/cypress/integration/api_token.js
```diff
@@ -0,0 +1,17 @@
...
+    cy.login(adminUsername, adminPassword);
+  });
+
+  it('token is generated', () => {
+    cy.menuGo('Collections > API Token');
+    cy.contains('Load token').click();
+
+    cy.get('[aria-label="Copyable input"]')
+      .invoke('val')
+      .should('match', /[0-9a-f]{40}?/);
```
himdel, line 15: the trailing question mark looks like a typo?

(LGTM apart from those 2 issues)

MilanPospisil: Hmm...documentation says that after length there should be $, not ?. Probably mistake.

himdel (commented):
LGTM, except there's still the `.directory` file

himdel (approved):
LGTM :+1: 

Time to remove `[WIP]`? :)

PR #680, opened 2021-07-22
Test repo management that table is shown correctly.

MilanPospisil:
Test that it can see tables in local and remote tab, that the tables have correct columns and data. It depends on initial state of the basic data. This may change in future as well as this test.

### DIFF
--- a/test/cypress/integration/repo_management.js
+++ b/test/cypress/integration/repo_management.js
@@ -5,20 +5,47 @@ describe('Repo Management tests', () => {
     baseUrl,
     'ui/repositories?page_size=10&tab=remote',
   );
+  let localRepoUrl = urljoin(baseUrl, 'ui/repositories?page_size=10');
   let adminUsername = Cypress.env('username');
   let adminPassword = Cypress.env('password');
 
   beforeEach(() => {
-    cy.visit(baseUrl);
+    cy.login(adminUsername, adminPassword);
   });
 
   it('admin user sees download_concurrency in remote config', () => {
-    cy.login(adminUsername, adminPassword);
     cy.visit(remoteRepoUrl);
     cy.get('[aria-label="Actions"]:first').click(); // click the kebab menu on the 'community' repo
     cy.contains('Edit').click();
     cy.contains('Show advanced options').click();
     cy.get('#download_concurrency').should('exist');
+  });
+  
+  it('can see table in repo tabs and it shows data correctly', () => {
+    cy.visit(localRepoUrl);
+    cy.get('[aria-labelledby="headers"]').contains('Distribution name');
+    cy.get('[aria-labelledby="headers"]').contains('Repository name');
+    cy.get('[aria-labelledby="headers"]').contains('Content count');
+    cy.get('[aria-labelledby="headers"]').contains('Last updated');
+    cy.get('[aria-labelledby="headers"]').contains('Repo URL');
+    cy.get('[aria-labelledby="headers"]').contains('CLI configuration');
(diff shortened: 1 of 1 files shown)


test/cypress/integration/repo_management.js
```diff
@@ -5,22 +5,49 @@ describe('Repo Management tests', () => {
     baseUrl,
     'ui/repositories?page_size=10&tab=remote',
   );
+  let localRepoUrl = urljoin(baseUrl, 'ui/repositories?page_size=10');
```
himdel, line 8: suggestion: `baseUrl` is static, so this could just be...

```
describe('Repo Management tests', () => {
  let remoteRepoUrl = '/ui/repositories?page_size=10&tab=remote';
  let localRepoUrl = '/ui/repositories?page_size=10';
...
```

test/cypress/integration/repo_management.js
```diff
@@ -5,22 +5,49 @@ describe('Repo Management tests', () => {
...
+    cy.get('[aria-labelledby="headers"]').contains('CLI configuration');
+    cy.contains('community');
+    cy.contains('published');
+    cy.contains('rejected');
+    cy.contains('rh-certified');
+    cy.contains('staging');
+    cy.get('button')
+      .contains('Remote')
+      .parent()
+      .click({ force: true });
```
himdel, line 40: if you need `force: true` that means the Tab header is not visible

either this is not needed or you need to wait for something?

test/cypress/integration/repo_management.js
```diff
@@ -5,22 +5,49 @@ describe('Repo Management tests', () => {
...
+    cy.get('[aria-labelledby="headers"]').contains('Remote name');
+    cy.get('[aria-labelledby="headers"]').contains('Repositories');
+    cy.get('[aria-labelledby="headers"]').contains('Last updated');
+    cy.get('[aria-labelledby="headers"]').contains('Last sync');
+    cy.get('[aria-labelledby="headers"]').contains('Sync status');
+
+    cy.contains('Configure');
```
himdel, line 47: this is wrong

each remote line *may* have a Configure or a Sync button
but we can't assume they will always be one of each

test/cypress/integration/repo_management.js
```diff
@@ -5,22 +5,49 @@ describe('Repo Management tests', () => {
...
 
+  it('can see table in repo tabs and it shows data correctly', () => {
+    cy.visit(localRepoUrl);
+    cy.get('[aria-labelledby="headers"]').contains('Distribution name');
+    cy.get('[aria-labelledby="headers"]').contains('Repository name');
+    cy.get('[aria-labelledby="headers"]').contains('Content count');
+    cy.get('[aria-labelledby="headers"]').contains('Last updated');
+    cy.get('[aria-labelledby="headers"]').contains('Repo URL');
+    cy.get('[aria-labelledby="headers"]').contains('CLI configuration');
+    cy.contains('community');
```
himdel, line 32: this is testing a hardcoded list of repos coming from the db

Test that there are some lines, but not the specific values.

test/cypress/integration/repo_management.js
MilanPospisil: It is not needed indeed.

test/cypress/integration/repo_management.js
MilanPospisil: The same as above.

himdel (approved):
LGTM :+1: 

Time to remove `[WIP]`? :)

PR #682, opened 2021-07-22
Add npm run start-pulp command to run with vagrant backend.

newswangerd:
This starts webpack and with `/api/galaxy` as the api base path and sets it to proxy to port 8080 on localhost.


himdel (approved):
LGTM, not seeing any problems, standalone still works :)
(and `start-pulp` *does* talk to the changed port / path)

PR #669, opened 2021-07-20
Fix left nav even listener.

Hyperkid123:
The left navigation listener was using a deprecated attribute to navigate within the application. Now the event attribute is sending the exact link `href` to remove any ambiguity.

This change can work with both the old and the current navigation events. On Monday (July 26th) we will stop supporting the old `APP_NAVIGATION` event shape. After that this condition
```js
const to = event.domEvent.href ? event.domEvent.href.replace(this.props.basename, '') : event.navId
```
can be reduced to
```js
const to = event.domEvent.href.replace(this.props.basename, '')
```

PR #351, opened 2021-04-19
Prettier 2.3

himdel:
This just updates the version of prettier from 1.9 to 2.3 ([changelog](https://github.com/prettier/prettier/blob/main/CHANGELOG.md#232)),
and fixes the `test-prettier` task to run prettier (typo missing semicolon).

---

The last commit just updates the style everywhere.

Cc @newswangerd

PR #684, opened 2021-07-23
npm run prettier

himdel:
didn't happen in #669, seeing the change in too many PRs now, fixing :)

Cc @ZitaNemeckova 

(And adding backport-4.3 as the change was backported.)

### DIFF
--- a/src/entry-dev.js
+++ b/src/entry-dev.js
@@ -9,7 +9,7 @@ import getBaseName from './utilities/getBaseName';
 
 // Entrypoint for compiling the app to run in insights dev mode.
 
-const basename = getBaseName(window.location.pathname)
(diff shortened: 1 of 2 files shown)


newswangerd:
@himdel what happened to our CI prettier check? CI used to fail if the code wasn't styled correctly.

himdel:
Hmm... any chance that's the `test-prettier` task? If so, I the problem is probably https://github.com/ansible/ansible-hub-ui/pull/351/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R79

I'll take a look into reenabling it in a separate PR :)

newswangerd:
that's the one

PR #685, opened 2021-07-23
Docs menu test

jerabekjiri:
Issue: https://issues.redhat.com/browse/AAH-630

Test includes:
* open docs dropdown
* checking for links and their href values 
* toggling about modal

### DIFF
--- a/test/cypress/integration/docs_menu.js
+++ b/test/cypress/integration/docs_menu.js
@@ -1,0 +1,41 @@
+describe('Token Management Tests', () => {
+  const adminUsername = Cypress.env('username');
+  const adminPassword = Cypress.env('password');
+
+  before(() => {
+    cy.deleteTestUsers();
+    cy.galaxykit('user create', adminUsername, adminPassword);
+  });
+
+  beforeEach(() => {
+    cy.visit('/');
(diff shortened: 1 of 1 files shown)


test/cypress/integration/docs_menu.js
```diff
@@ -0,0 +1,41 @@
+describe('Token Management Tests', () => {
+  const adminUsername = Cypress.env('username');
+  const adminPassword = Cypress.env('password');
+
+  before(() => {
+    cy.deleteTestUsers();
```
himdel, line 6: this looks good in principle, but in the specific case of admin, it doesn't actually do anything

`galaxykit -u admin -p admin  user delete admin` pretends to work, returns a 0 status, but the actual `DELETE /api/automation-hub/_ui/v1/users/1/` request fails when trying to delete the user that's doing the request.

And then

```
$ galaxykit -u admin -p admin  user create admin admin
User admin already existed
```

:)

himdel (approved):
LGTM :+1: , you may want to drop the `before`, but the test looks good :)

himdel:
LGTM, the pr-checks issue was fixed in #711 so rebasing should fix it. (Closing and reopening might also fix it, whichever is easier :).)

PR #703, opened 2021-07-26
[master-stable] Collection-info: replace Download tarball button with a warning banner when on cloud.redhat.com

himdel:
The API is already redirecting to console.redhat.com, so when using the UI on cloud, the download links lead to console, which fails authentication.

Hiding the download button when the domain matches cloud.redhat.com

Before:
before

After (only on cloud.redhat.com, not locally, not console.redhat.com):
after

### DIFF
--- a/src/components/collection-detail/collection-info.tsx
+++ b/src/components/collection-detail/collection-info.tsx
@@ -5,6 +5,7 @@ import * as moment from 'moment';
 import { Link } from 'react-router-dom';
 
 import {
+  Banner,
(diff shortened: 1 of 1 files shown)


himdel:
Merged in master-stable, prod-beta and prod-stable as of now :)
(not needed in master, this can be removed once cloud.redhat.com redirects the UI to console.redhat.com, presumably this  thursday)

himdel:
Cc @ZitaNemeckova FYI :)

PR #705, opened 2021-07-27
Fix nav listener basename replacement.

Hyperkid123:
Application in beta env has additional `/beta/` prefix in the router
basename. However the navigation event href attribute does not include
that prefix. Therefore the replacement fails and an incorrect pathname is
pushed to the router history. Which leads to permanent 404s.

Also, the `domEvent` is sometimes `undefined`. We don't want to perform any action when that is the case.


ZitaNemeckova (approved):
LGTM :+1:

Hyperkid123:
@ZitaNemeckova thanks! Please make sure this gets deployed to every env 🙂

ZitaNemeckova:
@Hyperkid123 Thank you! I'll make sure it gets to other env. as well.

PR #677, opened 2021-07-21
[l10n] Auto-mark strings for translation

himdel:
Fixes [AAH-800](https://issues.redhat.com/browse/AAH-800)

This PR:
* makes it possible to start marking strings for translation
  * this does not deal with plurals (`n_`) or non-translating (`N_`) yet, just the basic version
* auto-marks as many strings for translation as possible to do automatically (and fixes any errors generated during)

To achieve the first, this adds a simple `_` JS string template that either behaves as regular template string, or, when `localStorage.test_l10n = true` is set (+reload), marks every translatable string in french quotes...

```
_`Hello ${ 'world' }` #=> 'Hello world'
localStorage.test_l10n = true
_`Hello ${ 'world' }` #=> '>>Hello <<world'
```

l10n

The latter was just a series of perl invocations to convert anything that looks like a human string,
only strings starting with a capital letter could safely be converted in some cases,
multiline strings and strings with params are not quite possible to do with regex without some major preprocessing.

        perl -i -npe 's/(?<!_)`[A-Z][^`]*`/_$&/' src/**/*.*
        perl -i -npe s/=\''([A-Z][-a-zA-Z0-9_,. ]*)'\'/'={_`$1`}'/g src/**/*.*
        perl -i -npe s/\''([A-Z][-a-zA-Z0-9_,. ]+)'\''(?!:)/_`$1`'/g src/**/*.*
        perl -i -npe s/\''([A-Z][-a-zA-Z0-9_,.:?!@$%\[\]<> ]+)'\''(?!:)/_`$1`'/g src/**/*.*
        perl -i -npe s/=\''([A-Z][-a-zA-Z0-9_,.:?!@$%\[\]<> ]+)'\''/={_`$1`}'/g src/**/*.*
        perl -i -npe 's/"([A-Z][-a-zA-Z0-9_,.:?!@$%\[\]<> '\'']+)"(?!:)/_`$1`/g' src/**/*.*
        perl -i -npe 's/="([A-Z][-a-zA-Z0-9_,.:?!@$%\[\]<> '\'']+)"/={_`$1`}/g' src/**/*.*
        perl -i -npe 's/>([A-Z][-a-zA-Z0-9_,.:?!@$%\[\] '\'']+)</>{_`$1`}</g' src/**/*.*
[...]

PR #709, opened 2021-07-27
add state filter to imports page

ShaiahWren:
See issue: https://issues.redhat.com/browse/AAH-787
Implements combined filter for name and state.

Before:
Screen Shot 2021-07-27 at 11 25 50 AM
After:
Screen Shot 2021-07-27 at 11 23 47 AM


src/components/my-imports/import-list.tsx
```diff
@@ -65,24 +71,76 @@ export class ImportList extends React.Component<IProps, IState> {
...
-
-          <Button
-            variant={ButtonVariant.control}
-            aria-label='search button'
-            onClick={() => this.submitSearch()}
-          >
-            <SearchIcon />
-          </Button>
-        </InputGroup>
+        {/* <InputGroup className='search-box'>
```
himdel, line 74: no commented out code please :)

src/entry-dev.js
```diff
@@ -9,7 +9,7 @@ import getBaseName from './utilities/getBaseName';
 
 // Entrypoint for compiling the app to run in insights dev mode.
 
-const basename = getBaseName(window.location.pathname)
+const basename = getBaseName(window.location.pathname);
```
himdel, line 12: you'll need to rebase, this was fixed in #684

src/components/my-imports/import-list.tsx
```diff
@@ -65,24 +71,76 @@ export class ImportList extends React.Component<IProps, IState> {
...
+          <ToolbarContent>
+            <ToolbarGroup>
+              <ToolbarItem>
+                <CompoundFilter
+                  updateParams={this.props.updateParams}
+                  params={params}
+                  filterConfig={[
+                    {
+                      id: 'keywords',
+                      title: 'Name',
```
himdel, line 103: #677 is now merged, so all the human strings should be marked as `` _`Name` `` (not the ids, just the titles and niceNames values)

ShaiahWren:
@himdel  lmk if this is looking better. :)

himdel:
Cc @sbuenafe-rh

I think the AplliedFilter is missing some padding, it should not be touching the list of items.

(I don't know what design you've settled on, so I can't comment more, maybe the filter should move to the header, where there's space for it.)

himdel:
@ShaiahWren Adding or removing filters doesn't reset the page.

Apart from that, both code and functionality :+1: (Made a separate comment about the design above :))

sbuenafe-rh:
@himdel from the edge of the card to the filter, the margin should be 24px (aligned with the chip groups).

himdel:
Ok, so.. @ShaiahWren ...

127182144-3c1e67c0-15ca-4afc-907c-6d19d56cda6a

the red bits are padding that should not be there,
the green bits are padding that should be added

(And should we also adjust the width, so that CompundFilter takes up the whole width of the original search?)

ShaiahWren:
@himdel thanks for the review! :) I am just not sure how to reset the page to 1 when adding and removing filters.. can you enlighten me

himdel:
ooh, sure :) 

The change itself is just

```diff
-updateParams={this.props.updateParams}
+updateParams={p => { p['page'] = 1; this.props.updateParams(p); }}
```

and you'd have to apply it only in the places where it *should* be reset .. so `CompoundFilter` for when a filter is added, `AppliedFilter` when removed; `Pagination` already updates the page ; .. possibly also the `FormSelect` as changing the namespace should also reset page to 1.

Oh, and I notice both `submitSearch` and `handleEnter` are now dead code, you can remove them :) (used to be called by the code removed [here](https://github.com/ansible/ansible-hub-ui/pull/709/files#diff-c3fdde670bdf4d238ac93ff344f0f449e6726f2ade0c311302d90a830221fad8L72-L81))

src/components/my-imports/import-list.tsx
```diff
@@ -210,4 +243,8 @@ export class ImportList extends React.Component<IProps, IState> {
       </div>
     );
   }
+
+  private get updateParams() {
```
himdel, line 247: the code is using `this.props.updateParams` (or `updateParams` after deconstructing it from `this.props`), but I'm not seeing any `this.updateParams` calls

so this method is not used, no need to add it?

ShaiahWren:
@himdel I _thiink_ this is an improvment, I wasn't 100% sure what to do on the FormSelect :)

himdel:
> It also looks like the linter is not passing

> src/components/my-imports/my-imports.scss
>  56:1  ✖  Unexpected duplicate selector ".import-list", first used at line 46   no-duplicate-selectors

this is related to your changes, you should not have `.import-list`  twice in the css :)

himdel:
> I wasn't 100% sure what to do on the FormSelect :)

It looks about right, if going to page 2, and then changing the namespace resets the page to 1, it's a success :) (it looks ok :))

src/components/my-imports/my-imports.scss
```diff
@@ -39,7 +39,18 @@
   }
 }
 
+.pf-c-toolbar__content-section {
+  margin: 0;
+}
+
+.import-list > div {
```
himdel, line 46: `& > div` inside the `.import-list` on line 42/50

src/components/my-imports/my-imports.scss
```diff
@@ -39,7 +39,18 @@
   }
 }
 
+.pf-c-toolbar__content-section {
```
himdel, line 42: this one should go inside .import-list as well, if possible, otherwise it affects all toolbars

ShaiahWren:
@himdel I think this is looking better, lmk what you think :)

himdel:
Updated screenshot (with and without filters):

Untitled

Looks great! :+1: 

---

The only remaining problem I can see is that changing a namespace in the select now doesn't do anything, :eyes:

src/components/my-imports/import-list.tsx
```diff
@@ -194,11 +220,12 @@ export class ImportList extends React.Component<IProps, IState> {
         <div className='label'>{_`Namespace`}</div>
         <div className='selector'>
           <FormSelect
-            onChange={(val) =>
+            onChange={(val) => {
+              val['page'] = 1;
```
himdel, line 224: ok, so, all the other places pass the changed state with page as the first param to `updateParams`,
here .. `val` is only the value to use for the `namespace` field in params, it doesn't have a `['page']` as it's not an object.

So this needs to be something like...

```jsx
<FormSelect
  onChange={(val) => this.props.updateParams({ ...ParamHelper.setParam(this.props.params, 'namespace', val), page: 1 })}
```
this.props.updateParams(
￼                ParamHelper.setParam(this.props.params, 'namespace', val),

himdel:
Oh, and one more issue (sorry :)).. the EmptyState is no longer the right one given this change:

Untitled
"There have not been any imports on this namespace."

That message is fine if there are no filters, but we should be using EmptyStateFilter when there is an active filter.

---

And related, Clear All Filters should not remove `namespace`, unless we move the Namespace select inside the CompoundFilter. Otherwise, clearing all filters gives you imports from all namespaces, yet the namespace select is still set.

ShaiahWren:
@himdel I think this is better, only trouble is that I get the wrong NoData message when with an incorrect name__contains filter. 

Also, the PR checks failed... not sure how to sleuth that error :)

himdel:
@ShaiahWren the same as the other PRs, you need to run rebase on current master, npm install and *then* run prettier. Now it's changing all the files back.

ShaiahWren:
@himdel Any ideas for getting the correct NoData filter when only an incorrect name__contains filter is applied? The others seem to work fine :)

himdel:
Sure :) The filter is never named `name__contains` in imports, it's `keywords` :)

```diff
diff --git a/src/components/my-imports/import-list.tsx b/src/components/my-imports/import-list.tsx
index fb29669..c2c80c5 100644
--- a/src/components/my-imports/import-list.tsx
+++ b/src/components/my-imports/import-list.tsx
@@ -146,7 +146,7 @@ export class ImportList extends React.Component<IProps, IState> {
     }
     if (
       importList.length === 0 &&
-      !filterIsSet(this.props.params, ['name__contains', 'state'])
+      !filterIsSet(this.props.params, ['keywords', 'state'])
     ) {
       return (
         <EmptyStateNoData
```

ShaiahWren:
Thank you, Martin :)
Hopefully it's looking better now

src/components/my-imports/import-list.tsx
```diff
@@ -37,6 +38,7 @@ interface IProps {
 
 interface IState {
   kwField: string;
+  status: string;
```
himdel, line 41: `status` is unused

src/components/my-imports/import-list.tsx
```diff
@@ -45,6 +47,7 @@ export class ImportList extends React.Component<IProps, IState> {
 
     this.state = {
       kwField: '',
+      status: '',
```
himdel, line 50: `status` is unused

src/components/my-imports/import-list.tsx
```diff
@@ -2,21 +2,22 @@ import * as React from 'react';
 import cx from 'classnames';
 import './my-imports.scss';
 import {
-  TextInput,
   Pagination,
   FormSelect,
   FormSelectOption,
   InputGroup,
```
himdel, line 8: `InputGroup` is unused

src/components/my-imports/import-list.tsx
```diff
@@ -2,21 +2,22 @@ import * as React from 'react';
...
   Button,
   ButtonVariant,
```
himdel, line 10: `Button` & `ButtonVariant` is unused

src/components/my-imports/import-list.tsx
```diff
@@ -2,21 +2,22 @@ import * as React from 'react';
...
   Spinner,
+  Toolbar,
 } from '@patternfly/react-core';
 import { SearchIcon } from '@patternfly/react-icons';
```
himdel, line 14: `SearchIcon` is unused

himdel (approved):
LGTM, tested in the UI, not seeing any more issues :+1: :)

(Please do remove the 6 unused lines I flagged, but this is ready.)

ShaiahWren:
@himdel updates complete, want to look at it once more or shall I merge?

himdel:
Feel free to merge :+1: :)

PR #702, opened 2021-07-26
Github actions: add linter checks, merge commit check

himdel:
Follow up to #684 

`npm run test-prettier`
previously only did echo, missing semicolon so it didn't run prettier
changed to run a prettier:check task and moved the prettier-but-dont-change-stuff logic there

`npm run lint`
previously tried to run eslint, failed
now runs prettier for js & stylelint for sass

Cypress workflow - move checks for "npm install didn't have to modify package-lock.json" to the new pr-checks workflow.

pr-checks workflow: run linters, npm install check, and check for new merge commits in the PR

---

Also fixed 2 stylelint issues to get green:

    src/components/card-list-switcher/switcher.scss
     5:1  ￼  Expected selector ".icon" to come before selector ".icon:hover"   no-descending-specificity

=> moved `&:hover` style inside `.icon`

    src/containers/search/search.scss
     59:3  ￼  Unexpected duplicate selector ".search-page .pf-c-toolbar__content", first used at line 16   no-duplicate-selectors

=> removed duplicate css, already exists on line 16

PR #711, opened 2021-07-28
Fix prettier failing

ZitaNemeckova:
https://github.com/ansible/ansible-hub-ui/pull/702 introduced prettier check and https://github.com/ansible/ansible-hub-ui/pull/635 introduced a linting error.

This fixes it.


himdel (approved):
LGTM, fixes the issue

(though we may want to eventually fix prettier config to not split single line commands)

PR #716, opened 2021-07-29
[stable-4.3] [l10n] Add minimal support for `` _`Localized strings` ``

himdel:
master is now (in the process of being) localized, many strings changed from `'Foo'` to `` _`Foo` ``, backporting any such changes to 4.3 or 4.2 would break on unknown function `_`.

This backports the bare minimum of #677, adding a global template `_` function, but replacing the custom template with the equivalent of plain template strings.

This allows backporting marked strings, without enabling l10n support.


himdel:
Cc @ZitaNemeckova

ZitaNemeckova (approved):
Agreed!

PR #738, opened 2021-08-02
Profile/Edit Profile Tests

jerabekjiri:
Issue: [AAH-619](https://issues.redhat.com/browse/AAH-619)

- _a superuser cannot change its superuser rights:_ check if switch input is disabled
- _an user cannot set superusers rights:_ login as test user and check if switch is disabled
- _email must be email:_  testing mail with regex
- _groups input is readonly:_  check for groups input doesn't exist
- _User can save form:_  intercepting the PUT request and waiting for a successful status response and user notification
- _Cancel:_  change fields in form, cancel the form, and verify that data didn't change

### DIFF
--- a/src/components/user-form/user-form.tsx
+++ b/src/components/user-form/user-form.tsx
@@ -120,7 +120,12 @@ export class UserForm extends React.Component<IProps, IState> {
     );
 
     const readonlyGroups = () => (
-      <FormGroup fieldId='groups' key='readonlyGroups' label={_`Groups`}>
+      <FormGroup
+        fieldId='groups'
+        key='readonlyGroups'
+        label={_`Groups test`}
+        aria-labelledby='readonly-groups'
+      >
         {user.groups.map((group) => (
           <Label key={group.name}>{group.name}</Label>
         ))}
--- a/src/components/user-form/user-form.tsx
+++ b/src/components/user-form/user-form.tsx
@@ -123,7 +123,7 @@ export class UserForm extends React.Component<IProps, IState> {
       <FormGroup
         fieldId='groups'
         key='readonlyGroups'
-        label={_`Groups test`}
+        label={_`Groups`}
         aria-labelledby='readonly-groups'
       >
         {user.groups.map((group) => (
--- a/test/cypress/integration/profile.js
+++ b/test/cypress/integration/profile.js
@@ -1,6 +1,14 @@
 describe('My Profile Tests', () => {
-  var adminUsername = Cypress.env('username');
-  var adminPassword = Cypress.env('password');
+  const adminUsername = Cypress.env('username');
+  const adminPassword = Cypress.env('password');
+
+  const username = 'nopermission';
+  const password = 'n0permissi0n';
+
+  before(() => {
+    cy.deleteTestUsers();
+    cy.galaxykit('user create', username, password);
+  });
 
   beforeEach(() => {
     cy.login(adminUsername, adminPassword);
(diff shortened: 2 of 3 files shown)


test/cypress/integration/profile.js
```diff
@@ -31,4 +37,60 @@ describe('My Profile Tests', () => {
...
+    cy.get('[aria-label="user-dropdown"] button').click();
+    cy.get('a').contains('My profile').click();
+
+    cy.get('.pf-c-switch__input').should('be.disabled');
+  });
+
+  it('email must be email', () => {
+    cy.get('#email')
+      .invoke('val')
+      .should('match', /^[a-zA-Z0-9._-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,6}$/i);
```
himdel, line 59: this looks too strict .. and looks like it's only testing that test@example.com matches the regex?

What's the purpose here? 
Should this be checking that the field fails validation on invalid emails?

In general, checking emails for regex match is Considered Harmful, in favor of simply trying to send an email and requiring the user to click a link in there.
Because all of these are valid email addresses as well:
`foo+bar@example.com`, `"Hello I have spaces"@example.com`, `@1st.relay,@2nd.relay:user@final.domain`, `'*+-/=?^_{|}~#$@netmeister.org`, `relay.domain!user@domain`, `"🌮"@i❤️tacos.ws`

(more info in https://davidcel.is/posts/stop-validating-email-addresses-with-regex/ and https://www.netmeister.org/blog/email.html)

test/cypress/integration/profile.js
```diff
@@ -1,10 +1,17 @@
...
+  const password = 'n0permissi0n';
+
+  before(() => {
+    cy.deleteTestUsers();
+    cy.galaxykit('user create', username, password);
+  });
 
   beforeEach(() => {
-    cy.visit(baseUrl);
+    cy.visit('/');
```
himdel, line 14: (no need for `visit` right before `login`)

himdel:
(don't worry about the `group_permissions` test failures, fixed in #751)

.github/workflows/cypress.yml
```diff
@@ -124,7 +124,7 @@ jobs:
         npm run build-standalone
         rm -rf ../pulp_galaxy_ng/pulp_storage/assets/galaxy_ng/
         mv -v dist/ ../pulp_galaxy_ng/pulp_storage/assets/galaxy_ng
-        podman exec pulp bash -c "s6-svc -r /var/run/s6/services/pulpcore-api" # apply changes, api also serves static assets
+        podman exec pulp bash -c "sed -i 's/collectstatic --clear/collectstatic/' /var/run/s6/services/pulpcore-api/run"
```
himdel, line 127: #788 fixes this issue now :)

src/components/user-form/user-form.tsx
```diff
@@ -120,7 +120,12 @@ export class UserForm extends React.Component<IProps, IState> {
     );
 
     const readonlyGroups = () => (
-      <FormGroup fieldId='groups' key='readonlyGroups' label={_`Groups`}>
+      <FormGroup
+        fieldId='groups'
+        key='readonlyGroups'
+        label={_`Groups test`}
```
himdel, line 126: Groups test -> Groups ?

jerabekjiri: thanks, didn't see that one, should be fixed now :)

himdel:
Looks good, restarting cypress, we may still have an occasional group_management failure, likely not related to this PR, making sure :)

himdel (approved):

PR #742, opened 2021-08-03
Update to webpack 5, frontend-components-config 4, patternfly current

himdel:
The point is to update to webpack 5, but that requires updating `@redhat/cloud-services-config` in lockstep (or hardcoding it out, but it keeps changing :)),
and these mean updates to webpack loaders, and sass, which fails to compile older patternfly, but also allows updating to newer one, doing so.

A secondary goal is to set everything up for #810 string extraction, which means compiling all code with babel, not tsc, to support `@lingui/macro` ([docs](https://github.com/lingui/js-lingui/blob/main/docs/ref/macro.rst)), while preserving typescript type checking.

---

Changes:

* all js/jsx/ts/tsx files go through babel for compilation now
* a new `lint:ts` task ensures `npm run lint` fails on typescript errors
* updated `.babelrc` to remove modules included in current `@babel/preset-env`, and add `@babel/preset-typescript`
* updated `tsconfig.json` to prevent errors in incompatible `node_modules/@types/webpack`
* css extraction now runs in build mode (slower build but smaller output), remaining in js for dev mode (faster build); together with the typescript change, dev build goes down from ~60s to ~15s
* and actual webpack config updates for webpack 5 :)
* and 4 related code updates, 3 for load changed import paths, one for the `_` global not being available at module load time, fixed better in #810


ZitaNemeckova:
I believe there's no type check. I can change and it compiles with no complains
```diff
diff --git a/src/containers/search/search.tsx b/src/containers/search/search.tsx
index 5be95e6..0b72f34 100644
--- a/src/containers/search/search.tsx
+++ b/src/containers/search/search.tsx
@@ -32,8 +32,8 @@ interface IState {
   numberOfResults: number;
   params: {
     page?: number;
-    page_size?: number;
-    keywords?: string;
+    page_size?: string;
+    keywords?: string[];
     tags?: string[];
     view_type?: string;
   };
```

```
ui_1              | ℹ ｢wdm｣: Compiling...
ui_1              | ℹ ｢wdm｣: assets by status 6 MiB [cached] 14 assets
ui_1              | assets by path . 1.02 MiB
ui_1              |   assets by path js/*.js 1.02 MiB
ui_1              |     asset js/App.a0be0e9c903dae413068.js 1020 KiB [emitted] [immutable] (name: App) 1 related asset
ui_1              |     asset js/search.b89daa31412dd49a6c29.js 29.6 KiB [emitted] [immutable] (name: search) 1 related asset
ui_1              |   asset index.html 709 bytes [emitted]
ui_1              |   asset fed-mods.json 2 bytes [emitted]
ui_1              | Entrypoint App 6.71 MiB (6.29 MiB) = js/pfVendor.84f24568e853584538e9.js 2.77 MiB js/reactVendor.7242466850fec7a485e1.js 901 KiB js/rhcsVendor.c9b96979cec34e385a56.js 6.84 KiB js/vendor.1f254a1c2fab69728f48.js 2.07 MiB js/App.a0be0e9c903dae413068.js 1020 KiB 7 auxiliary assets
ui_1              | cached modules 8.15 MiB (javascript) 1.54 MiB (asset) [cached] 3321 modules
ui_1              | runtime modules 7.7 KiB 14 modules
ui_1              | ./src/containers/search/search.tsx 12.4 KiB [built] [code generated]
ui_1              | webpack 5.49.0 compiled successfully in 1583 ms
ui_1              | ℹ ｢wdm｣: Compiled successfully.
```

@himdel Did you want to solve it in this PR or in follow up?

Otherwise it works in standalone and insights.

newswangerd (approved):

himdel:
Added `fork-ts-checker` webpack plugin which runs the typechecking in parallel with compilation.
In dev, the ts errors are still not fatal, but should be visible again, both in webpack output and browser console :).

ZitaNemeckova (approved):
LGTM :+1:

PR #670, opened 2021-07-20
apply partial filter ui

ShaiahWren:
Issue: https://issues.redhat.com/browse/AAH-355

Before:
Screen Shot 2021-07-15 at 11 52 01 AM
After:
Screen Shot 2021-07-15 at 11 51 06 AM

Note: __icontains doesn't work for me, only __contains

PR #743, opened 2021-08-03
Fix sorting problem in Users list view

ZitaNemeckova:
Fixes:
 - default filter is set but not shown in UI
- clicking on sorting arrows will start never ending spin

Introduced by https://github.com/ansible/ansible-hub-ui/pull/670 -> no backporting needed

Before:
Screenshot 2021-08-03 at 12 08 42
Screenshot 2021-08-03 at 12 08 55

After:
Screenshot 2021-08-03 at 12 07 59


himdel (approved):

ShaiahWren (approved):
Works for me!

PR #744, opened 2021-08-03
Do not allow filter to search for empty string

ZitaNemeckova:
Fixes https://issues.redhat.com/browse/AAH-820

Before:
Screenshot 2021-08-03 at 12 48 52

After:
Screenshot 2021-08-03 at 12 44 08

@sbuenafe-rh WDYT? Thanks :)


ShaiahWren (approved):
LGTM :)

sbuenafe-rh:
@ZitaNemeckova I think the chip category would just need to reflect the same attribute name (Task name) and then everything else looks good!

PR #746, opened 2021-08-04
[WIP] Refactor logins  to cookieLogin of many tests

MilanPospisil:
Remaining tests up to date were refactored, so they now use cookieLogin for faster tests. Only one file was not refactored - group_management.js which failed every time. This will be work rather for new PR.

### DIFF
--- a/test/cypress/integration/docs_menu.js
+++ b/test/cypress/integration/docs_menu.js
@@ -4,7 +4,7 @@ describe('Token Management Tests', () => {
 
   beforeEach(() => {
     cy.visit('/');
-    cy.login(adminUsername, adminPassword);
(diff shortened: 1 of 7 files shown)


himdel:
Please remove every single line containing `baseUrl` (fixed in ...),
and call `cy.cookieReset()` from `before()` in all tests where `cookieLogin` is used :).

CSRF tokens are single use (single POST), so we can't really rely on being able to log in from the cache after they were used :(.
(It's ok within the test file, the authors can cope then :).)

Or (and given https://github.com/ansible/ansible-hub-ui/pull/747#issuecomment-893937569), we can just skip this PR completely and revisit once api login is in.

PR #747, opened 2021-08-04
More login tests and correction of cookieLogin command

MilanPospisil:
Bug found in cookieLogin. When user is deleted and created again, cookieLogin still has its cookies in local storage and tries to login using cookies that are not valid now. New functionality adds request to the server that returns ok if user is looged and error if its not. If error is received, local storage for that user is deleted and manual login using UI is done. Then, local storage is updated to new cookies.

### DIFF
--- a/test/cypress/integration/login.js
+++ b/test/cypress/integration/login.js
@@ -9,7 +9,7 @@ describe('Login helpers', () => {
     cy.galaxykit('user create', username, password);
   });
 
-  /*it('can login manually and logout as admin or different user', () => {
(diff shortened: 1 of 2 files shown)


himdel:
Aplogies, we won't go this way.

`cookieLogin` is limited by the csrf token being single (POST) use and relying on it being refreshed in subsequent requests.

Given that, and given you mentioned having solved API login,
it will be easier to introduce api login, switch to that, and only use cookie login within testfiles where it's definitely safe to do.

Does that sound reasonable?

himdel:
Closing in favor of #768

PR #748, opened 2021-08-05
Rename Name to Task name to be same as filter

ZitaNemeckova:
Solves https://github.com/ansible/ansible-hub-ui/pull/744#issuecomment-893473513

>@ZitaNemeckova I think the chip category would just need to reflect the same attribute name (Task name) and then everything else looks good!

Before:
Screenshot 2021-08-05 at 16 33 13

After:
Screenshot 2021-08-05 at 16 16 14


ShaiahWren (commented):
This change LGTM. 👍  

It does seems like there is excess veritcal space between compound filter and applied filter chips, probably because of the pagination bar. Perhaps padding-bottom on the compound filter can be removed?

Also, I think the applied filter chips should be aligned left with the compound filter.

PR #681, opened 2021-07-22
Test group permissions - if it can (or cant) see groups, add groups, delete or change groups.

MilanPospisil:
Test if users with certain group privilleges can see groups, add new groups or change groups. We tests user with no privilleges, user with no group at all, user with view only privilleges and user with all privilleges.

PR #751, opened 2021-08-06
Fix cypress failures on master

himdel:
Master cypress started failing in https://github.com/ansible/ansible-hub-ui/runs/3254310124?check_suite_focus=true,
after merging #681, and I found a couple of more issues locally when testing, and fixed a few more.

The main problem seems to be that `cookieLogin` won't work after the csrf token was used to do a POST, leading to interactions between tests ... adding a separate `cookieReset` call to call in `before` of any tests using it, until this is resolved.

The second problem was that our user/group helpers still don't wait for the last action, aded waits to `createGroup`, `removeGroup`, `addPermissions`, `removePermissions` (and fix tabs in comments).

Also adding a dependabot rule for `test/` so that dependabot keeps cypress up to date.

And adding `Task Management` to the menu test.

Also found 2 places where one `it` depends on data created in a previous `it` - merging the `it`s in `user_filter` test, creating a sepate group for deletion in `group_permissions`.

Other cleanups:

* remove every baseUrl from tests, remove any cy.visit before cy.login, login already visits the login page
* move login to beforeEach where possible
* add missing cleanup when creating users and not deleting them
* remove external links other than redhat, ansible & example.com
* remove a selector matching a pathname inside an SVG icon
* remove a test for the cookieLogout helper (see also https://github.com/ansible/ansible-hub-ui/pull/747#issuecomment-893937569)
* remove any `localhost:8002` hardcoded in tests
* remove `page_size` hardcoded in urls

:crossed_fingers:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,6 +6,11 @@ updates:
       interval: 'weekly'
 
   - package-ecosystem: 'npm'
+    directory: '/test'
+    schedule:
+      interval: 'weekly'
+
+  - package-ecosystem: 'npm'
     directory: '/'
     target-branch: 'stable-4.3'
     schedule:
--- a/test/cypress/integration/group_management.js
+++ b/test/cypress/integration/group_management.js
@@ -5,13 +5,16 @@ describe('Hub Group Management Tests', () => {
   beforeEach(() => {
     cy.deleteTestGroups();
     cy.deleteTestUsers();
+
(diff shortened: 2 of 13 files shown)


himdel:
(It went green, but I want to be doubly sure :))

ZitaNemeckova (approved):
LGTM 👍  Merging to make master 🟢 again :)

PR #785, opened 2021-08-09
remove timestamp for null values

ShaiahWren:
Hello :) 
See issue https://issues.redhat.com/browse/AAH-719, Task Management List View, which was sent back for better handling of null timestamp values for unfinished tasks. It was suggested that I either leave null values empty or indicate 'N/A'. I left it empty. It was just personal preference for making the interface cleaner. Other suggestions welcome. :)

Before:
AAH-719-invalid-date
After:
Screen Shot 2021-07-16 at 10 47 40 AM


src/containers/task-management/task-list-view.tsx
```diff
@@ -236,9 +236,13 @@ export class TaskListView extends React.Component<RouteComponentProps, IState> {
...
           <DateComponent date={started_at} />
         </td>
-        <td>
-          <DateComponent date={finished_at} />
-        </td>
+        {finished_at === null ? (
+          <td></td>
+        ) : (
+          <td>
+            <DateComponent date={finished_at} />
```
himdel, line 243: what about just

```
<td>
  {finished_at && (<DateComponent date={finished_at} />)}
</td>
```

(no need to duplicate the `td`)

src/containers/task-management/task-list-view.tsx
```diff
@@ -236,9 +236,13 @@ export class TaskListView extends React.Component<RouteComponentProps, IState> {
         <td>
           <DateComponent date={started_at} />
```
himdel, line 237: I think a task can exist without having started yet, you may want to add the same logic here

newswangerd (commented):
I would just modify `DateComponent` to handle null dates correctly, so that this problem can be solved everywhere.

himdel:
(removing review request for Zita, the review happened yesterday and it still spams slack)

ShaiahWren:
Hopefully, this will do the trick :)

ShaiahWren:
@himdel Really sorry about the fail :( I rebased and ran prettier, cypress still isn't happy

himdel:
@ShaiahWren No worries, the cypress failure should be fixed now, master is running green for me locally, it's just that Cypress is testing an older version of the UI now, fixed in #788.

I'd say this is ready :+1:

ShaiahWren:
Merging this pr even though cypress doesn't pass. Hopefully #788 fixes the cypress failure.

PR #790, opened 2021-08-10
[l10n] String translation suggestion 

jerabekjiri:
Issue: [AAH-801](https://issues.redhat.com/browse/AAH-801
)

Marked a few strings for translation and found a few possible bugs, not sure if they are resolved or discussed already.

A small thing, but we could translate the tooltip messages of ClipboardCopy component. [tooltip messages](https://github.com/jerabekjiri/ansible-hub-ui/blob/ea7b12adee0673a0d5f1c8b0e39c1eb4086e888a/src/components/collection-detail/collection-info.tsx#L107-L111)
Screenshot from 2021-08-10 16-38-19
I think the easiest way would be to have a reusable component (wrapper) for the ClipboardCopy component.

Should this be marked for translation (namespace, collections, users, groups, remotes, containers)?
Screenshot from 2021-08-10 16-08-28

Translate status (completed, failed, running, waiting) ?
translate_completed

I found a bug with translation in tabs, not sure if this was mentioned already. Switching tabs doesn't work with translations. We probably shouldn't put translation to URI. This occurred in namespace detail, repo management as well.
translation_broken

I also marked a strings that I think might be fit for the translation :)


src/components/cards/collection-card.tsx
```diff
@@ -63,7 +63,10 @@ export class CollectionCard extends React.Component<IProps> {
           </div>
           <div className='author'>
             <TextContent>
-              <Text component={TextVariants.small}>Provided by {company}</Text>
+              <Text component={TextVariants.small}>
+                {_`Provided by `}
+                {company}
```
himdel, line 68: `company `is just a plain string, right?

If so, this should be `` {_`Provided by ${company}`} ``, so that the translators can translate it to something like `${company} provided this` if necessary.

(same in collection-list-item)

src/components/collection-list/collection-list-item.tsx
```diff
@@ -111,7 +114,8 @@ export class CollectionListItem extends React.Component<IProps, {}> {
       <DataListCell isFilled={false} alignRight key='stats'>
         {controls ? <div className='entry'>{controls}</div> : null}
         <div className='right-col entry'>
-          Updated <DateComponent date={latest_version.created_at} />
+          {_`Updated `}
```
himdel, line 117: (works for now, these should eventually also do something like `Updated ${on}` ... except right now it doesn't work with react components, so, works for now :))

src/loaders/standalone/standalone-loader.tsx
```diff
@@ -102,7 +102,8 @@ class App extends React.Component<RouteComponentProps, IState> {
 
       userDropdownItems = [
         <DropdownItem isDisabled key='username'>
-          Username: {user.username}
+          {_`Username: `}
```
himdel, line 105: not sure if it's ever needed here, but technically this should also be `` {_`Username: ${user.username}`} ``

(better safe than sorry in these cases, if we can do that :))

src/containers/token/token-standalone.tsx
```diff
@@ -22,25 +22,28 @@ class TokenPage extends React.Component<RouteComponentProps, IState> {
   render() {
     const { token } = this.state;
 
+    const code = `\<code\>ansible-galaxy\<\/code\> `;
```
himdel, line 25: yeah, custom html inside the strings also won't work (yet)

himdel:
Thanks, looks good, except for the comments above :)

> A small thing, but we could translate the tooltip messages of ClipboardCopy component. tooltip messages
> I think the easiest way would be to have a reusable component (wrapper) for the ClipboardCopy component.

:+1:  on both :)

> Should this be marked for translation (namespace, collections, users, groups, remotes, containers)?

Yes, the entity types should be translated, we should probably also capitalize them in the process :+1: 

> Translate status (completed, failed, running, waiting) ?

Yes, as long as we know the possible values for the enum, we can have a constant to translate.

> I found a bug with translation in tabs, not sure if this was mentioned already. Switching tabs doesn't work with translations. We > probably shouldn't put translation to URI. This occurred in namespace detail, repo management as well.

Absolutely agreed, we should not put translated strings in the URL ... I suspect this is a place where we just assume that the machine-processable id and the human readable string are the same, or same +- capitalization, so we need an explicit map instead.

Are you looking into it? :)

> I also marked a strings that I think might be fit for the translation :)

:+1:

himdel:
(Shaiah feel free to review, but the bot has been pinging for 5 days, and the review has been done most of the time, removing the request.)

jerabekjiri:
- added ClipboardCopy and FileUpload wrapper components with translations
- reworked tabs, now they use `id` and `name` [code here](https://github.com/ansible/ansible-hub-ui/blob/1f1f2f7e29bbdc760102e5a3844ff602c76166ca/src/containers/edit-namespace/edit-namespace.tsx#L89-L92)
- Translated status and capitalized it in task list view
- Translated namespace, collections, users, groups, remotes, containers in Groups detail
- found some more strings for translation and fixed mentioned translations

What about translating date and time? 
Should we add `moment.locale` to DateComponent?

himdel:
> ...

:+1: I'll take a proper look in a moment :)

> What about translating date and time?
> Should we add moment.locale to DateComponent?

Good catch, yes, we need to be using localized date & time as well, both in DateComponent, and in CollectionInfo :+1:.

So yes, feel free to do that, moment.locale should work :).
But, if you're doing it, can you also look into what it would take to replace these with [Intl.RelativeTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat/RelativeTimeFormat) for `fromNow` and either [Intl.DateTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat), [Date.prototoype.toLocaleDateString](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString) or [Date.prototoype.toLocaleTimeString](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleTimeString) for `format`?

It *looks* like the native browser `Intl` API is now available everywhere, so if it suits our needs, it may be worth using :). (If not, I'm ok with moment-locale :).)

jerabekjiri:
Found out a similar bug to URL in tabs translations, but in Select component and added fix. 

before:
Screenshot from 2021-08-24 16-52-54

after:
Screenshot from 2021-08-24 16-48-46

src/containers/group-management/group-detail.tsx
```diff
@@ -284,7 +284,9 @@ class GroupDetail extends React.Component<RouteComponentProps, IState> {
               key={group.name}
               className={group.name}
             >
-              <FlexItem style={{ minWidth: '200px' }}>{group.name}</FlexItem>
+              <FlexItem style={{ minWidth: '200px' }}>
+                {twoWayMapper(group.name, Constants.GROUP_HUMAN_PERMISSIONS)}
```
himdel, line 288: it looks like GROUP_HUMAN_PERMISSIONS is translating the permission names, not the group names .. can we use a different Constant for these? :)

src/components/repo-selector/repo-selector.tsx
```diff
@@ -74,17 +76,21 @@ export class RepoSelector extends React.Component<IProps, IState> {
...
+                />
+              ))}
             </Select>
           </InputGroup>
         </FlexItem>
       </Flex>
     );
   }
 
-  private getRepoName(basePath) {
```
himdel, line 87: is the switch of getRepoName and getRepoBasePath intentional here? (No objections if it is (if it works), just surprise :))

src/constants.tsx
```diff
@@ -138,6 +138,12 @@ export class Constants {
     push_containerdistribution: _`Push distribution`,
     pull_containerdistribution: _`Pull distribution`,
     view_containerdistribution: _`View distribution`,
+    namespaces: _`Namespaces`,
```
himdel, line 141: ( https://github.com/ansible/ansible-hub-ui/pull/790/files#r699535818 )

himdel:
None of the changes like...

```diff
+    const username = <b>{user.username}</b>;
+    const groupname = <b>{group.name}</b>;
 
     return (
       <>
-        <b>{user.username}</b> will be removed from <b>{group.name}</b>.
+        {_`${username} will be removed from ${groupname}.`}
       </>
```

will work, they end up displaying `[object Object] will be removed from [object Object].`

For those we'll need to use `<Trans>` (https://lingui.js.org/tutorials/react.html#introducing-internationalization)..

```
-        <b>{user.username}</b> will be removed from <b>{group.name}</b>.
+        <Trans>{username} will be removed from {groupname}</Trans>
```

src/components/repo-selector/repo-selector.tsx
jerabekjiri: Yes, It's because `REPOSITORYNAMES`, I had to switch the keys for items, because we can't really translate like this
```
static REPOSITORYNAMES = {
    _`Published`:  'published',
   ...
  ``` 
  
  and with that, I switched names, so the method name does what it should :)

**Comment by himdel** [comment]
@jerabekjiri Heads up, https://github.com/ansible/ansible-hub-ui/pull/810 is merged, there should be nothing stopping you from using `<Trans` now :)

**Comment by jerabekjiri** [comment]
Fixed the merge conflicts and added some translation fixes. 

However, I found something strange happening in the search component:
Screenshot from 2021-09-03 14-18-57
and after translating
Screenshot from 2021-09-03 13-34-29

I tried changing from {0} to {selectTitleLC} and that didn't help as well
Screenshot from 2021-09-03 13-47-44

File: src/components/repositories/remote-form.tsx [hunk 8]
```diff
@@ -213,17 +220,11 @@ export class RemoteForm extends React.Component<IProps, IState> {
...
-                  <>
-                    This uses the same{' '}
-                    <a
-                      target='_blank'
-                      href='https://docs.ansible.com/ansible/latest/user_guide/collections_using.html#install-multiple-collections-with-a-requirements-file'
-                    >
-                      requirements.yml
-                    </a>{' '}
+                  <Trans>
+                    This uses the same {docsAnsibleLink}
```
himdel, line 224:
```
+#: src/components/repositories/remote-form.tsx:223
+msgid "This uses the same {docsAnsibleLink}format as the ansible-galaxy CLI with the caveat that roles aren't supported and the source 
parameter is not supported."
```

looks like we need to drop the newlines, or at least not have a newline after }, as the extractor strips the trailing whitespace.

src/containers/token/token-standalone.tsx
```diff
@@ -23,6 +23,8 @@ class TokenPage extends React.Component<RouteComponentProps, IState> {
   render() {
     const { token } = this.state;
 
+    const code = <code>ansible-galaxy</code>;
```
himdel, line 26: unused now?

src/loaders/standalone/standalone-loader.tsx
```diff
@@ -130,14 +130,16 @@ class App extends React.Component<RouteComponentProps, IState> {
           href='https://access.redhat.com/support'
           target='_blank'
         >
-          Customer Support <ExternalLinkAltIcon />
+          {t`Customer Support `}
```
himdel, line 133: trailing whitespace is stripped by the extractor, if you want to force a space after, it has to be ``{t`Customer Support`}{' '}``

himdel:
> However, I found something strange happening in the search component:

@jerabekjiri Good catch, the `test_l10n` logic didn't handle strings with params correctly, fixed in https://github.com/ansible/ansible-hub-ui/pull/874 :)

himdel (commented):
Except for the trailing whitespace issues above, LGTM :+1: :)

himdel (approved):
LGTM :+1: :)

PR #809, opened 2021-08-16
Enable external auth support and disabling of user/group mgmt

chambridge:
* Enable development setup for external authentication
* Disable user/group management

Related to https://github.com/ansible/galaxy_ng/pull/889


himdel:
Question, `isUserMgmtDisabled` seems to be used everywhere where there's a check for `user.model_permissions.{add_group,change_group,delete_group,add_user,change_user,delete_user}`

Would it make sense to just have the backend remove these from the list of permissions, instead of adding extra logic?

chambridge:
> Question, `isUserMgmtDisabled` seems to be used everywhere where there's a check for `user.model_permissions.{add_group,change_group,delete_group,add_user,change_user,delete_user}`
> 
> Would it make sense to just have the backend remove these from the list of permissions, instead of adding extra logic?

Thanks for the feedback! I will take a look to see if this can be updated in the backend.

chambridge:
@himdel I made the following update that reduced some of the usage you pointed out:
https://github.com/ansible/galaxy_ng/pull/889/files#diff-416aea42d5d99e187f7340543709cb3b126924b4fe30e6c36be869388c997845R132-R159

It was necessary to utilize the pattern for the user profile, to manage the human readable permission set, and for cases where permissions can be applied to groups (change_group permission); but you still need to hide adding/removing users from the group detail.

himdel:
Makes sense :+1: :)

One last thing I'm confused about...

src/loaders/standalone/routes.tsx
```diff
@@ -104,9 +105,13 @@ class AuthHandler extends React.Component<
     }
 
     if (!user && !noAuth) {
+      if (TARGET_ENVIRONMENT == Constants.DEV_TARGET_ENVIRONMENT) {
+        window.location.replace(UI_LOGIN_URI);
```
himdel, line 109: why do we need to `replace` instead of redirecting in dev env?

chambridge: The login + UI via Keycloak are hosted on the Django server so in the dev environment this is `http://localhost:5001/login/`. When I tried using redirect it just pushed the UI_LOGIN_URI on as a path to the existing `http://localhost:8002/ui`, so the redirect fails. The replace pushes it forward to the Keycloak login.

See the image below for the behavior that happened with the redirect:
image

himdel: aaah, it didn't recognize it as a full URI a tried to use it as a relative path, got it, thanks! :)

config/standalone.dev.webpack.config.js
```diff
@@ -4,6 +4,7 @@ const webpackBase = require('./webpack.base.config');
 const proxyHost = process.env.API_PROXY_HOST || 'localhost';
 const proxyPort = process.env.API_PROXY_PORT || '5001';
 const apiBasePath = process.env.API_BASE_PATH || '/api/automation-hub/';
+const uiLoginURI = process.env.UI_LOGIN_URI || '/login';
```
himdel, line 7: If we're just hardcoding it here, it's `/ui/login` for dev.

The `Paths` urls are relative to `UI_BASE_PATH`.

config/standalone.prod.webpack.config.js
```diff
@@ -10,5 +10,6 @@ module.exports = webpackBase({
   UI_USE_HTTPS: false,
   UI_DEBUG: false,
   TARGET_ENVIRONMENT: 'prod',
+  UI_LOGIN_URI: '/login',
```
himdel, line 13: also `/ui/login`

config/webpack.base.config.js
```diff
@@ -20,12 +20,13 @@ const defaultConfigs = [
   { name: 'DEPLOYMENT_MODE', default: 'standalone', scope: 'global' },
   { name: 'NAMESPACE_TERM', default: 'namespaces', scope: 'global' },
   { name: 'APPLICATION_NAME', default: 'Galaxy NG', scope: 'global' },
+  { name: 'UI_LOGIN_URI', default: '/login', scope: 'global' },
```
himdel, line 23: (probably makes sense to update the default as well)

chambridge:
@himdel Thanks for the above. I set the default to match what was defined int Paths here: https://github.com/ansible/ansible-hub-ui/pull/809/files#diff-ee5c055de3e53b22e883d37e2b0ef701779dfbd11691ca5a6504df271b53e008L29

I'm not sure updating it to `/ui/login` is actually the right fix. My concern now is that in a non-development build (i.e. static content in django server) the login path will still be `/ui/login` and not trigger the Keycloak login.

Perhaps you have a thought here?

himdel:
hm, I see what you mean .. we can't pass the same `UI_LOGIN_URI` both to `replace` and to  places which previously used `Paths.login`.

Sounds like we need to distinguish between external login uris for `replace` and internal login urls for `Redirect`.
Maybe a simple solution is to `` replace(UI_BASE + UI_LOGIN_URI) `` and keep `UI_LOGIN_URI='/login'`.

Though I wonder if we could make it simpler by leaving `Paths.login` as is, and adding `UI_EXTERNAL_LOGIN_URI` instead, which defaults to null, and whose presence always triggers a full `location.replace` redirect (regardless of dev vs prod).

chambridge:
@himdel I like the simplicity of the `UI_EXTERNAL_LOGIN_URI` option. I've reverted those previous changes. I also updated the server-side code so feature flags can be obtained if not authenticated so the login flow can be dynamically determined via that API.

https://github.com/ansible/galaxy_ng/pull/889/files#diff-85ab1792a81296d83b329359f5e9d8262535dad7cfda367e28ae6df427f2c2a6R9

newswangerd (changes requested):

src/containers/group-management/group-detail.tsx
```diff
@@ -569,7 +579,8 @@ class GroupDetail extends React.Component<RouteComponentProps, IState> {
           description={_`Users will appear once added to this group`}
           button={
             !!user &&
-            user.model_permissions.change_group && (
+            user.model_permissions.change_group &&
+            !isUserMgmtDisabled && (
```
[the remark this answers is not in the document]
chambridge: No, in this case we need change_group to be able to be set to true or false (unless you want to add a permission that differentiates between the ability to add users and the ability to set permissions on a group.

config/standalone.dev.webpack.config.js
```diff
@@ -33,6 +34,9 @@ module.exports = webpackBase({
   // Target compilation environment. Options: dev, prod
   TARGET_ENVIRONMENT: 'dev',
 
+  // Login URI to allow stand alone with and without keycloak
+  UI_EXTERNAL_LOGIN_URI: uiExternalLoginURI,
```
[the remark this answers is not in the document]
chambridge: I had tried that originally; the problem is that the redirect only redirects within the UI_BASE_PATH.

himdel:
@chambridge I'm on PTO this week, removing the review request for now, feel free to re-add :)

If you do end up going the route of "UI_EXTERNAL_LOGIN_URI can either be an absolute url, or a relative one absolute to UI root", this is currently misconfigured as it's missing the UI_BASE_PATH prefix, at least for standalone.

(Unless `/login` is considered the external url as everything else is `/ui/something`.)
Also, can you write down the intended defaults please? I'm *pretty* sure standalone.dev should have that null, not /login. (Or /ui/login instead of /login.) (I'm guessing standalone devel mode is never using external auth, as for the others, I can't tell the intent.)

chambridge:
@newswangerd based on my responses above I'm not sure what you would like me to update. Feel free to take a look at the responses and give an update; I'll re-request a review.

config/standalone.dev.webpack.config.js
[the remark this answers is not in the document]
newswangerd: Thanks for confirming! Shame it doesn't work.

newswangerd:
> @newswangerd based on my responses above I'm not sure what you would like me to update. Feel free to take a look at the responses and give an update; I'll re-request a review.

Looks like you addressed all my questions.

newswangerd (approved):

ZitaNemeckova:
In dev. env. it breaks with `UI_EXTERNAL_LOGIN_URI`.

chambridge:
@ZitaNemeckova Do you have some logs or a link? I'm not encountering an error so I'm having trouble reproducing to try and resolve.

ZitaNemeckova:
@chambridge It was my env. not having UI in sync with backend. Sorry for false alarm.

PR #812, opened 2021-08-18
Fix wrong setting of default sort

ZitaNemeckova:
Default sort is not set correctly. Go to Task Management. See arrow next to `Created on`

Before:
Screenshot 2021-08-18 at 12 28 59

After:
Screenshot 2021-08-18 at 12 27 21


jerabekjiri:
LGTM :), just a small thing, from a user perspective, wouldn't be better to sort it from the newest created to oldest by default?

PR #813, opened 2021-08-19
Fix missing titles in Alert notifications

jerabekjiri:
Issue: [AAH-828](https://issues.redhat.com/browse/AAH-828)

Fixed missing titles (null) and replaced them with descriptions. 
I noticed that in some cases alerts don't show like when creating or editing users for example, should this be fixed as well or it's intentional?

before:
Screenshot from 2021-08-19 12-11-33

after:
Screenshot from 2021-08-19 12-13-19


ZitaNemeckova (approved):
LGTM 👍

PR #815, opened 2021-08-20
Group list

MilanPospisil:
Test for paging, sorting and filtering of Group list page.

### DIFF
--- a/test/cypress/integration/group_list.js
+++ b/test/cypress/integration/group_list.js
@@ -22,50 +22,45 @@ describe('Group list tests for sorting, paging and filtering', () => {
 
   beforeEach(() => {
     cy.login(adminUsername, adminPassword);
+    cy.visit('/ui/group-list');
   });
 
   it('table contains all columns', () => {
-    cy.contains('Group');
+    cy.get('tr[aria-labelledby="headers"] th').contains('Group');
   });
 
   it('items are sorted alphabetically and paging is working', () => {
-    cy.visit('/ui/group-list');
-    cy.get('.body:first').contains(items[0].name);
+    cy.get('.body').contains(items[0].name);
 
-    cy.get('.body:first').get('[aria-label="Go to next page"]:first').click();
-    cy.get('.body:first').contains(items[10].name);
+    cy.get('.body').get('[aria-label="Go to next page"]:first').click();
+    cy.get('.body').contains(items[10].name);
 
-    cy.get('.body:first').get('[aria-label="Go to next page"]:first').click();
-    cy.get('.body:first').contains(items[20].name);
+    cy.get('.body').get('[aria-label="Go to next page"]:first').click();
+    cy.get('.body').contains(items[20].name);
   });
 
   it('sorting is working', () => {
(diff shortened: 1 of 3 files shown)


test/cypress/integration/group_list.js
```diff
@@ -0,0 +1,71 @@
...
+
+    items = sortBy(items, 'name');
+  });
+
+  beforeEach(() => {
+    cy.login(adminUsername, adminPassword);
+  });
+
+  it('table contains all columns', () => {
+    cy.contains('Group');
```
himdel, line 28: we're on the Group list screen.. there will always be *something* that contains "Group"

If the idea is to make sure we have all the columns, what about...

`cy.contains('tr[aria-labelledby="headers"] th', 'Group');`

himdel: Also, there is no navigation. This is probably finding "Group" in the menus.

Navigate to the right screen first, `cy.menuGo("User Access > Groups")` should do the trick :)

test/cypress/integration/group_list.js
```diff
@@ -0,0 +1,71 @@
...
+  });
+
+  it('items are sorted alphabetically and paging is working', () => {
+    cy.visit('/ui/group-list');
+    cy.get('.body:first').contains(items[0].name);
```
himdel, line 33: `.body` is the section containing the whole table. There is always just one.

It looks like the intent was to check the first line in the table contains this, so...

`cy.contains('tbody tr:first td:first', items[0].name);`

test/cypress/integration/group_list.js
```diff
@@ -0,0 +1,71 @@
...
+    cy.get('.body:first').get('[aria-label="Go to next page"]:first').click();
+    cy.get('.body:first').contains(items[10].name);
+
+    cy.get('.body:first').get('[aria-label="Go to next page"]:first').click();
+    cy.get('.body:first').contains(items[20].name);
+  });
+
+  it('sorting is working', () => {
+    cy.visit('/ui/group-list');
+    cy.get('.body:first').get('[data-cy="sort_name"]').click();
```
himdel, line 44: presumably..

```
cy.get('[data-cy="sort_name"]').click();
cy.contains('tbody tr:first td:first', items[20].name);
cy.contains('tbody tr td', items[0].name).should('not.exist');
```

```

test/cypress/integration/group_list.js
MilanPospisil: That is totally obvious, I forgot :-) And it finds group substring even in the table :-D

test/cypress/integration/group_list.js
MilanPospisil: Ok, I think I had one issue with it in the past, but that was maybe some exception.

test/cypress/integration/group_list.js
MilanPospisil: You are right, it is good idea to test if the item is right in the first line. But the item, that should not existed, should be tested in whole page, not only in first row, so I leave the contains in the whole body.

PR #832, opened 2021-08-26
Rework of CompoundFilter component

jerabekjiri:
Issue: [AAH-354](https://issues.redhat.com/browse/AAH-354)

Reworked CompoundFilter component to be controlled from outside, this allows clearing the `inputText`.

In `task-list-view.tsx` I had to move `loading` from the whole page just to the table, this was causing to break the CompoundFilter on multiple values. 

Not part of the issue, but I fixed the empty filter submit. Not entirely sure, but this feels like a bug, because it doesn't really filter anything and happens only on press enter. :arrow_down: 
Screenshot from 2021-08-26 13-33-43

I noticed on a lot of places in `AppliedFilters` we don't have `niceNames` values, I can fix this as part of this issue or new issue can be created for that.

### DIFF
--- a/src/containers/execution-environment/registry-list.tsx
+++ b/src/containers/execution-environment/registry-list.tsx
@@ -58,6 +58,7 @@ interface IState {
   remoteUnmodified?: RemoteType;
   showDeleteModal: boolean;
   showRemoteFormModal: boolean;
+  inputText: string;
 }
 
 class ExecutionEnvironmentRegistryList extends React.Component<
@@ -92,6 +93,7 @@ class ExecutionEnvironmentRegistryList extends React.Component<
       remoteUnmodified: null,
       showDeleteModal: false,
       showRemoteFormModal: false,
+      inputText: '',
     };
   }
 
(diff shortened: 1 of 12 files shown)


ZitaNemeckova:
@jerabekjiri Given you have plenty on your plate right now, please set this to draft. You can finish it once you have more time :)

jerabekjiri:
Good idea :)

src/containers/execution-environment/registry-list.tsx
```diff
@@ -240,6 +242,10 @@ class ExecutionEnvironmentRegistryList extends React.Component<
                       <ToolbarGroup>
                         <ToolbarItem>
                           <CompoundFilter
+                            inputText={this.state.inputText}
+                            onChange={(text) =>
+                              this.setState({ inputText: text })
+                            }
```
himdel, line 248: is this file missing a `this.setState({ inputText: '' });` in AppliedFilters.updateParams ?

himdel:
if/when backporting to 4.4, make sure to re-apply #1195 (backported faster)

PR #833, opened 2021-08-27
Repair Group management sporadic failure

MilanPospisil:
Another problem in group management was not waiting good enough for user to be deleted by gui. We added contains at the end of deleteUser command. The app works fine. 

o be honest Im not sure why this exactly works, because without it, galaxykit fails with 403 on select groups (in deleteTestGroups command in beforeEach). 

What was wrong previously:
Before this PR, It fails sometimes in situation after the deleteUser command. It then runs beforeEach which runs deleteTestGroups.
And in deleteTestGroups, there is galaxykit which wants to select groups and this fails with 403.

Removing deleteGroup and deleteUser solves the problem also, but we need to test deletion using GUI. But if we remove only deleteUser, it works. It was maybe because deleteGroup did waited using contains to ensure group is deleted. So we added this contains also to deleteUser and it started working.

### DIFF
--- a/test/cypress/support/commands.js
+++ b/test/cypress/support/commands.js
@@ -350,6 +350,7 @@ Cypress.Commands.add('deleteUser', {}, (username) => {
 
   // Wait for navigation
   cy.wait('@userList');
+  cy.contains(username).should('not.exist');
 });
 
 Cypress.Commands.add('deleteGroup', {}, (name) => {


himdel:
The failure this is trying to fix...

```
<?xml version="1.0" encoding="UTF-8"?>
<testsuites name="Mocha Tests" time="20.3460" tests="3" failures="1">
  <testsuite name="Root Suite" timestamp="2021-08-30T15:52:57" tests="0" file="cypress/integration/group_management.js" time="0.0000" failures="0">
  </testsuite>
  <testsuite name="Hub Group Management Tests" timestamp="2021-08-30T15:52:57" tests="3" time="20.3460" failures="1">
    <testcase name="Hub Group Management Tests admin user can create/delete a group" time="6.9920" classname="admin user can create/delete a group">
    </testcase>
    <testcase name="Hub Group Management Tests admin user can add/remove a user to/from a group" time="13.3540" classname="admin user can add/remove a user to/from a group">
    </testcase>
    <testcase name="Hub Group Management Tests &quot;before each&quot; hook for &quot;admin user can add/remove permissions to/from a group&quot;" time="0.0000" classname="&quot;before each&quot; hook for &quot;admin user can add/remove permissions to/from a group&quot;">
      <failure message="The following error originated from your application code, not from Cypress. It was caused by an unhandled promise rejection.

  &gt; Request failed with status code 403

When Cypress detects uncaught errors originating from your application it will automatically fail the current test.

This behavior is configurable, and you can choose to turn this off by listening to the `uncaught:exception` event.

https://on.cypress.io/uncaught-exception-from-application

Because this error occurred during a `before each` hook we are skipping the remaining tests in the current suite: `Hub Group Management Tests`" type="Error"><![CDATA[Error: The following error originated from your application code, not from Cypress. It was caused by an unhandled promise rejection.

  > Request failed with status code 403

When Cypress detects uncaught errors originating from your application it will automatically fail the current test.

This behavior is configurable, and you can choose to turn this off by listening to the `uncaught:exception` event.

https://on.cypress.io/uncaught-exception-from-application

Because this error occurred during a `before each` hook we are skipping the remaining tests in the current suite: `Hub Group Management Tests`
    at e.exports (http://localhost:8002/static/galaxy_ng/js/vendor.0d8ae4ebfa7000c56b88.js:2:27407)
    at e.exports (http://localhost:8002/static/galaxy_ng/js/vendor.0d8ae4ebfa7000c56b88.js:2:29809)
    at XMLHttpRequest.m.onreadystatechange (http://localhost:8002/static/galaxy_ng/js/vendor.0d8ae4ebfa7000c56b88.js:2:23627)
    at XMLHttpRequest.eval (eval at makeContentWindowListener (http://localhost:8002/__cypress/runner/cypress_runner.js:160967:10), <anonymous>:4:29)]]></failure>
    </testcase>
  </testsuite>
</testsuites>
```

for example in https://github.com/ansible/ansible-hub-ui/runs/3398730289?check_suite_focus=true#step:13:159

himdel:
> because without it, galaxykit fails with 403 on select groups (in deleteTestGroups command in beforeEach).

Not quite sure what it means, galaxykit doesn't interact with the UI, there are no selects in galaxykit.

If the failure is in/during `deleteTestGroups` it's either the GET request failing, or an individual DELETE request failing, or an unrelated UI request happening at the same time. Given the error, I think it's the last one (because galaxykit never outputs "Request failed with status code" AFAICT), caused by some request being fired after the test ended and cypress forgot cookies.

But agreed that waiting longer on that screen should fix it :)

---

Tested locally...

```bash
cd test/
rm cypress/integration/*
git checkout cypress/integration/group_management.js
for n in {1..50}; do echo; echo TEST $n; echo ; npm run cypress:chrome || echo ERR $n >> errors; done
wc -l errors
```

On master, this fails in 15 of the 50 runs,
with this PR there are still 3 failure out of 50, but different ones.

And this still brings us from 30% failure rate down to 6%, LGTM :+1:, merging, thanks :)

(The new 6% failure is https://gist.github.com/himdel/7dc3e4d342494ebd195894f80e4390c6 , I think this is just the extra `cy.login` calls inside the delete* helpers, removing in a separate PR.)

himdel:
(`backport-4.3` will do so separately, together with the other relevant changes for 4.3)

PR #840, opened 2021-08-30
User list test

MilanPospisil:
Test sorting, paging, filtering for user list.

### DIFF
--- a/test/cypress/integration/user_list.js
+++ b/test/cypress/integration/user_list.js
@@ -34,6 +34,17 @@ describe('User list tests for sorting, paging and filtering', () => {
     cy.contains('Email');
     cy.contains('Groups');
     cy.contains('Created');
+
+    [
+      'Username',
+      'First name',
+      'Last name',
+      'Email',
+      'Groups',
+      'Created',
+    ].forEach((item) => {
+      cy.get('tr[aria-labelledby="headers"] th').contains(item);
+    });
   });
 
   it('table contains some time informations for new users', () => {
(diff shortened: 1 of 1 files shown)


test/cypress/integration/user_list.js
```diff
@@ -0,0 +1,79 @@
...
+    items = sortBy(items, 'name');
+  });
+
+  beforeEach(() => {
+    cy.login(adminUsername, adminPassword);
+    cy.visit('/ui/users');
+  });
+
+  it('table contains all columns', () => {
+    cy.contains('Username');
```
himdel, line 31: same issues as in https://github.com/ansible/ansible-hub-ui/pull/815/files#r698711313, this needs to look in the headers, not the whole page

himdel: What about something like...

```js
['Username', 'First name', 'Last name', 'Email', 'Groups', 'Created'].forEach((column) => 
  cy.contains('tr[aria-labelledby="headers"] th', column)
);
```

MilanPospisil: Good idea.

himdel: :+1: you may still want to remove the first version ;)

(and maybe rebase, now that #815 is in, for the sort-table change)

MilanPospisil: Ok.

test/cypress/integration/user_list.js
```diff
@@ -0,0 +1,88 @@
...
+    cy.contains('First name');
+    cy.contains('Last name');
+    cy.contains('Email');
+    cy.contains('Groups');
+    cy.contains('Created');
+
```
himdel, line 37:
```suggestion
```

PR #856, opened 2021-08-30
group_manager - more test fixes

himdel:
Follow-up to #833,

`cy.deleteUser` and `cy.deleteGroup` are now used only in `group_management`, which is always logged in as admin => we no longer need to logout & login as admin (previously changing the login to admin was needed for uses replaced by `deleteTestUsers`). Removing, no other helpers change login :).

`group_management` is doing `deleteTestUsers` & `deleteTestGroups` cleanup in `beforeEach`, instead of doing it just once in `before` - fixing for consistency as well :). The individual tests in group_management actually do their own cleanup.

`deleteUser` and `deleteGroup` have over-eager intercepts, intercepting `DELETE .../group/123/model_permission/456` when expecting to intercept `DELETE .../group/123` => changed `**` to `?*` or `*` to be more specific

---

This *may* fix a ~6% sporadic failure in group_management, as described in https://github.com/ansible/ansible-hub-ui/pull/833#issuecomment-908633723:
https://gist.github.com/himdel/7dc3e4d342494ebd195894f80e4390c6
`api/automation-hub/_ui/v1/auth/login/ ... 400: Bad Request`

(but not sure, as I can't test this reliably, seems like the model_permission request storm from `addAllPermissions` kills the pulp worker)

### DIFF
--- a/test/cypress/integration/group_management.js
+++ b/test/cypress/integration/group_management.js
@@ -2,10 +2,12 @@ describe('Hub Group Management Tests', () => {
   var adminUsername = Cypress.env('username');
   var adminPassword = Cypress.env('password');
 
-  beforeEach(() => {
+  before(() => {
     cy.deleteTestGroups();
     cy.deleteTestUsers();
+  });
 
+  beforeEach(() => {
     cy.login(adminUsername, adminPassword);
   });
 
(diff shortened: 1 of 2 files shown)


himdel:
(note to self: backport to 4.3 together with #833)

himdel:
Testing locally...

```
cd test/
rm cypress/integration/*
git checkout cypress/integration/group_management.js
for n in {1..50}; do echo; echo TEST $n; echo ; npm run cypress:chrome || echo ERR $n >> errors; done
wc -l errors
```

...

ZitaNemeckova (approved):
LGTM 👍

PR #857, opened 2021-08-30
Group detail - wait for Permissions to save, fix Cancel

himdel:
we're currently firing up to 19 requests when saving the group Permissions form,
this seems to be an API tech debt, but for now, we need to wait for those requests to finish before starting something else.

This disables the save/cancel buttons when saving,
and waits for all the requests to complete before changing back to view state.

---

Also fixed the cancel button to restore the original state, so that it doesn't seem like Cancel applies the changes :)

---

This is vaguely related to #833 and #856, as firing those 19 requests is either related to the remaining sporadic failure (more in #856), or preventing me from testing the fix.

Test-wise, the effect of this is that `cy.addAllPermissions` and `cy.removeAllPermissions` doesn't leave some `model_permissions` API requests hanging, and waits for their completion.

### DIFF
--- a/src/containers/group-management/group-detail.tsx
+++ b/src/containers/group-management/group-detail.tsx
@@ -60,6 +60,7 @@ interface IState {
   options: { id: number; name: string }[];
   selected: { id: number; name: string }[];
   editPermissions: boolean;
+  savingPermissions: boolean;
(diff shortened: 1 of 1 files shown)


ZitaNemeckova (approved):
Tested. LGTM 👍

PR #859, opened 2021-08-31
Fix Users Redirect

MilanPospisil:
Redirect fix for users - user detail, create and edit. This fix uses Redirect component instead of pushing to history.

### DIFF
--- a/src/containers/user-management/user-edit.tsx
+++ b/src/containers/user-management/user-edit.tsx
@@ -39,7 +39,7 @@ class UserEdit extends React.Component<RouteComponentProps, IState> {
 
   render() {
     if (this.state.redirect) {
-      return <Redirect to={this.state.redirect} />;
+      return <Redirect push to={this.state.redirect} />;
     }
 
     const { user, errorMessages, unauthorized } = this.state;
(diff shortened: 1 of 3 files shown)


src/containers/user-management/user-edit.tsx
```diff
@@ -82,9 +87,9 @@ class UserEdit extends React.Component<RouteComponentProps, IState> {
       .then(() => {
         //redirect to login page when password of logged user is changed
         if (this.context.user.id === user.id && user.password) {
-          this.props.history.push(Paths.login);
+          this.setState({ redirect: Paths.login });
```
himdel, line 90: Heads up: redirecting to login is special, because we also support external login, and Redirect can only redirect to local URLs.

But, when using external login, it's not possible to edit users or groups, so UserEdit can safely use Redirect :+1: 

(Just something to watch out for, more in#809.)

MilanPospisil: Ok, thanks for info. So its only your note, right? Im not supposed to do anything with it.

himdel: Right :)

himdel:
Soo, this all works, but there is one difference .. `history.push` *adds* a new history item, while `Redirect` replaces the existing one (so the back button no longer works).

Looks like Redirect has a [`push`](https://reactrouter.com/web/api/Redirect/push-bool) prop that makes it push, can you add it please? :)

himdel (approved):

PR #860, opened 2021-08-31
eslint config with typescript support, some basic rules

himdel:
This re-enables eslint, making `npm run lint:js` run both eslint and prettier,
configures it to use the typescript-capable `@typescript-eslint/parser`,
starts off with `eslint:recommended`, `react/recommended`, `@typescript-eslint/recommended` and `prettier` rules,

and for now just disables anything that creates errors, as we can address individual rules in follow-up PRs
(except a couple of single-file issues fixed manually here)

(The current config generated too many errors, we can add individual rules as needed in separate PRs as well, leaving in a reminder comment for now.)


ZitaNemeckova (approved):
LGTM :+1:

PR #864, opened 2021-09-01
Fix Groups redirect using Redirect component

MilanPospisil:
Fix redirect in group detail using Redirect component instead of history.


src/containers/group-management/group-detail.tsx
```diff
@@ -130,6 +136,10 @@ class GroupDetail extends React.Component<RouteComponentProps, IState> {
   }
 
   render() {
+    if (this.state.redirect) {
+      return <Redirect to={this.state.redirect} />;
```
himdel, line 140: :+1:, but also https://github.com/ansible/ansible-hub-ui/pull/859#issuecomment-910920191

```suggestion
      return <Redirect push to={this.state.redirect} />;
```

MilanPospisil: Ok.

himdel (approved):

PR #865, opened 2021-09-01
Fix of redirect in namespaces list and detail

MilanPospisil:
Fix of redirect in namespaces list and detail using Redirect component instead of history.


src/containers/namespace-list/namespace-list.tsx
```diff
@@ -99,6 +100,10 @@ export class NamespaceList extends React.Component<IProps, IState> {
   }
 
   render() {
+    if (this.state.redirect) {
+      return <Redirect to={this.state.redirect} />;
```
himdel, line 104: :+1:, but also https://github.com/ansible/ansible-hub-ui/pull/859#issuecomment-910920191

```suggestion
      return <Redirect push to={this.state.redirect} />;
```

MilanPospisil: Ok.

himdel (approved):

PR #866, opened 2021-09-01
Fix redirect in settings/user-profile.tsx

MilanPospisil:
Fix redirect in settings/user-profile.tsx using Redirect component instead of history.

### DIFF
--- a/src/containers/settings/user-profile.tsx
+++ b/src/containers/settings/user-profile.tsx
@@ -1,5 +1,10 @@
 import * as React from 'react';
-import { withRouter, RouteComponentProps, Link } from 'react-router-dom';
+import {
+  withRouter,
+  RouteComponentProps,
+  Link,
+  Redirect,
+} from 'react-router-dom';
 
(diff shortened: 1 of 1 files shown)


src/containers/settings/user-profile.tsx
```diff
@@ -46,10 +52,14 @@ class UserProfile extends React.Component<RouteComponentProps, IState> {
         this.initialState = { ...result };
         this.setState({ user: result });
       })
-      .catch(() => this.props.history.push(Paths.notFound));
+      .catch(() => this.setState({ redirect: Paths.notFound }));
   }
 
   render() {
+    if (this.state.redirect) {
+      return <Redirect to={this.state.redirect} />;
```
himdel, line 60: :+1:, but also https://github.com/ansible/ansible-hub-ui/pull/859#issuecomment-910920191

```suggestion
      return <Redirect push to={this.state.redirect} />;
```

MilanPospisil: Ok.

PR #867, opened 2021-09-01
cypress: update galaxykit

himdel:
new galaxykit was released, making sure we're up to date :)

Cc @hendersonreed Does it make sense to just remove the version constraint so we're always up to date?
(I assume no breaking changes? :))

(Looks like we're going from 0.1.0 to 0.5.1. ([pip](https://pypi.org/project/galaxykit/))

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -22,7 +22,7 @@ jobs:
 
     - name: "Install galaxykit dependency"
       run: |
-        pip install galaxykit==0.1.0
+        pip install galaxykit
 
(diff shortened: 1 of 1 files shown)


hendersonreed:
I think removing the version constraint should be fine here, since you're using the CLI, which undergoes fewer changes. There were breaking changes in this release but only for the consumers of the Python module, not the CLI.

So I say go for it!

himdel:
Thanks! :)

ZitaNemeckova (approved):
LGTM :+1:

PR #868, opened 2021-09-01
cypress reporter: use spec for output, junit for test/results/

himdel:
changes the cypress standard output from XML to

  Token Management Tests
    ✓ user can open docs dropdown menu (2065ms)
    ✓ user can toggle about modal (2260ms)

(while keeping `test/results/` in junit format - I'm assuming it's there for a reason? :))

### DIFF
--- a/test/cypress.json
+++ b/test/cypress.json
@@ -2,9 +2,8 @@
   "viewportWidth": 1200,
   "viewportHeight": 800,
   "baseUrl":"http://localhost:8002",
-  "reporter": "junit",
+  "reporter": "cypress-multi-reporters",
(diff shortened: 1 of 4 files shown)


ZitaNemeckova (approved):
LGTM 👍  but I'd like @ironfroggy or @hendersonreed to have a look as well, please.

hendersonreed:
This looks good to me, I don't think we have anything relying on the output of these tests at the moment :smile:

himdel:
Aah, good to know, thanks :) Might make sense to experiment with the output format for nicer errors on github :)

PR #871, opened 2021-09-02
Namespace deletion

jerabekjiri:
Issue: [AAH-714](https://issues.redhat.com/browse/AAH-714)

- Added`DeleteCollectionModal` that could be used for deleting collections, namespaces and containers
- With working backend for deleting namespace, I added logic for deleting namespace on UI
- Namespace can be deleted only if no collections in the namespace

- Delete namespace modal
Screenshot from 2021-09-03 00-36-08

- Error alert if something went wrong
Screenshot from 2021-09-07 14-31-35

- Success alert 
Screenshot from 2021-09-07 14-24-02
 
 Success alert is shown on the namespace list page, the flow goes like that: 
1. DELETE namespace 
2. If success -> redirect to namespacelist
3. show success alert


ZitaNemeckova:
I have a `arista` namespace with a published collection. When a go to arita detail page and change a repository from `pusblished` to `community` I can try to delete it but it fails with `{"errors":[{"status":"400","code":"invalid","title":"Invalid input.","detail":"Namespace arista cannot be deleted because there are still collections associated with it."}]}`

src/containers/collection-detail/delete-collection-modal.tsx
```diff
@@ -0,0 +1,123 @@
...
+          label={t`I understand that this action cannot be undone.`}
+          id='delete_confirm'
+        />
+      </DeleteModal>
+    );
+  }
+
+  private deleteCollection(collection) {}
+
+  private deleteNamespace({ name }) {
```
ZitaNemeckova, line 106: It would be better to move this function to the container and pass it as a prop.

src/containers/collection-detail/delete-collection-modal.tsx
```diff
@@ -0,0 +1,123 @@
...
+          this.setState({ confirmDelete: false });
+          closeModal();
+        }}
+        deleteAction={() =>
+          namespace
+            ? this.deleteNamespace(namespace)
+            : this.deleteCollection(collection)
+        }
+        title={
+          namespace
```
ZitaNemeckova, line 53: It would be better to pass title and description in params so the component as dumb as possible.

ZitaNemeckova (changes requested):
It's on a good way :) There's a fwe things that need rework.
- There' s "collection deletion" logic that should come in it's own PR. 
- Any component should be as dumb as possible. The logic should stay in a container if possible. 
- Why not use `DeleteModal` component instead of create a new component?

jerabekjiri:
- Created a new component `DeleteModalWithConfirm` which uses `DeleleModal`, but already includes confirmation UI part and logic for that, this could be nice for reusing later, since all content deletion modals are same and could be reused like that :)
- Fixed mentioned error by getting all namespace's repositories and checking if it has collections.
- Removed collection deletion leftovers
> Any component should be as dumb as possible. The logic should stay in a container if possible.

Good suggestion, thanks, It really made no sense mixing collection and namespace into one delete modal, also when there's container deletion and might others in future :grin: :+1:

jerabekjiri:
I will investigate that test fail more, my Chrome crashes on this local test, but in Firefox it passes normally.

src/api/namespace.ts
```diff
@@ -6,6 +6,10 @@ class API extends HubAPI {
   constructor() {
     super();
   }
+
+  deleteNamespace(name) {
```
himdel, line 10: `NamespaceAPI.delete(name)` already exists from the base class

ZitaNemeckova (approved):
LGTM 👍 There's a style that should be moved to sccs file in followup PR.

PR #896, opened 2021-09-06
CollectionsFilter / Roles filter: use /_ui/v1/tags/{collections,roles} API to populate filter tags

himdel:
Depends on https://github.com/ansible/galaxy_ng/pull/1931

role list:

(typeahead in filter steals focus from input field when typing - fixing as separate issue, unrelated to tags)

role namespace role list:

(tags counts are NOT scoped to the namespace)

collections:

(tag counts will appear with newer pulp_ansible - the ui can display them as long as the api serves them)

### DIFF
--- a/src/containers/ansible-role/role-list.tsx
+++ b/src/containers/ansible-role/role-list.tsx
@@ -1,7 +1,7 @@
 import { t } from '@lingui/macro';
 import { DataList } from '@patternfly/react-core';
 import React from 'react';
-import { LegacyRoleAPI, LegacyRoleListType } from 'src/api';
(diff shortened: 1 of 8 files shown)


src/containers/ansible-role/role-list.tsx
```diff
@@ -80,6 +82,21 @@ class AnsibleRoleList extends React.Component<RouteProps, RolesState> {
       );
   }
 
+  // TODO call - but only HubListToolbar knows inputText
```
himdel, line 85: use `HubListToolbar` w/ `typeaheads={{ tags: this.loadTags }}` (and take `inputText` as param, not state)

PR #869, opened 2021-09-02
Redirect changed to Redirect push to to keep history.

MilanPospisil:
Refactoring of Redirect to => Redirect push to. It adds to browser history instead of replacing the url, so back can work properly.

PR #897, opened 2021-09-06
Redirect: add missing push

himdel:
Follow-up to #869 (change `<Redirect` to `<Redirect push` to preserve browser history)

a couple of Redirects were missed in #869, updating :)
Cc @MilanPospisil


ZitaNemeckova (approved):
LGTM 👍

PR #900, opened 2021-09-07
Enable view-only collection browsing

ZitaNemeckova:
Depends on https://github.com/ansible/galaxy_ng/pull/909

`GALAXY_ENABLE_UNAUTHENTICATED_COLLECTION_ACCESS = True` to allow anonymous user to view collections and namespaces and documentation
`GALAXY_ENABLE_UNAUTHENTICATED_COLLECTION_DOWNLOAD = True` to allow anonymous user to download a collection

- [x] Guest user has Login button instead of profile
- [x] Logging in will set a user correctly
- [x] Guest user can download a collection only when `GALAXY_ENABLE_UNAUTHENTICATED_COLLECTION_DOWNLOAD = True`
- [x] Allow user to go to Collections without logging in
- [x] Guest user cannot access other pages via URL
- [x] Remove View only button on Login page

Screenshot 2021-09-07 at 15 56 02
Screenshot 2021-09-07 at 12 41 57


src/components/collection-detail/collection-info.scss
```diff
@@ -33,3 +33,6 @@
 .info-panel {
   padding: 8px;
 }
+.pf-c-alert.pf-m-inline.pf-m-warning > h4 {
```
himdel, line 36: this will affect all inline alerts in the app, is that intentional? or missing some custom class to ensure this only happens on the right screen?

src/containers/namespace-detail/namespace-detail.tsx
```diff
@@ -340,11 +340,19 @@ export class NamespaceDetail extends React.Component<IProps, IState> {
         this.context.selectedRepo,
       ),
       NamespaceAPI.get(this.props.match.params['namespace']),
-      MyNamespaceAPI.get(this.props.match.params['namespace']).catch(
+      MyNamespaceAPI.get(this.props.match.params['namespace']).catch((e) => {
+        // TODO this needs fixing on backend to return nothing in these cases with 200 status
+        // if view only mode is enabled disregard errors and hope
+        if (
+          this.context.settings.GALAXY_ENABLE_UNAUTHENTICATED_COLLECTION_ACCESS
```
himdel, line 347: this should check that you are actually using the anonymous user, otherwise it eats all errors when the setting is enabled

src/loaders/standalone/standalone-loader.tsx
```diff
@@ -328,31 +337,41 @@ class App extends React.Component<RouteComponentProps, IState> {
           url: formatPath(Paths.searchByRepo, {
             repo: this.state.selectedRepo,
           }),
+          condition: ({ user, settings }) =>
+            !!user || settings.GALAXY_ENABLE_UNAUTHENTICATED_COLLECTION_ACCESS,
```
himdel, line 341: are these obsolete now?

or can `user` still be null, while ..being `is_anonymous` elsewhere?

src/containers/user-management/user-detail.tsx
```diff
@@ -41,24 +42,33 @@ class UserDetail extends React.Component<RouteComponentProps, IState> {
...
       return <Redirect push to={this.state.redirect} />;
     }
 
-    const { userDetail, errorMessages, alerts, showDeleteModal } = this.state;
+    const { userDetail, errorMessages, alerts, showDeleteModal, unauthorised } =
+      this.state;
     const { user } = this.context;
 
+    if (unauthorised) {
+      return <EmptyStateUnauthorized />;
```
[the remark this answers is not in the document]
ZitaNemeckova: @himdel Actually you want to show unauthorised empty state when you don't have rights not redirect to login by default. I fixed so the access logic is at ne place not two different. 

And `GALAXY_ENABLE_UNAUTHENTICATED_COLLECTION_ACCESS` doesn't let you make any irelevant API calls.

src/components/collection-detail/collection-info.scss
ZitaNemeckova: Fixed :)

src/containers/namespace-detail/namespace-detail.tsx
ZitaNemeckova: Fixed :)

src/loaders/standalone/standalone-loader.tsx
ZitaNemeckova: Fixed :)

src/containers/user-management/user-detail.tsx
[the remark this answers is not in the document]
himdel: Aaah, makes sense, thanks! :)

himdel:
Looks good :+1: , but the tests seem to fail now.. (and a rebase is needed)

src/loaders/standalone/standalone-loader.tsx
ZitaNemeckova: `settings.GALAXY_ENABLE_UNAUTHENTICATED_COLLECTION_ACCESS || !user.is_anonymous` because `settings.GALAXY_ENABLE_UNAUTHENTICATED_COLLECTION_ACCESS` is `False` by default.

himdel (approved):
with `GALAXY_ENABLE_UNAUTHENTICATED_COLLECTION_ACCESS = False`:

all screens work as intended :+1:  (redirect to login)

with `GALAXY_ENABLE_UNAUTHENTICATED_COLLECTION_ACCESS = True`:

all screens work as intended :+1: (show Unauthorized)
except for the newly added EE > Registries screen which just spins now, but I can fix this separately :)

ZitaNemeckova:
@himdel fixed EE > Registries.

himdel (approved):

PR #920, opened 2021-09-07
Add chrome 2 configurations.

Hyperkid123:
part of: https://issues.redhat.com/browse/RHCLOUD-15377
Fixes [AAH-672](https://issues.redhat.com/browse/AAH-672)

Adds chrome 2 setup and configuration to the build. has to be merged with: https://github.com/RedHatInsights/cloud-services-config/pull/893. The CSC changes will deploy to ci/qa/stage beta. After looking at your release script, it should be the same for this repository. Deployments to other rnvironments have to be coordinated.

cc @ZitaNemeckova


config/webpack-ts-overrides.js
```diff
@@ -1,12 +1,33 @@
...
+  __dirname,
+  `../src/${isBuild ? 'entry.js' : 'entry-dev.js'}`,
+);
+
 module.exports = {
   devtool: 'source-map',
   module: {
     rules: [
+      /**
+       * Chrome render loader blocks the ReactDOM.render call if the application is loaded within the platfrom.
```
himdel, line 16: Um.. would this be the same as *not* calling ReactDOM.render in entry.js and entry-dev.js?

We're not actually using those files for anything else, so can we actually change it there?

config/webpack.base.config.js
```diff
@@ -99,6 +99,25 @@ module.exports = (inputConfigs) => {
   plugins.push(new webpack.DefinePlugin(globals));
   plugins.push(new ForkTsCheckerWebpackPlugin());
 
+  /**
+   * Generates remote containers for chrome 2
+   */
+  plugins.push(
+    require('@redhat-cloud-services/frontend-components-config/federated-modules')(
```
himdel, line 106: what does this do, would it affect standalone builds in any way?
(As in, do we need to wrap this so it doesn't get called in standalone?)

himdel: (I'm guessing the answer is yes, given the build failure...

```
2021-09-07T14:16:35.7388801Z ERROR in ./src/loaders/insights/Routes.js 71:9-73:54
2021-09-07T14:16:35.7390679Z Module not found: Error: Can't resolve '../../containers/namespace-list/namespaces' in '/home/runner/work/ansible-hub-ui/ansible-hub-ui/ansible-hub-ui/src/loaders/insights'
2021-09-07T14:16:35.7393562Z resolve '../../containers/namespace-list/namespaces' in '/home/runner/work/ansible-hub-ui/ansible-hub-ui/ansible-hub-ui/src/loaders/insights'
```

config/webpack-ts-overrides.js
Hyperkid123: IF you don't use them, then it can be removed. I am trying to make as few changes as I can. I will remove it and also the render call.

config/webpack.base.config.js
Hyperkid123: I am not 100% convinced those two things are related. The plugin creates a new webpack remote container entry point but does not change the rest of the application. I was running the prod build locally and it was passing :thinking: I will mimic what the CI does and see what's up with it.

Hyperkid123: @himdel the output log is correct. The file `src/containers/namespace-list/namespaces.tsx` doesn't exist so there is nothing to import.

I assume it is configured here: https://github.com/ansible/ansible-hub-ui/blob/master/config/standalone.prod.webpack.config.js#L9 and used here: https://github.com/ansible/ansible-hub-ui/blob/master/src/loaders/insights/Routes.js#L88

The only `namespances.*` file I have found is in the cypress tests.

What I don't understand is, why the build passes if I remove the plugin :thinking:. I will dig through it a bit more.

Hyperkid123: OK I think I understand now. The module federation plugin uses a different method for async chunk optimization, mroe specifically code splitting.

Normally webpack would not complain about an async chunk not existing at build time and would pass the build. The error would be present only at runtime. But because the file actually does not exist and webpack is trying to move it to a proper chunk due to enabled module sharing (react and react-dom, pf, fec and others are shared across the platform to reduce the bandwidth). And since the file does not exist, it fails to do so. 

But I see you have `customConfigs.DEPLOYMENT_MODE` variable, so we can wrap the plugin in a condition and should be OK.

himdel: Thanks so much for looking into this! I didn't even consider we had the path wrong, sorry about that :)

Hyperkid123: I've added the condition anyway.

himdel:
Tested start-standalone, build-standalone :+1: 

With insights mode, I get an empty screen with just the menu and navbar, the main area is blank (except for the lightbulb)..


Is that expected?

profiles/local-frontend-and-api.js
```diff
@@ -14,7 +14,7 @@ routes[`/${SECTION}/${APP_ID}`]      = { host: `https://${localhost}:${FRONTEND_
 routes[`/beta/apps/${APP_ID}`]       = { host: `https://${localhost}:${FRONTEND_PORT}` };
 routes[`/apps/${APP_ID}`]            = { host: `https://${localhost}:${FRONTEND_PORT}` };
 
-routes[`/api/automation-hub/`]       = { host: `http://${localhost}:${API_PORT}` };
-routes[`/static/`]       = { host: `http://${localhost}:${API_PORT}` };
+// routes[`/api/automation-hub/`]       = { host: `http://${localhost}:${API_PORT}` };
```
himdel, line 17: should this also happen in `profiles/galaxy-ng-compose.js`? (That's the one that seems to be used by the docker env.)

Hyperkid123: @himdel in order fo the chrome2 version to work, the CSC must be updated. Did you try running it with this PR: https://github.com/RedHatInsights/cloud-services-config/pull/893?

himdel: > @himdel in order fo the chrome2 version to work, the CSC must be updated. Did you try running it with this PR: RedHatInsights/cloud-services-config#893?

How would I do that? It doesn't seem to be a dependency of anything on our side.

But, if you're confident it works, maybe we can just merge it and start the process of getting this to stage&beta...

Hyperkid123: @himdel there are docs here for local testing: https://github.com/RedHatInsights/cloud-services-config#testing-your-changes-locally

> But, if you're confident it works, maybe we can just merge it and start the process of getting this to stage&beta...

I am I was running it locally myself :slightly_smiling_face: we just have to make the CSC PR makes it into the same branches as this code.

Hyperkid123:
@himdel please ping me once you will be merging the PR. I will make sure the chrome config is up to date so we don't have an outage.

ryelo:
@himdel Any updates on this? We're currently blocked on releasing until this is merged and released.

himdel:
@ryelo We just deployed the first up-to-date release to stage last week, I think it just needs to go through QE now and we're unstuck, I'll check.

We definitely need to wait for QE before this gets to prod-beta, prod-stable, but we *may* be able to merge this to master / stage now.

Hyperkid123:
@himdel if we could get this to the stage env that would be great. It will unblock some of our tasks.

If you decide to merge this change, let me know. We need to coordinate with CSC changes in here: https://github.com/RedHatInsights/cloud-services-config/pull/893

himdel:
:+1:

The current plan is to merge and deploy to stage tomorrow, before US working hours,
then we'll need to wait for QE approval before getting it to prod, and sync up on the deploy.

ryelo:
Awesome. Thanks! Let us know if you run into any issues and work with @Hyperkid123 to get the CSC PR merged in.

PR #921, opened 2021-09-08
Translation checked.

MilanPospisil:
Checking for translation in the whole src folder.

### DIFF
--- a/src/components/execution-environment/repository-form.tsx
+++ b/src/components/execution-environment/repository-form.tsx
@@ -95,8 +95,8 @@ export class RepositoryForm extends React.Component<IProps, IState> {
             label={t`Groups with access`}
           >
             <div className='pf-c-form__helper-text'>
-              {t`Adding groups provides access to all repositories in the "
-                ${namespace}" container namespace.`}
+              {t`Adding groups provides access to all repositories in the
+                "${namespace}" container namespace.`}
             </div>
             <ObjectPermissionField
               groups={this.state.selectedGroups}
--- a/src/components/execution-environment/repository-form.tsx
+++ b/src/components/execution-environment/repository-form.tsx
@@ -95,8 +95,8 @@ export class RepositoryForm extends React.Component<IProps, IState> {
             label={t`Groups with access`}
           >
             <div className='pf-c-form__helper-text'>
-              Adding groups provides access to all repositories in the "
-              {namespace}" container namespace.
+              {t`Adding groups provides access to all repositories in the "
+                ${namespace}" container namespace.`}
             </div>
             <ObjectPermissionField
               groups={this.state.selectedGroups}
--- a/src/components/group-management/group-modal.tsx
+++ b/src/components/group-management/group-modal.tsx
@@ -42,7 +42,6 @@ export class GroupModal extends React.Component<IProps, IState> {
         isOpen={true}
         title={''}
         header={<h2>{t`Create a group`}</h2>}
-        aria-label={t`group-modal`}
         actions={[
           <Button
             isDisabled={
--- a/src/components/group-management/group-modal.tsx
+++ b/src/components/group-management/group-modal.tsx
@@ -42,7 +42,7 @@ export class GroupModal extends React.Component<IProps, IState> {
         isOpen={true}
         title={''}
         header={<h2>{t`Create a group`}</h2>}
-        aria-label='group-modal'
+        aria-label={t`group-modal`}
         actions={[
           <Button
             isDisabled={
--- a/src/components/markdown-editor/markdown-editor.tsx
+++ b/src/components/markdown-editor/markdown-editor.tsx
@@ -25,7 +25,7 @@ export class MarkdownEditor extends React.Component<IProps, {}> {
               <FormGroup fieldId='resources' helperText={helperText}>
                 <div id='markdown-title'>{t`Raw Markdown`}</div>
                 <TextArea
-                  aria-labelledby='markdown-title'
+                  aria-labelledby={t`markdown-title`}
                   id='resources'
                   value={text}
                   onChange={(value) => updateText(value)}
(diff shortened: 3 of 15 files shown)


src/components/group-management/group-modal.tsx
```diff
@@ -42,7 +42,7 @@ export class GroupModal extends React.Component<IProps, IState> {
         isOpen={true}
         title={''}
         header={<h2>{t`Create a group`}</h2>}
-        aria-label='group-modal'
+        aria-label={t`group-modal`}
```
himdel, line 45: group-modal is not a string we should be translating, or showing to the user.

But [aria-label](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute) **is** something that can be shown to the user.

=> this is an existing bug, we should either remove this label, or provide an actual label.

I suggest removing, the modal already has a header, unless the prop is required.

src/components/markdown-editor/markdown-editor.tsx
```diff
@@ -25,7 +25,7 @@ export class MarkdownEditor extends React.Component<IProps, {}> {
               <FormGroup fieldId='resources' helperText={helperText}>
                 <div id='markdown-title'>{t`Raw Markdown`}</div>
                 <TextArea
-                  aria-labelledby='markdown-title'
+                  aria-labelledby={t`markdown-title`}
```
himdel, line 28: never translate `aria-labelledby`, it's expected to contain the id of the element with a proper label

src/components/execution-environment/repository-form.tsx
```diff
@@ -95,8 +95,8 @@ export class RepositoryForm extends React.Component<IProps, IState> {
             label={t`Groups with access`}
           >
             <div className='pf-c-form__helper-text'>
-              Adding groups provides access to all repositories in the "
-              {namespace}" container namespace.
+              {t`Adding groups provides access to all repositories in the "
```
himdel, line 98: the line break in the middle of `"${namespace}"` is problematic (adds an extra whitespace between the quote and the param), can you move the quote to the next line?

src/components/group-management/group-modal.tsx
MilanPospisil: Ok.

src/components/markdown-editor/markdown-editor.tsx
MilanPospisil: Ok.

MilanPospisil: Ok.

src/components/execution-environment/repository-form.tsx
MilanPospisil: Ok.

src/utilities/write-only-fields.ts
```diff
@@ -9,7 +10,7 @@ export function isFieldSet(
     console.log(field);
     return field.is_set;
   } else {
-    throw `Field ${name} is not in writeOnlyFields`;
+    throw t`Field ${name} is not in writeOnlyFields`;
```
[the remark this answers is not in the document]
MilanPospisil: Ok.

src/components/patternfly-wrappers/toolbar.tsx
```diff
@@ -75,12 +76,12 @@ export class Toolbar extends React.Component<IProps, IState> {
...
                   onKeyPress={(e) => this.handleEnter(e)}
                   type='search'
-                  aria-label='search text input'
+                  aria-label={t`search text input`}
                   placeholder={searchPlaceholder}
                 />
                 <Button
                   variant={ButtonVariant.control}
-                  aria-label='search button'
+                  aria-label={`search button`}
```
himdel, line 84: missing `t`?

src/containers/execution-environment-detail/execution_environment_detail_activities.tsx
```diff
@@ -128,23 +128,29 @@ class ExecutionEnvironmentDetailActivities extends React.Component<
                   if (!!removed) {
                     activityDescription = (
                       <React.Fragment>
-                        <TagLink tag={action.tag_name} /> was moved to{' '}
-                        <ShaLink digest={action.manifest_digest} /> from
-                        <ShaLink digest={removed.manifest_digest} />
+                        <Trans>
+                          <TagLink tag={action.tag_name} /> was moved to{' '}
+                          <ShaLink digest={action.manifest_digest} /> from
```
himdel, line 133:
```suggestion
                          <ShaLink digest={action.manifest_digest} /> from{' '}
```

(otherwise `npm run gettext:extract` extracts this as `"<0/> was moved to <1/> from<2/>"`)

src/utilities/write-only-fields.ts
```diff
@@ -1,4 +1,5 @@
 import { WriteOnlyFieldType } from 'src/api';
+import { t } from '@lingui/macro';
```
himdel, line 2: (unused export now)

src/components/patternfly-wrappers/toolbar.tsx
MilanPospisil: I will check aria labels using search, It seems I forgot to do it. I only check files manually, but this can be checked automatically.

src/containers/execution-environment-detail/execution_environment_detail_activities.tsx
MilanPospisil: Ok.

src/utilities/write-only-fields.ts
MilanPospisil: Ok.

src/components/sort-table/sort-table.tsx
```diff
@@ -75,7 +76,7 @@ export class SortTable extends React.Component<IProps> {
   render() {
     return (
       <thead>
-        <tr aria-labelledby='headers'>
+        <tr aria-labelledby={t`headers`}>
```
himdel, line 79: sounds like you need to redo the automatic bit without aria-labelledby :)
only aria-label should have human strings, aria-labelledby is using element IDs

MilanPospisil: That was not using automatic, I used <aria-label='> pattern and indeed, last commit did not contain this. This must be from manual checking, I will find all aria-labelledby to see if it is right.

MilanPospisil: Wierd, in my local version, there is no mistake...

himdel (approved):

PR #923, opened 2021-09-08
Simple test for Task list

MilanPospisil:
Using repository management/remote, we click Sync button and generate tasks. Then in Tasks list, we do basic check for table headers and other simple things.

### DIFF
--- a/test/cypress/integration/task_list.js
+++ b/test/cypress/integration/task_list.js
@@ -10,8 +10,14 @@ describe('Task table contains correct headers and filter', () => {
       'POST',
       Cypress.env('prefix') + '/content/rh-certified/v3/sync/',
     ).as('sync');
+
+    cy.intercept('GET', Cypress.env('prefix') + '_ui/v1/remotes/?*').as(
+      'remotes',
+    );
+
     cy.contains('button', 'Sync').click();
     cy.wait('@sync');
+    cy.wait('@remotes');
   });
 
   it('table contains all columns and filter', () => {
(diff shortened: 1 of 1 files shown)


test/cypress/integration/task_list.js
```diff
@@ -0,0 +1,22 @@
+describe('Task table contains correct headers and filter', () => {
+  let adminUsername = Cypress.env('username');
+  let adminPassword = Cypress.env('password');
+
+  before(() => {
+    cy.login(adminUsername, adminPassword);
+    cy.visit('/ui/repositories?tab=remote');
+    cy.contains('button', 'Sync').click();
```
himdel, line 8: this is potentially introducing another sporadic failure, the Sync button triggers a http request, and a redirect/reload after the response

we need to wait for the new page to load before `before` ends :)

MilanPospisil: Ok. But since we are testing now only headers on the table, the sync is not needed for the test anyway. But you are right, at least, we will test if sync dont fail.

himdel: yeah, I'm more worried about the redirect cancelling our navigation, even when we wait for the request, the end of `before`, `cy.login`, and `cy.visit` could still get called before the Redirect has a chance to do anything.

Would it be possible to check that the `Last synced` column has change, to ensure we're waiting for the reload?

MilanPospisil: I will try wait also for the remote request.

himdel (approved):
LGTM :)

PR #942, opened 2021-09-14
Task management detail test

MilanPospisil:
Simple test for task management detail. First, we click Sync button on remote repo. Then we check for very basic things in task detail page.

### DIFF
--- a/test/cypress/integration/task_management_detail.js
+++ b/test/cypress/integration/task_management_detail.js
@@ -11,9 +11,9 @@ describe('Task detail', () => {
       Cypress.env('prefix') + '/content/rh-certified/v3/sync/',
     ).as('sync');
 
-    let remotes = Cypress.env('prefix') + '_ui/v1/remotes/?*';
-
-    cy.intercept('GET', remotes).as('remotes');
+    cy.intercept('GET', Cypress.env('prefix') + '_ui/v1/remotes/?*').as(
+      'remotes',
+    );
 
     cy.contains('button', 'Sync').click();
     cy.wait('@sync');
--- a/test/cypress/integration/task_management_detail.js
+++ b/test/cypress/integration/task_management_detail.js
@@ -36,11 +36,6 @@ describe('Task detail', () => {
       'Parent task',
       'Child task',
       'Reserve resources',
-      'Type',
-      'Name',
-      'remotes',
-      'rh-certified',
-      'repositories',
     ].forEach((item) => {
       cy.contains('.card-area', item);
     });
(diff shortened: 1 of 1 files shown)


test/cypress/integration/task_management_detail.js
```diff
@@ -0,0 +1,38 @@
...
+    cy.contains('button', 'Sync').click();
+    cy.wait('@sync');
+  });
+
+  it('contains correct headers and field names.', () => {
+    cy.login(adminUsername, adminPassword);
+    cy.visit('/ui/tasks');
+    cy.contains('Pulp Ansible: Collections sync').click();
+
+    cy.contains('Pulp Ansible: Collections sync');
```
himdel, line 22: something like `cy.contains('tr[aria-labelledby="headers"] th', '...');`?
(https://github.com/ansible/ansible-hub-ui/pull/840#discussion_r698719005)

test/cypress/integration/task_management_detail.js
```diff
@@ -0,0 +1,38 @@
...
+  before(() => {
+    cy.login(adminUsername, adminPassword);
+    cy.visit('/ui/repositories?tab=remote');
+
+    cy.intercept(
+      'POST',
+      Cypress.env('prefix') + '/content/rh-certified/v3/sync/',
+    ).as('sync');
+    cy.contains('button', 'Sync').click();
+    cy.wait('@sync');
```
himdel, line 14: just waiting for the request might not be enough, a redirect follows
(https://github.com/ansible/ansible-hub-ui/pull/923#discussion_r708449133)

test/cypress/integration/task_management_detail.js
MilanPospisil: I tried something, but Im not sure if I understood you well.

himdel: Pretty close :+1: 

the idea was https://github.com/ansible/ansible-hub-ui/pull/840/files#diff-097758856b3f7f2993b38bbdc8c19ecdfd2084b8184064908d9713b6f5ec0345R30-R41

though looks like only some of these here are headers?

What about two lists, one using the same selector as the users/groups list test,
and the other ... I don't know what those things are, so maybe adjust the selectors to be clear what you're actually searching for? (the test seems to indicate that `remotes`  should be on the page, but doesn't make it clear what it actually is)

MilanPospisil: Maybe you confused this PR with PR that tests tasks list? This is task detail, there are no table headers, everything is divs and sections.

himdel (approved):
LGTM, thanks :)

PR #944, opened 2021-09-15
Collection dependencies tab

jerabekjiri:
Issue: [AAH-752](https://issues.redhat.com/browse/AAH-752)

- Added displaying the number of dependencies in the `CollectionList` page. I had to add a plural option for the `NumericLabel` component. It was implemented just to put `s` at the end of the keyword, so what was happening => `123 Dependencys` and what we needed `2 Dependencies` (`2 Dependencys` => `2 Dependencies`).
Screenshot from 2021-09-15 01-56-48

### Dependencies page
- added `CollectionDependenciesList` component to display list of dependencies
- added `CollectionUsedbyDependenciesList` component to show all collections that are using this collection. 

Screenshot from 2021-09-14 23-03-58

- Should we show an empty state on no `dependencies` or `usedby dependencies` list?
- Now, when I'm thinking about it, maybe `reverseDependencies` would be a better variable name than `usedByDependencies`? :smile:

To generate test collections run `make api/create-test-collections`. More here [#936](https://github.com/ansible/galaxy_ng/pull/936)

I still have some issues with generating test collections and therefore `Filter collection name` Input that uses `/collection-versions/?dependency=namespace.collection&name?=test_dependency` doesn't seem to be working on my end (pagination and sort works though), so I'll test this more when I figure it out.

### DIFF
--- a/src/api/response-types/collection.ts
+++ b/src/api/response-types/collection.ts
@@ -11,6 +11,7 @@ export class CollectionVersion {
     contents: ContentSummaryType[];
     description: string;
     tags: string[];
+    dependencies: any; //FIXME: any for testing
   };
   created_at: string;
   // contents: ContentSummaryType[]; // deprecated
@@ -35,6 +36,7 @@ export class CollectionVersionDetail extends CollectionVersion {
     documentation: string;
     issues: string;
     repository: string;
+    dependencies: any;
   };
   requires_ansible?: string;
   docs_blob: DocsBlobType;
--- a/src/api/response-types/collection.ts
+++ b/src/api/response-types/collection.ts
@@ -11,7 +11,7 @@ export class CollectionVersion {
     contents: ContentSummaryType[];
     description: string;
     tags: string[];
-    dependencies: any; //FIXME: any for testing
+    dependencies: DependencyType[];
   };
   created_at: string;
   // contents: ContentSummaryType[]; // deprecated
@@ -36,7 +36,7 @@ export class CollectionVersionDetail extends CollectionVersion {
     documentation: string;
     issues: string;
     repository: string;
-    dependencies: any;
+    dependencies: DependencyType[];
   };
   requires_ansible?: string;
   docs_blob: DocsBlobType;
@@ -122,4 +122,8 @@ export class CollectionDetailType {
 
 export class CollectionUsedByDependencies extends CollectionDetailType {
   version: string;
+}
+
+export class DependencyType {
+  [namespaceCollection: string]: string;
 }
--- a/src/api/response-types/collection.ts
+++ b/src/api/response-types/collection.ts
@@ -119,3 +119,7 @@ export class CollectionDetailType {
     company: string;
   };
 }
+
+export class CollectionUsedByDependencies extends CollectionDetailType {
+  version: string;
+}
--- a/src/components/collection-dependencies-list/collection-usedby-dependencies-list.tsx
+++ b/src/components/collection-dependencies-list/collection-usedby-dependencies-list.tsx
@@ -4,9 +4,14 @@ import { Link } from 'react-router-dom';
 
 import { CollectionUsedByDependencies } from 'src/api';
 
-import { Pagination, Toolbar } from 'src/components';
+import {
+  Pagination,
+  Toolbar,
+  EmptyStateNoData,
+  EmptyStateFilter,
+} from 'src/components';
 
-import { ParamHelper } from 'src/utilities/param-helper';
+import { ParamHelper, filterIsSet } from 'src/utilities';
 import { formatPath, Paths } from 'src/paths';
 
 import 'src/containers/collection-detail/collection-dependencies.scss';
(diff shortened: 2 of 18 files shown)


jerabekjiri:
- fixed plural labels to apply to CollectionCard as well and reworked it to Constants
- added bulleted list to `CollectionDependenciesList`
- added empty state screens
  - Empty state when collection doesn't have dependencies and usedby dependencies 
Screenshot from 2021-09-15 20-38-19
  - Empty state when no dependencies in collection
Screenshot from 2021-09-15 21-32-03
 - Empty state when no usedby dependencies in collection 
Screenshot from 2021-09-15 21-30-52
  - Empty state on no results in usedby dependencies after filtering 
  
Screenshot from 2021-09-15 22-39-43

I'll ask Chris B. if the empty state texts are correct. And my collection filtering still doesn't work, need to look into it more :)

src/api/response-types/collection.ts
```diff
@@ -11,6 +11,7 @@ export class CollectionVersion {
     contents: ContentSummaryType[];
     description: string;
     tags: string[];
+    dependencies: any; //FIXME: any for testing
```
ZitaNemeckova, line 14: Please create a dependency type instead of `any` :)

src/components/collection-dependencies-list/collection-usedby-dependencies-list.tsx
```diff
@@ -0,0 +1,137 @@
...
+                <SearchInput
+                  value={params.name || ''}
+                  onChange={(val) =>
+                    updateParams(ParamHelper.setParam(params, 'name', val))
+                  }
+                  onClear={() =>
+                    updateParams(ParamHelper.setParam(params, 'name', ''))
+                  }
+                  aria-label='filter-collection-name'
+                  placeholder={t`Filter collection name`}
```
ZitaNemeckova, line 71: Missing `by` and maybe `collection` is not needed -> "Filter by name".

src/containers/collection-detail/collection-dependencies.tsx
```diff
@@ -0,0 +1,206 @@
...
+      .catch(() => {
+        this.props.history.push(Paths.notFound);
+      });
+  }
+
+  get updateParams() {
+    return ParamHelper.updateParamsMixin();
+  }
+
+  private combineParams(...params) {
```
ZitaNemeckova, line 199: Unused function. Please remove.

src/containers/collection-detail/collection-dependencies.tsx
```diff
@@ -0,0 +1,206 @@
...
+              )}
+            </div>
+          </section>
+        </Main>
+      </React.Fragment>
+    );
+  }
+
+  private loadData(forceReload = false) {
+    this.loadCollection(forceReload, () => this.loadUsedByDependencies());
```
ZitaNemeckova, line 163: There should be a loading logic so a loading spinner aprears before data is loaaded instead of EmptyState.

ZitaNemeckova:
@jerabekjiri It seems to me that API endpoint doesn't filter/sort by name yet so I would drop the filter/sort and add it once the functionality is there. WDYT?

jerabekjiri:
@ZitaNemeckova thanks for the feedback, the filtering by name should be working after this [#974](https://github.com/ansible/galaxy_ng/pull/974)

src/components/collection-dependencies-list/collection-usedby-dependencies-list.tsx
jerabekjiri: According to this [Dependencies prototype](https://marvelapp.com/prototype/a0c7h0j/screen/80780016/handoff) it's `Filter collection name`. Should I change it to `Filter by name` or ask Susan to verify?

ZitaNemeckova: Let's verify with @sbuenafe-rh :)

src/containers/collection-detail/collection-dependencies.tsx
jerabekjiri: I'm using both functions in code, mostly like that
```
updateParams={(p) => {
   this.updateParams(this.combineParams(this.state.params, p), () =>
       this.loadData(true),
   );
}}
```

ZitaNemeckova: My IDE lied to me :D Please ignore :)

jerabekjiri:
@ZitaNemeckova Fixed the issues, added the `LoadingPageSpinner` component for loading `usedByDependencies`, added the alert if an error occurs in loading dependencies request. Also PR#974 is merged now, so filtering by name should be working now.

cbudz:
Screens:

Empty state when no dependencies in collection

- replace "This collection is dependent on the following collections" with "This collections depends on the following collections" or even "This collections requires the following collections for use"

Empty state when no usedby dependencies in collection

- replace "No collection is using this dependency" with "Not required for use by other collections"

jerabekjiri:
Updated screen with changed messages 
Screenshot from 2021-09-22 23-20-59

ZitaNemeckova (approved):
LGTM 👍 

Note: Need to ask API to do partial match.

PR #970, opened 2021-09-21
Change EmptyStateNoData icon from SearchIcon to CubesIcon

ZitaNemeckova:
Before:
Screenshot 2021-09-21 at 12 44 21

After:
Screenshot 2021-09-21 at 12 43 28


ZitaNemeckova:
@himdel EmptyStateFilter should have `SearchIcon` as according to https://www.patternfly.org/v4/components/empty-state/ I cannot find guidelines for EmptyState with button that we use but it's according to Susan.

himdel (approved):

PR #979, opened 2021-09-22
Add image delete

ZitaNemeckova:
Screenshot 2021-09-22 at 14 19 47
Screenshot 2021-09-22 at 14 19 57

### DIFF
--- a/src/containers/execution-environment-detail/execution_environment_detail_images.tsx
+++ b/src/containers/execution-environment-detail/execution_environment_detail_images.tsx
@@ -3,7 +3,11 @@ import * as React from 'react';
 import './execution-environment-detail.scss';
 
 import { pickBy } from 'lodash';
-import { ExecutionEnvironmentAPI, ContainerManifestType, TaskAPI } from 'src/api';
+import {
+  ExecutionEnvironmentAPI,
+  ContainerManifestType,
+  TaskAPI,
+} from 'src/api';
 import { formatPath, Paths } from 'src/paths';
 import {
   ParamHelper,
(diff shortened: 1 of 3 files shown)


src/containers/execution-environment-detail/execution_environment_detail_images.tsx
```diff
@@ -156,6 +182,30 @@ class ExecutionEnvironmentDetailImages extends React.Component<
           alerts={this.state.alerts}
           closeAlert={(i) => this.closeAlert(i)}
         />
+        {deleteModalVisible && (
+          <DeleteModal
+            title={'Permanently delete image'}
```
himdel, line 187: missing l10n

src/containers/execution-environment-detail/execution_environment_detail_images.tsx
```diff
@@ -156,6 +182,30 @@ class ExecutionEnvironmentDetailImages extends React.Component<
...
+                deleteModalVisible: false,
+                selectedImage: null,
+                confirmDelete: false,
+              })
+            }
+            deleteAction={() => this.deleteImage()}
+            isDisabled={!confirmDelete}
+          >
+            <Trans>
+              Deleting <b>{selectedImage.digest}</b> and its data will be lost.
```
himdel, line 199: you may want to `const { digest } = selectedImage` somewhere, so that the thing inside {} is not an expression

(`{0}` vs `{digest}` for translators)

himdel (approved):
Works for me :+1: 

(there seems to be an API error where this causes a `list index outr of range` API error on all subsequent requests, but this doesn't seem to be UI related)

PR #1001, opened 2021-09-29
Task detail page refreshes when that task not done

ZitaNemeckova:
Fixes https://issues.redhat.com/browse/AAH-954

Task detail page refreshes after 10s if the task is not done yet. No refresh when the task is done.

### DIFF
--- a/src/containers/task-management/task_detail.tsx
+++ b/src/containers/task-management/task_detail.tsx
@@ -421,7 +421,6 @@ class TaskDetail extends React.Component<RouteComponentProps, IState> {
             let url = resource.replace('/pulp/api/v3/', '');
             let id = parsePulpIDFromURL(url);
             let urlParts = resource.split('/');
-            console.log(urlParts);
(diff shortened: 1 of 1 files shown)


src/containers/task-management/task_detail.tsx
```diff
@@ -70,7 +70,16 @@ class TaskDetail extends React.Component<RouteComponentProps, IState> {
 
   componentDidMount() {
     this.loadContent();
+    const refresh = setInterval(() => this.loadContent(), 10000);
+    if (
```
himdel, line 74: this logic would have to live in loadContent, otherwise it can only cancel the refresh on page load

himdel: (or, since you know the task id, you can do `waitForTask(...).then(() => clearInterval)`)

ZitaNemeckova: Fixed. I'm not making the interval when not necessary and cleaning it in all possible cases.

himdel (approved):
Works for me :+1: :)

PR #1004, opened 2021-09-29
RemoteForm - don't mention the token when disabled

himdel:
Fixes [AAH-950](https://issues.redhat.com/browse/AAH-950)

Remove the "this is not required when using a token" bit from username/password helper text when the token is disabled (the EE Registry form).


ZitaNemeckova (approved):
LGTM 👍

PR #872, opened 2021-09-02
EE: Publish to Controller

himdel:
Fixes [AAH-889](https://issues.redhat.com/browse/AAH-889)
API https://github.com/ansible/galaxy_ng/pull/941 (controllers; merged) & https://github.com/ansible/galaxy_ng/pull/957 (tags; merged)
Design https://marvelapp.com/prototype/27ee1cd6/screen/81904383

This allows publishing execution environments to Controller.

The target URL Controller-side is `$CONTROLLER_HOST/#/execution_environments/add?image=$OUR_HOST/$image_name:$tag` or `...?image=$OUR_HOST/sha256:$digest` (`OUR_HOST` is without protocol, as in the `podman pull` command).

(Remaining TODO: "Learn more" link - right url)

Accessible from:

* [x] EE list - list action
  * preselects last modified tag
  * 
* [x] EE detail (Detail, Activity, Images) header
  * preselects last modified tag
  * 
* [x] EE Images list - list action
  * preselects last modified tag for chosen digest
  * 
* (no links from container manifest)

---

Extra changes:

Also changes the `Edit tags` button to `Manage tags`, to fit the modal title.

merge `ImagesAPI` into `ExecutionEnvironmentsAPI` - every image-related method except `list` was already there.

`APISearchTypeAhead`: add `toggleIcon` prop, pass to `Select` (to allow showing the tag icon)

`ShaLabel`: add `grey` & `long` bool props 

`getContainersURL`: moved the "give me the server for podman pull" logic to a helper

`EmptyStateNoData`, `EmptyStateCustom` - allow description to be a react node, not ony string

---

Testing:

~~Needs~~ a change in `galaxy_ng/galaxy_ng/app/settings.py` to see any controllers:
```diff
diff --git a/galaxy_ng/app/settings.py b/galaxy_ng/app/settings.py
index 7a84dfb..beeecac 100644
--- a/galaxy_ng/app/settings.py
+++ b/galaxy_ng/app/settings.py
@@ -314,1 +314,1 @@
-CONNECTED_ANSIBLE_CONTROLLERS = []
+CONNECTED_ANSIBLE_CONTROLLERS = ['https://example.com']
```

OR You can add `PULP_CONNECTED_ANSIBLE_CONTROLLERS="['https://example.com']"` to `.compose.env`

PR #1005, opened 2021-09-29
Use in Controller: grid layout for top of the table

himdel:
This is fixing the alignment issue left unresolved in https://github.com/ansible/ansible-hub-ui/pull/872#issuecomment-926116168.

Cc @sbuenafe-rh .. which option do you think would be better, aling EE,Tag & Digest:

or just Tag & Digest:


sbuenafe-rh:
@himdel you're right, aligning EE, Tag, & Digest looks better.  Thanks for showing the options!

himdel:
Thanks, going with EE&Tag&Digest then :)

ZitaNemeckova (approved):
LGTM :+1:

PR #1032, opened 2021-10-04
Loading spinner when deleting collection and namespace content

jerabekjiri:
No issue

Quick PR addition for loading spinner in a modal window when deleting collection and namespace, sometimes the request takes longer, so the user knows the request is still pending :)
Screenshot from 2021-10-04 17-42-26


himdel:
Nice catch :+1:,
should we also do the same for container deletions?

jerabekjiri:
@himdel I think it works well and it was already implemented by Zita in `ConfirmModal` component

PR #1035, opened 2021-10-04
EE list, EE detail - alert on sync success/error

himdel:
Fixes [AAH-967](https://issues.redhat.com/browse/AAH-967)

This just adds an alert on sync success/failure on both the EE list and EE detail screens.

The logic is copied from Remote Registries sync, I verified this does show an error alert nicely :).
(There seems to be an API issue where sync 500s now, but theoretically it should show the task link on success :) .. fixed in https://github.com/ansible/galaxy_ng/pull/1006)


ZitaNemeckova (approved):
LGTM 👍

PR #1037, opened 2021-10-05
Fix breadcrumbs in EE detail page to make it consistent with nav

ZitaNemeckova:
Before:
Screenshot 2021-10-05 at 12 18 35

After:
Screenshot 2021-10-05 at 12 16 54


himdel:
Good catch :+1:,
we also have these...

```
src/containers/execution-environment-list/execution_environment_list.tsx
306:          title: t`Container registry type`,

src/containers/execution-environment-manifest/execution-environment-manifest.tsx
100:                  name: t`Container Registry`,
```

PR #1038, opened 2021-10-05
Fix tabs in EE

ZitaNemeckova:
Fixes https://issues.redhat.com/browse/AAH-969

How to reproduce:
- Go to http://localhost:8002/ui/containers
- Click on a container
- Click on any tab
- Click on it again

The page loads correctly first time and second time there's nothing

Before:

Screenshot 2021-10-05 at 12 19 51

After:

Screenshot 2021-10-05 at 12 21 52


himdel (approved):
Works for me :+1:  :)

(Eventually we should make it so clicking the tab again reloads it, but that can wait for page.)

PR #1044, opened 2021-10-06
display backend authentication provider on user detail page

ShaiahWren:
Good morning Zita && Martin :)

Issue: [AAH-952](https://issues.redhat.com/browse/AAH-952)

Assignment: Display the backend authentication provider on the user detail page. If user is not an SSO user this should return 'django,' if user _is_ SSO user, that auth provider should be displayed. 

Note: Testing should be done with someone who has Keycloak working. 

Before:
Screen Shot 2021-10-06 at 10 37 05 AM

After:
Screen Shot 2021-10-06 at 10 36 30 AM

Have a nice day :)

### DIFF
--- a/src/api/response-types/user.ts
+++ b/src/api/response-types/user.ts
@@ -42,6 +42,7 @@ export class UserType {
   last_name?: string;
   email?: string;
   groups: { id: number; name: string }[];
+  auth_provider?: [];
   date_joined?: string;
   password?: string;
   model_permissions?: Permissions;
--- a/src/components/user-form/user-form.tsx
+++ b/src/components/user-form/user-form.tsx
@@ -120,6 +120,19 @@ export class UserForm extends React.Component<IProps, IState> {
       </FormGroup>
     );
 
+    const readonlyAuth = () => (
+      <FormGroup
+        fieldId='auth_provider'
+        key='readonlyAuth'
+        label={t`Authentication provider`}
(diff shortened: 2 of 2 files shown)


src/containers/user-management/user-create.tsx
```diff
@@ -30,6 +30,7 @@ class UserCreate extends React.Component<RouteComponentProps, IState> {
         last_name: '',
         email: '',
         password: '',
+        auth_provider: [],
```
himdel, line 33: this means we're now submitting `auth_provider`, a read-only field, when creating/editing users.

It seems to work, but is it guaranteed to work? Should we be keeping this outside the user state?

ShaiahWren: When I remove it from user state I get this error message:
Screen Shot 2021-10-06 at 2 04 37 PM
Where else should I put it?

himdel: you did the API, right? If the API explicitly filters this out when editing, we can keep this as is. But if not, it can't be left inside `user`.

himdel: if we go with the second option, the type of this `user` in state just can't be `UserType` and it should work :)

himdel (approved):

awcrosby (approved):
tested with keycloak, LGTM

ShaiahWren:
Pr closed.

PR #1053, opened 2021-10-07
standalone-loader, routes: fix keycloak redirect by setting featureFlags even if loading user/settings fails

himdel:
Fixes [AAH-990](https://issues.redhat.com/browse/AAH-990)

When using keycloak (`COMPOSE_PROFILE=standalone-keycloak`) (introduced in #809),
clicking the logout button, or navigating to a screen when not logged in should redirect to the keycloak login screen.

That relies on logic in `AuthHandler#render` (`src/loaders/standalone/routes.tsx`)

```
    if (!user && !noAuth) {
      if (isExternalAuth && UI_EXTERNAL_LOGIN_URI) {
        return (redirect to keycloak);
      }
      return (redirect to our login);
```

which broke after #900: https://github.com/ansible/ansible-hub-ui/pull/900/files#diff-76c5d9078cc60d85b811f04071a06094c28652a2577143e6aed4da35dba1cd50L88-R105

the change caused `featureFlags` not to get set if the user or settings API request fails, which happens when the user is not logged in (and view-only is disabled),
which caused us to redirect to the wrong login screen.

Updated `standalone-router` `updateInitialData` to make it easier to set these independently,
and made sure `featureFlags` gets set even when other requests fail.

---

When using view-only + keycloak, the redirect still won't happen (user is not null), maybe we should consider adding a "go to login" (the right one) button to the `EmptyStateUnauthorized` screen.

Cc @awcrosby @ZitaNemeckova


ZitaNemeckova:
LGTM code-wise :+1: 

@awcrosby Can you please verify it's working in UI.

himdel:
Due to popular demand, added the Login button to `EmptyStateUnauthorized`,
unified all the Login buttons to use `LoginLink`,
and made sure `LoginLink` respects SSO.

This means SSO in view-only mode can still reach the SSO just by clicking the Login link/button

PR #1084, opened 2021-10-13
Add test for manual Collection upload

ZitaNemeckova:
Fixes https://issues.redhat.com/browse/AAH-1011


test/cypress/integration/collection_upload.js
```diff
@@ -0,0 +1,31 @@
...
+    cy.galaxykit('-i namespace create', 'ansible');
+    cy.menuGo('Collections > Namespaces');
+    cy.intercept('GET', Cypress.env('prefix') + '_ui/v1/repo/published/*').as(
+      'namespaces',
+    );
+
+    cy.get('a[href="/ui/repo/published/ansible"]').click();
+    cy.wait('@namespaces');
+    cy.contains('Upload collection').click();
+    cy.get('input[type="file"]').attachFile(filepath);
```
himdel, line 24:
```suggestion
    // test/cypress/fixtures/collections/ansible-network-1.2.0.tar.gz
    cy.fixture('collections/ansible-network-1.2.0.tar.gz', 'binary')
      .then(Cypress.Blob.binaryStringToBlob)
      .then(fileContent => {
        cy.get('input[type="file"]').attachFile({
          fileContent,
          fileName: 'ansible-network-1.2.0.tar.gz',
          mimeType: 'application/gzip',
        });
      });
```

(as per https://www.npmjs.com/package/cypress-file-upload#working-with-raw-file-contents)

himdel: (this works, but sometimes waiting for "Completed" fails because the task takes more than 4 seconds)

package.json
```diff
@@ -81,8 +81,8 @@
         "lint:ls": "ls-lint",
         "lint:sass": "stylelint 'src/**/*.scss' --config .stylelintrc.json",
         "lint:ts": "tsc",
-        "prettier": "prettier --write 'src/**' 'config/**' 'test/**'",
-        "prettier:check": "prettier -l 'src/**' 'config/**' 'test/**'",
+        "prettier": "prettier --write 'src/**' 'config/**' 'test/cypress/integration/**' 'test/cypress/support/**'",
```
himdel, line 84: can we add `test/cypress/fixtures/` (or `*.tar.gz`) to `.prettierignore` instead? (or, if not, there's also `test/cypress/plugins`)

himdel (approved):
(Apart from the prettier issue above..)

LGTM :+1: 

We may want to consider adding a `before` that delete namespaces and collections .. but the test doesn't seem to need it, and collection deletion is being handled by other PRs, so I'm fine with postponing that until we're merging the collection tests into one :).

PR #1115, opened 2021-10-20
Add missing handling API errors on UI 

jerabekjiri:
Issue: [AAH-517](https://issues.redhat.com/browse/AAH-517)

- A few Examples of error alerts:
(ignore the alert description "Request failed with status code 403", just for testing purposes :)) 
Screenshot from 2021-10-18 13-50-03
Screenshot from 2021-10-18 14-42-48
Screenshot from 2021-10-14 19-44-25
Screenshot from 2021-10-17 15-43-30

- show success alert after editing namespace (after the redirect to `namespace-detail`). This wasn't possible before, because of local state `alerts`  and local component `AlertList`
Screenshot from 2021-10-17 19-52-08

I was thinking maybe it would be a better idea to move from local alerts in every component to push alerts to global state (context) and display it from one place always? in this case I added it to `App` component. As I mentioned before, this would make it easier to show alerts after changing URL (redirecting). What do you think about this?  @himdel @ZitaNemeckova 

Also if there's an API error in fetching table list, currently I just show an empty state and display error alert. I guess this might be confusing for users, this would be probably solved by the `500 error screen`, so maybe I should do follow-up PR for that (since we already have UI mockups for it) and later add `500 error screen` to all table list responses?    
Screenshot from 2021-10-17 19-21-01

### DIFF
--- a/src/components/collection-detail/collection-info.tsx
+++ b/src/components/collection-detail/collection-info.tsx
@@ -17,7 +17,14 @@ import {
 import { DownloadIcon } from '@patternfly/react-icons';
 
 import { CollectionDetailType, CollectionAPI } from 'src/api';
-import { Tag, ClipboardCopy, LoginLink } from 'src/components';
+import {
+  Tag,
+  ClipboardCopy,
+  LoginLink,
+  AlertType,
+  closeAlertMixin,
+  AlertList,
+} from 'src/components';
 import { Paths, formatPath } from 'src/paths';
 import { AppContext } from 'src/loaders/app-context';
 import { userLanguage } from 'src/l10n';
@@ -29,13 +36,21 @@ interface IProps extends CollectionDetailType {
   updateParams: (params) => void;
 }
 
-export class CollectionInfo extends React.Component<IProps> {
+interface IState {
+  alerts: AlertType[];
+}
+
+export class CollectionInfo extends React.Component<IProps, IState> {
   downloadLinkRef: any;
   static contextType = AppContext;
 
   constructor(props) {
     super(props);
     this.downloadLinkRef = React.createRef();
+
+    this.state = {
+      alerts: [],
+    };
   }
 
   render() {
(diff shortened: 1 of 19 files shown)


src/loaders/standalone/standalone-loader.tsx
```diff
@@ -308,6 +327,10 @@ class App extends React.Component<RouteComponentProps, IState> {
 
     return this.ctx(
       <Page isManagedSidebar={true} header={Header} sidebar={Sidebar}>
+        <AlertList
```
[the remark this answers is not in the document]
jerabekjiri: Good idea :)

src/components/execution-environment/repository-form.tsx
```diff
@@ -80,16 +82,25 @@ export class RepositoryForm extends React.Component<IProps, IState> {
...
+            });
+          }
+        })
+        .catch((e) => {
+          this.props.addAlert(
+            'danger',
+            t`Error loading registries.`,
+            e?.message,
+          );
+          this.props.onCancel();
```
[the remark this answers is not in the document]
jerabekjiri: >this would hide the tab on error, right?

yes 

>I'd say we should never do that in modals.. but then again this is when fit fails to load, so.. I guess it's ok :)

it's a bit clumsy, but we don't have any error state for `fails to load list` right now, and showing an empty state is not a good idea, also we can't let user proceed if there's a fail loading a data, so I was thinking this was the only option for now.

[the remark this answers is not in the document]
jerabekjiri: Thinking about it now, this might be better solution in `RepositoryForm` component and disabling the save button, so user can't finish current action?
Screenshot from 2021-11-01 13-27-38

jerabekjiri:
The question is should we let the user finish the action (create a new namespace for example) even if this field isn't mandatory (group field is optional, therefore possibly he could create a new namespace and set this later in the app)? or never let user finish the action and keep the `create` button disabled if an error occurred? @himdel @ZitaNemeckova 
Screenshot from 2021-11-05 11-36-51

himdel:
@jerabekjiri I think we can keep this simple ... the error itself shouldn't stop the user from finishing actions.

If the field is required, the group will never be selected, so the user will not be able to save anyway.

jerabekjiri:
>If the field is required, the group will never be selected, so the user will not be able to save anyway.

makes sense, fixed :)

src/components/collection-detail/collection-info.tsx
```diff
@@ -29,13 +36,21 @@ interface IProps extends CollectionDetailType {
...
+
+export class CollectionInfo extends React.Component<IProps, IState> {
   downloadLinkRef: any;
   static contextType = AppContext;
 
   constructor(props) {
     super(props);
     this.downloadLinkRef = React.createRef();
+
+    this.state = {
```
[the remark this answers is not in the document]
himdel: @jerabekjiri this is still relevant.. `CollectionInfo` `state.alerts` is no longer used, so no need to add state, right?

src/components/namespace-modal/namespace-modal.tsx
```diff
@@ -143,12 +150,33 @@ export class NamespaceModal extends React.Component<IProps, IState> {
...
+                ]}
+                groups={newGroups}
+                setGroups={(g) => this.setState({ newGroups: g })}
+                menuAppendTo='parent'
+                onError={(err) =>
+                  this.setState({
+                    formErrors: {
+                      ...this.state.formErrors,
+                      groups: {
+                        title: 'Error loading groups.',
```
himdel, line 171: l10n

src/components/namespace-form/namespace-form.tsx
```diff
@@ -83,16 +101,33 @@ export class NamespaceForm extends React.Component<IProps, IState> {
...
+                const newNS = { ...namespace };
+                newNS.groups = g;
+                this.props.updateNamespace(newNS);
+              }}
+              onError={(err) =>
+                this.setState({
+                  formErrors: {
+                    ...this.state.formErrors,
+                    groups: {
+                      title: 'Error loading groups.',
```
himdel, line 122: l10n

src/components/execution-environment/repository-form.tsx
```diff
@@ -361,12 +424,25 @@ export class RepositoryForm extends React.Component<IProps, IState> {
...
+          selectedGroups: cloneDeep(result.data.groups),
+          originalSelectedGroups: result.data.groups,
+        }),
+      )
+      .catch((e) => {
+        this.setState({
+          formErrors: {
+            ...this.state.formErrors,
+            groups: {
+              title: 'Error loading groups.',
```
himdel, line 439: l10n

src/components/execution-environment/repository-form.tsx
```diff
@@ -328,21 +369,43 @@ export class RepositoryForm extends React.Component<IProps, IState> {
...
+                    !this.props.permissions.includes(
+                      'container.change_containernamespace',
+                    )
+                  }
+                  onError={(err) =>
+                    this.setState({
+                      formErrors: {
+                        ...this.state.formErrors,
+                        groups: {
+                          title: 'Error loading groups.',
```
himdel, line 399: l10n

src/components/execution-environment/repository-form.tsx
```diff
@@ -75,21 +85,39 @@ export class RepositoryForm extends React.Component<IProps, IState> {
...
+        })
+        .catch((e) => {
           this.setState({
-            registrySelection: this.state.registries.filter(
-              ({ id }) => id === this.props.registry,
-            ),
+            formErrors: {
+              ...this.state.formErrors,
+              registries: {
+                title: 'Error loading registries.',
```
himdel, line 114: l10n

src/containers/namespace-list/namespace-list.tsx
```diff
@@ -73,6 +75,7 @@ export class NamespaceList extends React.Component<IProps, IState> {
       hasPermission: true,
       isModalOpen: false,
       loading: true,
+      alerts: [],
```
himdel, line 78: looks like `state.alerts` is never used here?

src/components/user-form/user-form.tsx
```diff
@@ -56,13 +60,20 @@ export class UserForm extends React.Component<IProps, IState> {
...
+      },
     };
   }
 
   componentDidMount() {
     this.loadGroups('');
   }
 
+  componentWillUnmount() {
+    this.context.setAlerts([]);
```
himdel, line 74: I'm not sure about this one.. user-list/detail doesn't seem to use context.alerts at all, why would user-form?

src/components/user-form/user-form.tsx
```diff
@@ -271,9 +288,30 @@ export class UserForm extends React.Component<IProps, IState> {
...
+              ...this.state.formErrors,
+              groups: {
+                variant: 'danger',
+                title: t`Error loading groups.`,
+                description: e?.message,
+              },
+            },
+          });
+        },
+        /*this.context.setAlerts([
```
himdel, line 306: (commented out code)

src/containers/edit-namespace/edit-namespace.tsx
```diff
@@ -69,10 +73,34 @@ class EditNamespace extends React.Component<RouteComponentProps, IState> {
...
+                namespace: this.props.match.params['namespace'],
+                repo: this.context.selectedRepo,
+              }),
+            },
+            () => {
+              this.context.setAlerts([
+                ...this.context.alerts,
+                {
+                  variant: 'danger',
+                  title: `Error loading active user.`,
```
himdel, line 97: l10n

src/components/user-form/user-form.tsx
jerabekjiri: you're right, I forgot to remove it, thanks :)

jerabekjiri:
@himdel thanks for pointing out mistakes, I completely forgot about l10n here, sorry about that

himdel:
No worries, looks like it's all fixed now, not seeing any new issues :+1:  :)

himdel (approved):

jerabekjiri:
@himdel Resolved conflicts, could you review one more before merging, please? :)

himdel:
it *looks* fine, but cypress disagrees...

oh, and https://github.com/ansible/ansible-hub-ui/pull/1115/files#r749877345 is still relevant, I think? :)

jerabekjiri:
I must have messed it up during the conflict resolving, should be fixed and updated

awcrosby:

jerabekjiri:
Added `Issue: AAH-517` to the latest commit description

ZitaNemeckova (approved):
LGTM 👍

himdel:
backporting https://github.com/ansible/ansible-hub-ui/pull/1220 depends on this one, backporting as well

hendersonreed:
@himdel I'm trying to move this to Done in Jira, and was interested in your thoughts about adding tests for error handling down the line. I think we can stub 500 network responses using https://docs.cypress.io/guides/guides/network-requests#Stub-Responses and then verify that errors appear as they ought to.

I don't think it's something that we need to necessarily add to this in order to consider this work `done` but wanted to mention it here and get your thoughts on it.

himdel:
@hendersonreed Great idea, I agree we should be testing these as well, and stubbing the responses sounds like a perfect solution. (I guess if we should also test a couple of real failure states where we can, but stubbing allows us to test everything.)

(I've at least added a link from https://github.com/himdel/ansible-hub-ui/issues/1 for now.)

PR #1137, opened 2021-10-26
Delete entire collection cypress test

jerabekjiri:
Issue: [AAH-1007](https://issues.redhat.com/browse/AAH-1007)

Inspired from @ShaiahWren's PR for collection version deletion [#1117](https://github.com/ansible/ansible-hub-ui/pull/1117).

I added a recursive function to wait for the task to finish deleting collection (`state='completed'`).

PR #1141, opened 2021-10-27
tests: edit local and remote ee

ShaiahWren:
Issue: https://issues.redhat.com/browse/AAH-1002

Cypress test for editing local and remote EEs. 

Note: the test `it('edits a local container)` is not correct... I wrote it on the assumption that it would appear in the UI, but now I know that it doesn't...

### DIFF
--- a/test/cypress/integration/execution_environments_edit.js
+++ b/test/cypress/integration/execution_environments_edit.js
@@ -3,8 +3,57 @@ describe('execution environments', () => {
   let adminPassword = Cypress.env('password');
   let num = (~~(Math.random() * 1000000)).toString();
 
+  function deleteRegistries() {
+    cy.intercept(
+      'GET',
+      Cypress.env('prefix') + '_ui/v1/execution-environments/registries/?*',
+    ).as('registries');
+
+    cy.visit('/ui/registries');
+
+    cy.wait('@registries').then((result) => {
+      var data = result.response.body.data;
+      data.forEach((element) => {
+        cy.get(
+          'tr[aria-labelledby="' +
+            element.name +
+            '"] button[aria-label="Actions"]',
+        ).click();
+        cy.contains('a', 'Delete').click();
+        cy.contains('button', 'Delete').click();
+      });
+    });
+  }
+
+  function deleteContainers() {
+    cy.intercept(
+      'GET',
+      Cypress.env('prefix') + '_ui/v1/execution-environments/repositories/?*',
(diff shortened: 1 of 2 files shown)


himdel:
The `addLocalContainer` issue should be resolved by https://github.com/ansible/ansible-hub-ui/pull/1142 .. you'll need to adjust `cypress.env.json` accordingly for things to (hopefully) start working locally :)

test/cypress/integration/execution_environments_edit.js
```diff
@@ -0,0 +1,46 @@
+describe('execution environments', () => {
+  let adminUsername = Cypress.env('username');
+  let adminPassword = Cypress.env('password');
+  let num = (~~(Math.random() * 1000000)).toString();
```
himdel, line 4: Note that this was a suggestion to get unstuck with container errors, it shouldn't be in the final test.

`before` should start by removing all existing containers and registries,
making sure to create that one docker registry,
adding a local and remote container,

and then you're working with a known state which only has the 2 containers you need :).

(you *can* keep `num` if you like, but even with num, without deletion this only works 5 times, then the new containers start appearing on page 2, and test fail because they don't find them)

ShaiahWren:
@himdel , am I correct that we are waiting on #1142 before cypress will pass on this pr?

ZitaNemeckova:
@ShaiahWren yes. If #1142 works for you locally please review/approve that PR.

test/cypress/integration/execution_environments_edit.js
```diff
@@ -0,0 +1,95 @@
...
+    cy.login(adminUsername, adminPassword);
+    cy.menuGo('Execution Environments > Execution Environments');
+  });
+
+  it('edits a remote container', () => {
+    cy.contains('a', `remotepine${num}`).click();
+    cy.get('.pf-c-button.pf-m-primary').contains('Edit').click();
+    cy.get('#description').type('This is the description.');
+    cy.contains('button', 'Save').click();
+    cy.wait(10000); // FIXME have a reload request, wait for it; can't wait for an unspecified number of task requests
```
himdel, line 77: we can fix this one later, https://github.com/ansible/ansible-hub-ui/pull/1137 adds a `waitForTaskToFinish` that could be useful for this one

himdel (approved):
LGTM, works for me :+1: 

Nice job on the delete functions, we can move them to helpers and use elsewhere as well :)

PR #1149, opened 2021-10-30
test/README: add checklists

himdel:
Fixes AAH-1004

This tries to set up checklist for list views, detail views, shared actions, and new/edit forms.


himdel:
Readt for review, the actual change is https://github.com/ansible/ansible-hub-ui/pull/1149/commits/c026b980cd519c034d37674544e40b668c188fea?short_path=5de36ac#diff-5de36acd90308dc62abf7855a686ee7052ffb6e762c756fd735fb0c9fbd9595d :)

test/README.md
```diff
@@ -1,139 +1,198 @@
...
+`beforeEach`:
+  * use for `cy.login` + `cy.visit('/')` if all tests use the same login, then individual tests can start with `cy.menuGo`
+
+`it`:
+  * start with `cy.login` (only) if necessary
+  * do `cy.visit` or `cy.menuGo` to ensure the right place
+
+`it` + UI helpers:
+  * always make sure to wait for the last thing that happens automatically when initiating a user action
+  * for example, when adding a user, waiting for the API response is not enough, since the UI then redirects to the user list screen => wait for the Users list screen to finish loading
```
MilanPospisil, line 98: I would like to also add one thing - when you are waiting for API response, that may not be enough, because items must be then rendered and this may cause sporadic failure in situation when you are looking for something that should not exist (for example using contains). Because cypress may (or may not, depends on randomized timing) run before rendering and it then sees that element dont exist, however element may appear after rendering. When you are testing that element do exist, this problem does not appear, because if the contains is not valid (element is not rendered yet), cypres test it several times after several seconds until the element is rendered (to ensure page is in valid state). In this situation, test may fail only if rendering last more than several seconds, which will happen only in rare occasion. So to ensure correct and stable tests, you should use some contains for waiting for some element to be rendered (so it will wait for page render) before you do some test for 'should not exist' element. For example when testing elements in the list, first test the element that should be in the list, then element that should not be in the list.

test/README.md
```diff
@@ -1,139 +1,198 @@
...
+
+`after` / `afterEach`:
```
MilanPospisil, line 100: Im using this for cleaning after tests, but this will also slow down testing and yes, we should not rely on it. Cleaning the DB in before is crucial. Cleaning after is nice thing to have because of maintaining clean database, but at the expense of slower tests. What do you thing we should use?

test/README.md
```diff
@@ -1,139 +1,198 @@
...
+* check title, columns, sanity check data
+* check filter, empty state filter
+* check perpage and paging (can set to 2)
+* check sort, do it from page 2, should reset page to 1
+* check screen actions, list item actions
+  * skip create/edit, those should live in a form screen test
+  * prefer to test nontrivial extra actions (including delete) in a separate per-feature test .. which can test both list/detail screen for that button
+  * test the rest
+* [NOT YET] switch to a user with only the relevant permissions per test, not admin (needs permission-aware helpers first) .. only need the Foo button in foo action test, can test not in list screen
+
```
MilanPospisil, line 130: I would also like to add that you dont need to test all the items in the list, only some items and some missing items (contains.should.not.exist) is enough. Additionaly, this seems that our current test does not meet those requirements and I suggest we should add some common functionality if it is possible.

test/README.md
himdel: Tthe readme already says just waiting for the API is not enough, so this sounds like it's solved?

Or is there something I should change here?

himdel: (Also, this is not coming from this PR.)

test/README.md
himdel: This part is also not a change in this PR.

But yeah, as you (and indeed the guidelines) say, it's a nice to have, not something we should rely on.

test/README.md
himdel: I don't think any part suggest anybody should test all the items, we're testing features, not components here :+1: 

> current tests

Yeah, that's fine, this is for new tests.

> common functionality

Out of scope, but yes, shared helpers are always nice :)

test/README.md
MilanPospisil: I just wanted to point one problem that may be dangerous, when you "wait" for the render using contains(something).should.not.exists, this will not actualy wait for render, because it will not see the element in the current page state and return true immideately, then element appear in render. Whereas positive contains always wait several seconds if it not sees the element until element eventualy appear in render.

himdel: good point, adding a note about `should.not.exist` specifically :+1:

himdel:
I'll just merge this, it's a README change only, we can always update these later.

PR #1150, opened 2021-10-30
console-logger: log all console.log args, not only the first one

himdel:
When CONSOLE_LOG_TO_TERMINAL=true is used, `console.log('foo', 'bar')` will only log `foo` to the console.
Making sure all args are stringified.

(Or, we could switch to https://github.com/flotwig/cypress-log-to-output which seems related, and uses `JSON.stringify` on args.)


ZitaNemeckova (approved):
LGTM 👍

PR #1151, opened 2021-10-30
EE delete modal: don't hide name while still waiting for task

himdel:
the modal shows the name of the EE to remove, but only until we reset `selectedItem` (where we're taking the name from), which was happening before `waitForTask` is done, making the name disappear mid-delete.

Unset `selectedItem` only when closing the modal instead.


ZitaNemeckova (approved):
LGTM 👍

himdel:
Fixes https://issues.redhat.com/browse/AAH-1182

PR #1217, opened 2021-11-08
Fix create namespace problem

MilanPospisil:
When creating namespace, there is search dropdown list. Before this PR, it worked this way:

In dropdown list, there is list of available groups. When one group is selected, it appears in bottom of the drop down in selected group permission area and it dissapears from the search drop down. This is fine. But when the group is removed, it is not available in search drop down again. This PR fixed this by adding the removed group back to drop down.

No Issue.

### DIFF
--- a/src/components/permissions/obect-permission-field.tsx
+++ b/src/components/permissions/obect-permission-field.tsx
@@ -88,6 +88,7 @@ export class ObjectPermissionField extends React.Component<IProps, IState> {
         newGroups.push(g);
       }
     }
+
     this.props.setGroups(newGroups);
 
     const newSearchGroups = [...this.state.searchGroups];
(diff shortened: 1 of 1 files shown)


src/components/permissions/obect-permission-field.tsx
```diff
@@ -90,6 +90,10 @@ export class ObjectPermissionField extends React.Component<IProps, IState> {
     }
 
     this.props.setGroups(newGroups);
+
+    const newSearchGroups = [...this.state.searchGroups];
```
himdel, line 94:
```suggestion
    const newSearchGroups = [...this.state.searchGroups,group];
```

himdel:
Looks good, is there an associated issue, or simply fixing a bug you've found?

himdel (approved):

PR #1219, opened 2021-11-09
Fix create namespace modal

MilanPospisil:
From Issue AAH-824:

- [x]  Resize modal to standard (shorter width)

- [x] Remove "please, provide a namespace" text under "name" field 

- [x]  Move popover question mark (input group) to form (next to field label) 

- [x]  Reword "name" field alert message from "Name can only contain [A-Za-z0-9_]" to "Name can only contain letters and numbers" 

- [x] Remove popover extra right padding 

- [x] Change "namespace owners" select prompt from "find a group" to "select a group", let selected groups appear in field as chips 

- [x] Add "Select namespace owner permissions" title above permissions section

Completed: 1, 2, 4, 6 (partialy), 7

I was not sure what 3, 5 and part of 6 ment to be, maybe it is already corrected.

### DIFF
--- a/src/components/namespace-modal/namespace-modal.tsx
+++ b/src/components/namespace-modal/namespace-modal.tsx
@@ -118,6 +118,7 @@ export class NamespaceModal extends React.Component<IProps, IState> {
             labelIcon={
               <HelperText
                 content={t`Namespace names are limited to alphanumeric characters and underscores, must have a minimum length of 2 characters and cannot start with an ‘_’.`}
+                header={t`Namespace name`}
               />
             }
           >
--- a/src/components/namespace-modal/namespace-modal.tsx
+++ b/src/components/namespace-modal/namespace-modal.tsx
@@ -147,6 +147,7 @@ export class NamespaceModal extends React.Component<IProps, IState> {
               groups={newGroups}
               setGroups={(g) => this.setState({ newGroups: g })}
               menuAppendTo='parent'
+              selectPermissionCaption={t`Select namespace owner permissions`}
             />
           </FormGroup>
         </Form>
--- a/src/components/namespace-modal/namespace-modal.tsx
+++ b/src/components/namespace-modal/namespace-modal.tsx
@@ -43,7 +43,7 @@ export class NamespaceModal extends React.Component<IProps, IState> {
     if (name == '') {
       error['name'] = t`Please, provide the namespace name`;
     } else if (!/^[a-zA-Z0-9_]+$/.test(name)) {
-      error['name'] = t`Name can only contain [A-Za-z0-9_]`;
+      error['name'] = t`Name can only contain letters and numbers`;
     } else if (name.length <= 2) {
       error['name'] = t`Name must be longer than 2 characters`;
     } else if (name.startsWith('_')) {
@@ -90,7 +90,7 @@ export class NamespaceModal extends React.Component<IProps, IState> {
     const { newNamespaceName, newGroups, newNamespaceNameValid } = this.state;
     return (
       <Modal
-        variant='large'
+        variant='small'
         title={t`Create a new namespace`}
         isOpen={this.props.isOpen}
         onClose={this.toggleModal}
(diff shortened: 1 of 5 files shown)


src/components/permissions/permission-chip-selector.tsx
```diff
@@ -30,10 +31,15 @@ export class PermissionChipSelector extends React.Component<IProps, IState> {
   }
 
   render() {
+    let selectPermissionCaption = this.props.selectPermissionCaption;
```
himdel, line 34:
```suggestion
    let selectPermissionCaption = this.props.selectPermissionCaption || t`Select permissions`;
```

himdel:
the test failures look relevant, probably want to update the tests :)

src/components/permissions/permission-chip-selector.tsx
MilanPospisil: Good idea.

himdel:
* [x] Resize modal to standard (shorter width)

|before|after|
|-|-|
|  |  |

* [ ] Remove "please, provide a namespace" text under "name" field
* [ ] Move popover question mark (input group) to form (next to field label)
* [ ] Reword "name" field alert message from "Name can only contain [A-Za-z0-9_]" to "Name can only contain letters and numbers"
* [ ] Remove popover extra right padding
* [ ] Change "namespace owners" select prompt from "find a group" to "select a group", let selected groups appear in field as chips
* [ ] Add "Select namespace owner permissions" title above permissions section

src/components/namespace-modal/namespace-modal.tsx
```diff
@@ -39,11 +39,8 @@ export class NamespaceModal extends React.Component<IProps, IState> {
   private newNamespaceNameIsValid() {
     const error: any = this.state.errorMessages;
     const name: string = this.state.newNamespaceName;
-
-    if (name == '') {
-      error['name'] = t`Please, provide the namespace name`;
```
himdel, line 44: I think this one needs to stay (and only the static helperText should be removed)

Otherwise typing something in and then removing it gives you `Name can only contain letters and numbers`

himdel:
Unless @ZitaNemeckova disagress, I'd just remove the `selectPermissionCaption` changes, as those don't seem to fit the description, and we can merge the rest, and address the unchecked point at the next UX meeting

himdel:
So, @MilanPospisil I talked with @sbuenafe-rh today, and we have couple of changes....

* the modal width should go to medium, not small (mostly because the permission select needs the space)
* "let selected groups appear in field as chips" is obsolete (from before you could set permissions for groups)
* "title above permissions section" - Susan will get back to us on this one, but the idea was ta add a form label before the groups

MilanPospisil:
Remove padding in helper text now looks like this. But this change is for all helpers, is it that you wanted or should I go with some props options in helper component?
Screenshot_20211119_112655

test/cypress/integration/namespace_form.js
```diff
@@ -32,20 +32,28 @@ describe('A namespace form', () => {
   });
 
   it('should give message before typing', () => {
```
himdel, line 34: the test name no longer reflects reality, looks like we should go with `should give message if input is empty`,
because the next test should be `should give message if input has incorrect characters` and those two are testing the same thing :)

src/components/namespace-modal/namespace-modal.tsx
```diff
@@ -143,6 +142,7 @@ export class NamespaceModal extends React.Component<IProps, IState> {
             fieldId='groups'
             helperTextInvalid={this.state.errorMessages['groups']}
           >
+            {t`Select namespace owner permissions`}
```
himdel, line 145: this no longer belongs here, we're not  adding any title (https://github.com/ansible/ansible-hub-ui/pull/1219#issuecomment-970382928)

himdel:
> Remove padding in helper text now looks like this. But this change is for all helpers, is it that you wanted or should I go with some props options in helper component?

Yeah, for all helpers is good :+1: 

But it doesn't *look* that good yet, I'd still suggest making that a separate PR :)

The text is now too close to the close button vertically, I think we'd need to push the close button closer to the top right corner (and the same distance from each), and keep at least the same padding between the button and the text.

MilanPospisil:
So I removed the title and fixed the layout in helper. I hope it is all, Im starting to get lost in so much issues in one PR :-) Maybe I should divide this kind of problems in two PRs in future.

himdel (approved):
LGTM, I agee all the points are fixed now :+1: 

And we can talk about the proper way to solve all the other helper text paddings on the next UX meeting, the header you added works correctly here :)

PR #1220, opened 2021-11-09
Add check that EE name is correct when typed in

ZitaNemeckova:
Fixes https://issues.redhat.com/browse/AAH-968

Steps to reproduce:

- go to http://localhost:8002/ui/containers
- click "Add execution environment" button
- set the name to something invalid like "i/am/an/invalid/container/name"
- click save

(Screenshots are in French because I'm looking for possible problems with translation)
Before:
Screenshot 2021-11-09 at 12 23 36

After:
Screenshot 2021-11-09 at 12 22 11

Screenshot 2021-11-11 at 11 39 11

This is not fixed in https://github.com/ansible/ansible-hub-ui/pull/1115 and is not related to that PR. 

@himdel please have a look, thanks :)

### DIFF
--- a/src/components/execution-environment/repository-form.tsx
+++ b/src/components/execution-environment/repository-form.tsx
@@ -46,8 +46,8 @@ interface IProps {
   registry?: string; // pk
   upstreamName?: string;
   remotePulpId?: string;
-
   addAlert?: (variant, title, description?) => void;
+  formError: { title: string; detail: string }[];
 }
 
 interface IState {
@@ -125,7 +125,8 @@ export class RepositoryForm extends React.Component<IProps, IState> {
   }
 
   render() {
-    const { onSave, onCancel, namespace, isNew, isRemote } = this.props;
+    const { onSave, onCancel, namespace, isNew, isRemote, formError } =
+      this.props;
     const {
       name,
       description,
@@ -151,13 +152,8 @@ export class RepositoryForm extends React.Component<IProps, IState> {
           <Button
             key='save'
             variant='primary'
+            isDisabled={!this.formIsValid()}
             onClick={() => onSave(this.onSave())}
-            isDisabled={
-              (this.state.name.length === 0 ||
-                this.state.upstreamName.length === 0 ||
-                this.state.registrySelection.length === 0) &&
-              isRemote
-            }
           >
             {t`Save`}
           </Button>,
@@ -198,12 +194,15 @@ export class RepositoryForm extends React.Component<IProps, IState> {
                 key='name'
                 fieldId='name'
                 label={t`Name`}
+                helperTextInvalid={t`Container names can only contain alphanumeric characters, ".", "_", "-" and a up to one "/".`}
+                validated={this.validateName(name)}
               >
                 <TextInput
                   id='name'
                   value={name}
                   isDisabled={!isNew}
                   onChange={(value) => this.setState({ name: value })}
+                  validated={this.validateName(name)}
                 />
               </FormGroup>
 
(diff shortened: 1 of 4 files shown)


src/components/execution-environment/repository-form.tsx
```diff
@@ -349,6 +357,22 @@ export class RepositoryForm extends React.Component<IProps, IState> {
     );
   }
 
+  private validateName(name) {
+    const regex = new RegExp('^[0-9A-Za-z._-]*/?[0-9A-Za-z._-]*$');
```
himdel, line 361:
```suggestion
    const regex = /^([0-9A-Za-z._-]+\/)?[0-9A-Za-z._-]+$/;
```

(javascript has regex literals, so no need to use a string; and wrapping the part up to and including the slash lets us have the whole bit optional and then we can use `+` to require at least one character before/after the slash (prevents `/foo` or `foo/` or `/` from being valid))

src/components/execution-environment/repository-form.tsx
```diff
@@ -349,6 +357,22 @@ export class RepositoryForm extends React.Component<IProps, IState> {
...
+    if (regex.test(name)) {
+      return 'default';
+    } else {
+      return 'error';
+    }
+  }
+
+  private formIsValid() {
+    return (
+      this.props.isRemote ||
```
himdel, line 371: `isRemote` -> `isNew` (both remote and local have names, but only new items have editable names)

src/containers/execution-environment-list/execution_environment_list.tsx
```diff
@@ -450,12 +450,19 @@ class ExecutionEnvironmentList extends React.Component<
...
+                      ? t`Execution environment added.`
+                      : t`Execution environment saved.`,
+                  }),
                 },
                 () => this.queryEnvironments(),
               );
             })
             .catch(() => {
               this.setState({
+                showRemoteModal: false,
```
himdel, line 465: this is problematic

Hiding the modal on the first error means we don't allow the user to change what was wrong and save again.

Maybe you want to introduce inline alerts to the modal, instead of showing errors on the parent screen, hidden by the modal?

himdel:
(test failures seem relevant...)

src/components/execution-environment/repository-form.tsx
```diff
@@ -343,12 +353,34 @@ export class RepositoryForm extends React.Component<IProps, IState> {
...
+      return 'default';
+    } else {
+      return 'error';
+    }
+  }
+
+  private formIsValid() {
+    return (
+      !this.props.isNew ||
+      (this.validateName(this.state.name) && !!this.state.name)
```
himdel, line 380: should this be..

```suggestion
      (this.state.name && this.validateName(this.state.name))
```

to prevent running validation on empty names?
Or it could be just

```suggestion
      this.validateName(this.state.name)
```

since the regex will return false on null and empty string?

src/components/execution-environment/repository-form.tsx
```diff
@@ -354,12 +354,39 @@ export class RepositoryForm extends React.Component<IProps, IState> {
...
+  }
+
+  private formIsValid() {
+    return (
+      !(
+        this.props.isRemote &&
+        this.state.name.length === 0 &&
+        this.state.upstreamName.length === 0 &&
+        this.state.registrySelection.length === 0
+      ) && this.validateName(this.state.name)
```
himdel, line 386: 2 issues...

* `validateName` returns `'default' | 'error'` .. both are true, so it always returns true
* what actually changed (after resolving the double `!()`) is...

```diff
-isRemote && (name.length === 0 || upstreamName.length === 0 || registrySelection.length === 0)
+isRemote && name.length === 0 && upstreamName.length === 0 && registrySelection.length === 0
```

is that intentional?

Also, can we make it more readable? Maybe something like..

```js
if (!isRemote) {
  // no validation for local
  return true;
}

const nameValid = name && this.validateName(name) === 'default';
return nameValid && upstreamName && registrySelection.length;
```

awcrosby:

src/components/execution-environment/repository-form.tsx
```diff
@@ -169,12 +166,15 @@ export class RepositoryForm extends React.Component<IProps, IState> {
                 key='name'
                 fieldId='name'
                 label={t`Name`}
+                helperTextInvalid={t`Container names can only contain alphanumeric numbers, ".", "_", "-" and a up to one "/".`}
```
himdel, line 169:
```suggestion
                helperTextInvalid={t`Container names can only contain alphanumeric characters, ".", "_", "-" and a up to one "/".`}
```

src/components/execution-environment/repository-form.tsx
```diff
@@ -354,12 +354,38 @@ export class RepositoryForm extends React.Component<IProps, IState> {
                 )
               }
             ></ObjectPermissionField>
+            {!!formError &&
```
himdel, line 357: Can we move this outside the FormGroup?

That way, there will be spacing between "Groups with access" and the alerts..

src/containers/execution-environment-list/execution_environment_list.tsx
```diff
@@ -67,6 +67,7 @@ interface IState {
   confirmDelete: boolean;
   isDeletionPending: boolean;
   inputText: string;
+  formError: { title: string; detail: string }[];
```
himdel, line 70: this is missing a corresponding change in `src/containers/execution-environment-detail/base.tsx` for the detail screen
(all the changes in this file I mean)

right now it just hides the modal on error and pops up a generic alert

(unless that's out of scope?)

himdel:
Needs a rebase, but LGTM :)
(I think the conflict was just about removing `!!`)

ZitaNemeckova:
@himdel rebased and ready for review :)

himdel:
Tested in the UI,  looks like everything works, just one more thing... :)

(correct:) Namespaces > Create ...there is no error when name is empty, until you enter something and remove it again.
(here:) EE > Add execution environment ...shows the validation error right away:



(and we may need a better message for when the field is empty .. namespaces uses `Please, provide the namespace name
`)

ZitaNemeckova:
@himdel Fixed as suggested. Just had to move it to the `validateName` because PF component works with `"default"/"error"` instead of boolean. So it's ready for another round of review :)

src/components/execution-environment/repository-form.tsx
```diff
@@ -374,55 +373,77 @@ export class RepositoryForm extends React.Component<IProps, IState> {
...
+            </FormGroup>
+          )}
         </Form>
       </Modal>
     );
   }
 
+  private validateName(name) {
+    const regex = /^([0-9A-Za-z._-]+\/)?[0-9A-Za-z._-]+$/;
+    if (name === '' || regex.test(name)) {
```
ZitaNemeckova, line 430: @himdel This is the change you are looking for.

himdel: Works for me :), `formIsValid` still ensures non-empty name :+1:

himdel (approved):

himdel:
backporting manually in #1497

PR #1227, opened 2021-11-12
Remove filter drop down

MilanPospisil:
Remove filter drop down when it has only 1 item. It shows text of the selected item instead.

### DIFF
--- a/src/components/patternfly-wrappers/compound-filter.tsx
+++ b/src/components/patternfly-wrappers/compound-filter.tsx
@@ -78,21 +78,30 @@ export class CompoundFilter extends React.Component<IProps, IState> {
       </DropdownItem>
     ));
 
+    let select;
+    if (filterConfig.length != 1)
+    {
+      select = <StatefulDropdown
+      toggleType='dropdown'
+      defaultText={
+        <span>
+          <FilterIcon />
+          {'   '}
+          {selectedFilter.title}
+        </span>
+      }
+      position='left'
+      isPlain={false}
+      items={filterOptions}
+      />; 
+    }else
+    {
+      select = <div style={{verticalAlign : 'center'}}>{filterConfig[0].title}</div>;
+    }
+
     return (
       <InputGroup>
-        <StatefulDropdown
-          toggleType='dropdown'
-          defaultText={
-            <span>
-              <FilterIcon />
-              {'   '}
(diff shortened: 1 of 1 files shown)


MilanPospisil:
NOTE - what do you Martin think about the margin solution? I think its not good, but I could not make vertical align with table cell display work :( Though the solution worked in my testing html file, but it is not working in ansible...

My local test file works:
<html>
<div style="border-size:1; border-style:solid; height: 100px; display:table-cell;vertical-align:middle">
	Text!
</div>
</html>

ZitaNemeckova:
@MilanPospisil Before and After screenshots, please :)

MilanPospisil:
Before
old
:
After
new

himdel:
@sbuenafe-rh Do we have any designs for this please?

AFAICT right now the top padding is too much, the text should align with the "Filter by" text, I think it's off by ~2px (too low).
And we need some sort of right padding, to have some space between the text and the input field.

(by `the text` I mean "Container repository name")

src/components/patternfly-wrappers/compound-filter.tsx
```diff
@@ -93,6 +94,14 @@ export class CompoundFilter extends React.Component<IProps, IState> {
           isPlain={false}
           items={filterOptions}
         />
+      );
+    } else {
+      select = <div style={{ margin: '10px' }}>{filterConfig[0].title}</div>;
```
himdel, line 99: (I think you'll want *something* like `margin: '8px 4px 0 0'`, to only set the top and right one (+- numbers, or possibly margin vspadding))

But if it helps, `.pf-c-input-group` is a [flex element](https://css-tricks.com/snippets/css/a-guide-to-flexbox/), it might help with the centering, if centering is enough.

sbuenafe-rh:
@himdel it should only be the Filter by container repository name input field. **Container repository name** text in the front of this field should be removed.

MilanPospisil:
> 
> 
> @himdel it should only be the Filter by container repository name input field. **Container repository name** text in the front of this field should be removed.

So the name should not be visible at all?

himdel:
> So the name should not be visible at all?

That's right, the user can still see it in the placeholder or in active filters.




LGTM :+1:

himdel (approved):

PR #1249, opened 2021-11-15
Add GH Action to weekly update i18n

ZitaNemeckova:
To prevent weird strings in production when they are not extracted and compiled.

Fixes https://issues.redhat.com/browse/AAH-1123

### DIFF
--- a/.github/workflows/i18n.yml
+++ b/.github/workflows/i18n.yml
@@ -27,7 +27,6 @@ jobs:
         git config --global user.email 'gh_actions@users.noreply.github.com'
         git remote set-url origin https://x-access-token:${{ secrets.GITHUB_TOKEN }}@github.com/ansible/ansible-hub-ui.git
         git add locale/
-        git commit -m "Automated updated of i18n strings on $(date +'%Y-%m-%d')"
-        git push --set-upstream origin
-      env:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        if git commit -m "Automated updated of i18n strings on $(date +'%Y-%m-%d')"; then
+          git push --set-upstream origin
+        if
--- a/.github/workflows/i18n.yml
+++ b/.github/workflows/i18n.yml
@@ -4,8 +4,6 @@ on:
   # allow running manually
   workflow_dispatch:
   # run weekly
-  push:
-    branches: [ 'master', 'stable-4.4' ]
   schedule:
     - cron:  '30 5 * * 0'
 
--- a/.github/workflows/i18n.yml
+++ b/.github/workflows/i18n.yml
@@ -1,0 +1,35 @@
+name: i18n
+
+on:
+  # allow running manually
+  workflow_dispatch:
+  # run weekly
+  push:
+    branches: [ 'master', 'stable-4.4' ]
+  schedule:
+    - cron:  '30 5 * * 0'
+
+jobs:
+  i18n:
+    runs-on: ubuntu-latest
+    steps:
+
+    - uses: actions/checkout@v2
+    - name: "extract strings"
+      run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/i18n.yml
```diff
@@ -0,0 +1,35 @@
...
+    - name: "compile strings"
+      run: npm run gettext:compile
+
+    - name: "commit"
+      run: |
+        git config --global user.name 'GH Actions'
+        git config --global user.email 'gh_actions@users.noreply.github.com'
+        git remote set-url origin https://x-access-token:${{ secrets.GITHUB_TOKEN }}@github.com/ansible/ansible-hub-ui.git
+        git add locale/
+        git commit -m "Automated updated of i18n strings on $(date +'%Y-%m-%d')"
```
himdel, line 32: if there is nothing to push, git commit will fail ... which is a good thing because then we don't need to push...
.. but a bad thing because if will still create an "action failed" notification

```suggestion
        if git commit -m "Automated updated of i18n strings on $(date +'%Y-%m-%d')"; then
          git push --set-upstream origin
        fi
```

.github/workflows/i18n.yml
```diff
@@ -0,0 +1,35 @@
+name: i18n
+
+on:
+  # allow running manually
+  workflow_dispatch:
+  # run weekly
+  push:
+    branches: [ 'master', 'stable-4.4' ]
+  schedule:
+    - cron:  '30 5 * * 0'
```
himdel, line 10: So, is the idea to run this after every merge on master/stable-4.4 *and* on every sunday for master only?

(Asking because the title indicates this is only happening weekly.)

.github/workflows/i18n.yml
```diff
@@ -0,0 +1,35 @@
...
+    - name: "commit"
+      run: |
+        git config --global user.name 'GH Actions'
+        git config --global user.email 'gh_actions@users.noreply.github.com'
+        git remote set-url origin https://x-access-token:${{ secrets.GITHUB_TOKEN }}@github.com/ansible/ansible-hub-ui.git
+        git add locale/
+        git commit -m "Automated updated of i18n strings on $(date +'%Y-%m-%d')"
+        git push --set-upstream origin
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
himdel, line 35: This one is probably unused since we're passing the token in the url.

(`git push` doesn't read `GITHUB_TOKEN` from environment, only github-aware tools do)

.github/workflows/i18n.yml
ZitaNemeckova: Let's go weekly.

awcrosby:

himdel (approved):
:shipit: LGTM :+1: :)

.github/workflows/i18n.yml
```diff
@@ -0,0 +1,32 @@
...
+
+    - name: "commit"
+      run: |
+        git config --global user.name 'GH Actions'
+        git config --global user.email 'gh_actions@users.noreply.github.com'
+        git remote set-url origin https://x-access-token:${{ secrets.GITHUB_TOKEN }}@github.com/ansible/ansible-hub-ui.git
+        git add locale/
+        if git commit -m "Automated updated of i18n strings on $(date +'%Y-%m-%d')"; then
+          git push --set-upstream origin
+        if
```
himdel, line 32: Ah :)

```suggestion
        fi
```

https://github.com/ansible/ansible-hub-ui/runs/4572955455?check_suite_focus=true#step:5:10

himdel: (fixed in https://github.com/ansible/ansible-hub-ui/pull/1460)

PR #1252, opened 2021-11-16
Ux issues in delete modals

ShaiahWren:
This pr addresses UX issues discussed in [AAH-1104](https://issues.redhat.com/browse/AAH-1104)

UI changes:
Execution Environments: 
* add space between line with checkbox and line above it
* title on delete modal does _not_ contain "Permanently"
*
Remote Registries: 
* name of the item is in bold (the translator for this line is removed)
* after hitting delete on the list view page the page reloads -> delete item is not in the list

Collections:
* title on delete modal does _not_ contain "Permanently"

Namespaces:
* title on delete modal does _not_ contain "Permanently"

### DIFF
--- a/src/components/headers/collection-header.tsx
+++ b/src/components/headers/collection-header.tsx
@@ -267,8 +267,8 @@ export class CollectionHeader extends React.Component<IProps, IState> {
             isDisabled={!confirmDelete || isDeletionPending}
             title={
               collectionVersion
-                ? t`Permanently delete collection version?`
-                : t`Permanently delete collection?`
+                ? t`Delete collection version`
+                : t`Delete collection`
             }
             confirmButtonTitle={t`Delete`}
           >
--- a/src/components/headers/collection-header.tsx
+++ b/src/components/headers/collection-header.tsx
@@ -267,8 +267,8 @@ export class CollectionHeader extends React.Component<IProps, IState> {
             isDisabled={!confirmDelete || isDeletionPending}
             title={
               collectionVersion
-                ? t`Delete collection version`
-                : t`Delete collection`
+                ? t`Delete collection version?`
+                : t`Delete collection?`
             }
           >
             <>
--- a/src/components/headers/collection-header.tsx
+++ b/src/components/headers/collection-header.tsx
@@ -32,8 +32,8 @@ import {
   AlertList,
   AlertType,
   closeAlertMixin,
-  ConfirmModal,
   StatefulDropdown,
+  DeleteModal,
 } from 'src/components';
 
 import { CollectionAPI, CollectionDetailType } from 'src/api';
@@ -254,10 +254,10 @@ export class CollectionHeader extends React.Component<IProps, IState> {
           />
         </Modal>
         {deleteCollection && (
-          <ConfirmModal
+          <DeleteModal
             spinner={isDeletionPending}
             cancelAction={this.closeModal}
-            confirmAction={() =>
+            deleteAction={() =>
               this.setState({ isDeletionPending: true }, () => {
                 !!collectionVersion
                   ? this.deleteCollectionVersion(collectionVersion)
@@ -270,7 +270,6 @@ export class CollectionHeader extends React.Component<IProps, IState> {
                 ? t`Delete collection version`
                 : t`Delete collection`
             }
-            confirmButtonTitle={t`Delete`}
           >
             <>
               <Text style={{ paddingBottom: 'var(--pf-global--spacer--md)' }}>
@@ -309,7 +308,7 @@ export class CollectionHeader extends React.Component<IProps, IState> {
                 id='delete_confirm'
               />
             </>
-          </ConfirmModal>
+          </DeleteModal>
         )}
         <BaseHeader
           className={className}
--- a/src/containers/execution-environment-detail/execution_environment_detail_images.tsx
+++ b/src/containers/execution-environment-detail/execution_environment_detail_images.tsx
@@ -205,7 +205,7 @@ class ExecutionEnvironmentDetailImages extends React.Component<
         {deleteModalVisible && (
           <DeleteModal
             spinner={isDeletionPending}
-            title={t`Permanently delete image?`}
+            title={t`Delete image`}
             cancelAction={() =>
               this.setState({
                 deleteModalVisible: false,
--- a/src/containers/execution-environment-detail/execution_environment_detail_images.tsx
+++ b/src/containers/execution-environment-detail/execution_environment_detail_images.tsx
@@ -20,6 +20,7 @@ import {
   Checkbox,
   DropdownItem,
   LabelGroup,
+  Text,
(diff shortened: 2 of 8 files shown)


src/containers/execution-environment/registry-list.tsx
```diff
@@ -458,7 +462,8 @@ class ExecutionEnvironmentRegistryList extends React.Component<
       .catch(() =>
         this.addAlert(t`Failed to delete remote registry ${name}`, 'danger'),
       )
-      .then(() =>
+      .then(
+        () => this.queryRegistries(),
         this.setState({ showDeleteModal: false, remoteToEdit: null }),
```
himdel, line 467: this change causes the modal to disappear right away, instead of waiting for the promise (because it's now evaulated right away, instead of when the callback gets called)

```suggestion
        () => { this.queryRegistries(); this.setState({ showDeleteModal: false, remoteToEdit: null }); }
```

test/cypress/support/commands.js
```diff
@@ -618,6 +618,7 @@ Cypress.Commands.add('deleteRegistries', {}, () => {
   cy.wait('@registries').then((result) => {
     var data = result.response.body.data;
     data.forEach((element) => {
+      cy.wait(10000);
```
himdel, line 621: this might be caused by https://github.com/ansible/ansible-hub-ui/pull/1252/files#r755483687 .. try removing it after?

Also waiting 10 seconds per each registry is ..not optimal so we should try to fix this if possible :)

himdel:
There's still this one...

```
src/containers/execution-environment-detail/execution_environment_detail_images.tsx
209:            title={t`Permanently delete image`}
```

there's a checkbox there as well, so I think that should also be changed to Delete image?

src/containers/execution-environment/registry-list.tsx
ShaiahWren: Hello:)
I tried this change for removing the cy.wait(10000), and unfortunately cypress still did not pass. I see this error:
`Timed out retrying after 4050ms: cy.click() failed because this element is detached from the DOM.

<button aria-label="Actions" id="pf-dropdown-toggle-id-13" class="pf-c-dropdown__toggle pf-m-plain" type="button" aria-expanded="false" aria-haspopup="true">...</button>

Cypress requires elements be attached in the DOM to interact with them.

The previous command that ran was:

  > cy.get()

This DOM element likely became detached somewhere between the previous and current command.

Common situations why this happens:
  - Your JS framework re-rendered asynchronously
  - Your app code reacted to an event firing and removed the element

You typically need to re-query for the element or add 'guards' which delay Cypress from running new commands.Learn more
cypress/support/commands.js:626:9
  624 |           element.name +
  625 |           '"] button[aria-label="Actions"]',
> 626 |       ).click();
      |         ^
  627 |       cy.contains('a', 'Delete').click();
  628 |       cy.contains('button', 'Delete').click();
  629 |     });`

himdel:
No more "permanently deleted" anywhere :+1: 
Names of deleted item in bold everywhere :+1: 
Space before checkbox eveywhere with checkbox :+1: (no checkbox in registry, users, groups, group-users)
Wait and reload/navigate everywhere :+1: 

We can absolutely merge those parts :)
Still looking into the test issue...

And I noticed 2 more inconsitencies we may want to fix here:

* some of these modals have a question mark in the title, others don't  => proposing settling on the variant **with** a question mark for deletion confirmation modals
* most of these use the danger (=red) button variant for actually doing the deletion, some use a primary (=blue) button instead

|title|variant|note|
|-|-|-|
|Delete group?|danger||
|Remove user from group?|danger||
|Delete user?|danger||
|Delete remote registry?|danger||
|Delete container|danger|should be "Delete Execution Environment?" now|
|Delete image|danger|just missing `?`|
|Delete namespace?|primary|change to danger|
|Delete collection version|primary|missing `?`, change to danger|
|Delete collection|primary|missing `?`, change to danger|

test/cypress/support/commands.js
himdel: Ah, we're now reloading the screen after a registry deletion succeeds, but not waiting for it.
So this loop finds the matching item and tries to click it, while the list reloads and the item gets replaced with a new identical one.

Looks like this should help...

```diff
diff --git a/test/cypress/support/commands.js b/test/cypress/support/commands.js
index b6abf4b..89083db 100644
--- a/test/cypress/support/commands.js
+++ b/test/cypress/support/commands.js
@@ -618,7 +618,6 @@ Cypress.Commands.add('deleteRegistries', {}, () => {
   cy.wait('@registries').then((result) => {
     var data = result.response.body.data;
     data.forEach((element) => {
-      cy.wait(10000);
       cy.get(
         'tr[aria-labelledby="' +
           element.name +
           '"] button[aria-label="Actions"]',
       ).click();
       cy.contains('a', 'Delete').click();
       cy.contains('button', 'Delete').click();
+      cy.wait('@registries');
     });
   });
 });
```

src/components/headers/collection-header.tsx
```diff
@@ -267,10 +267,9 @@ export class CollectionHeader extends React.Component<IProps, IState> {
             isDisabled={!confirmDelete || isDeletionPending}
             title={
               collectionVersion
-                ? t`Permanently delete collection version?`
-                : t`Permanently delete collection?`
+                ? t`Delete collection version`
```
himdel, line 270: so this PR seems to be removing the question marks everywhere now? probably not intended?

src/containers/execution-environment-detail/execution_environment_detail_images.tsx
```diff
@@ -206,7 +207,7 @@ class ExecutionEnvironmentDetailImages extends React.Component<
         {deleteModalVisible && (
           <DeleteModal
             spinner={isDeletionPending}
-            title={t`Permanently delete image?`}
+            title={t`Delete image`}
```
himdel, line 210: (?)

src/containers/execution-environment-list/execution_environment_list.tsx
```diff
@@ -180,16 +181,18 @@ class ExecutionEnvironmentList extends React.Component<
         {deleteModalVisible && (
           <DeleteModal
             spinner={isDeletionPending}
-            title={'Permanently delete container?'}
+            title={'Delete container'}
```
himdel, line 184: (?)

ShaiahWren:
I think this is ready :) @himdel want to take another look?

src/containers/execution-environment-detail/execution_environment_detail_images.tsx
himdel: I think this one is still missing

ShaiahWren:
Hopefully, better now:)

himdel (approved):

himdel:
Works for me :+1: :)

ShaiahWren:
Issue: AAH-1104

PR #1253, opened 2021-11-17
Test collections list

MilanPospisil:
Test for Collections list. Menu => Collections -> Collections. It tests paging, filter, page size and cards/list switcher. it does not delete its data, because galaxykit does not supports it yet.


test/cypress/integration/collections_list.js
```diff
@@ -0,0 +1,75 @@
...
+    // this page cant sort items, so we can only count them, there should be at least 21 items that we inserted
+    cy.get('.collection-container').get('article').should('have.length', 10);
+
+    cy.get('.cards').get('[aria-label="Go to next page"]:first').click();
+    cy.get('.collection-container').get('article').should('have.length', 10);
+
+    cy.get('.cards').get('[aria-label="Go to next page"]:first').click();
+    // some remaining data can be there from previous tests
+    var remaining = items.length - 20;
+    if (remaining > 10) remaining = 10;
```
himdel, line 49:
```suggestion
    const remaining = items.length > 30 : items.length - 20 : 10;
```

test/cypress/integration/collections_list.js
```diff
@@ -0,0 +1,75 @@
...
+    // some data may be present from previous test, so we must load them, we can not expect
+    // that only our test data are in database.
+    cy.intercept(
+      'GET',
+      Cypress.env('prefix') +
+        '_ui/v1/repo/published/?deprecated=false&offset=0&limit=100',
+    ).as('data');
+    cy.visit('/ui/repo/published?page_size=100&view_type=null&page=1');
+
+    let data = [];
```
himdel, line 28: unused var?

test/cypress/integration/collections_list.js
MilanPospisil: Do you ment rather: const remaining = items.length > 30 ? 10 : items.length - 20

test/cypress/integration/collections_list.js
```diff
@@ -9,7 +9,7 @@ describe('Collections list Tests', () => {
     cy.login(adminUsername, adminPassword);
 
     // insert test data
-    range(21).forEach((i) => {
+    range(41).forEach((i) => {
```
himdel, line 12: Please don't.

Change the number of per page items if you need to test more pages. There is absolutely no reason to create 40 collections, I was already questioning 20 :)

MilanPospisil: Ah, this was only my local test if it works, this should not supposed go to PR :-)

himdel (approved):

PR #1255, opened 2021-11-18
Remove extra padding from Namespace Detail Collections List

MilanPospisil:
The same problem as #1222


himdel:
| before | after |
|-|-|
|  |  |

himdel (approved):

PR #1256, opened 2021-11-18
Implement kebab toggle for delete and edit modals

ShaiahWren:
See issue: https://issues.redhat.com/browse/AAH-829

Assignments:
1. replace delete button with kebab that contains delete and edit options
2. implement edit from Group list view

This was done by converting delete button to a kebab with delete __and__ edit options from the group list view. The edit dropdown option is now connected to logic that a) redirects to group detail and b) opens edit mode by assigning a new param `isEditing` to `true` in the group detail component. `editPermissions` state is assigned to `isEditing`.


src/containers/group-management/group-list.tsx
```diff
@@ -380,20 +413,8 @@ class GroupList extends React.Component<RouteComponentProps, IState> {
           </Link>
         </td>
         <td>
-          {!!user && user.model_permissions.delete_group && (
-            <Button
-              aria-label={t`Delete`}
```
himdel, line 385: can you keep the `aria-label` on these buttons?

Or we need to fix tests to look for the button text, instead of the label..

src/containers/group-management/group-list.tsx
```diff
@@ -368,6 +370,38 @@ class GroupList extends React.Component<RouteComponentProps, IState> {
...
+          onClick={() => {
+            this.setState({
+              selectedGroup: { ...group },
+              redirect: formatPath(Paths.groupDetail, {
+                group: group.id,
+              }),
+            });
+          }}
+        >
+          <Trans>Edit group</Trans>
```
himdel, line 386: should this just be `Edit`? I'm not seeing `Edit {thing}` used anywhere but tab/modal titles, the other edit buttons just say Edit

src/containers/group-management/group-list.tsx
```diff
@@ -368,6 +370,38 @@ class GroupList extends React.Component<RouteComponentProps, IState> {
...
   private renderTableRow(group: any, index: number) {
     const { user } = this.context;
+    const dropdownItems = [
+      <React.Fragment>
+        <DropdownItem
+          key='edit'
+          onClick={() => {
+            this.setState({
+              selectedGroup: { ...group },
+              redirect: formatPath(Paths.groupDetail, {
```
himdel, line 380: is this missing something? It redirects to the group detail, but doesn't trigger edit mode.

ShaiahWren: Yes, it is missing something:) I took a look today. 
In group-list I added a param 'isEditing' to this.state. This I set to false.
On the route redirect in group-detail I _attempted_ to update the isEditing param from group-detail to true. My syntax must not be correct and I'm not finding any insight from online sources. 

Then I added logic to renderPermissions: ` isViewOnly={!editPermissions && !isEditing}` so when either variable is true it should be in edit mode.

If you could offer input regarding my syntax, or faulty logic, I would appreciate it very much. :)

himdel (approved):
Works for me :+1: :)

ShaiahWren:
Issue: AAH-829

PR #1257, opened 2021-11-18
create single  'Get Token' button for all local repos

ShaiahWren:
Issue: https://issues.redhat.com/browse/AAH-903

Assignment:

1. Only one "Get Token" button for the list view in the header. Only when 'local' repos tab is selected
2. Button does not redirect to a new tab

Before: 
Screen Shot 2021-11-18 at 4 09 15 PM

After:  
Screen Shot 2021-11-30 at 4 16 16 PM

### DIFF
--- a/src/containers/repositories/repository-list.tsx
+++ b/src/containers/repositories/repository-list.tsx
@@ -1,7 +1,6 @@
 import { t } from '@lingui/macro';
 import * as React from 'react';
-
-import { withRouter, RouteComponentProps } from 'react-router-dom';
+import { withRouter, RouteComponentProps, Link } from 'react-router-dom';
 
 import {
   BaseHeader,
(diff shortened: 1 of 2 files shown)


himdel (approved):

src/containers/repositories/repository-list.tsx
```diff
@@ -179,6 +181,13 @@ class RepositoryList extends React.Component<RouteComponentProps, IState> {
                   />
                 </div>
               </div>
+              {params.tab == 'local' && (
```
ZitaNemeckova, line 184: @ShaiahWren This should be in `BaseHeader` param `pageControls`. See https://github.com/ansible/ansible-hub-ui/blob/2e1946f7c7e344c7bc4d6a017e0199ba6aeb6870/src/containers/group-management/group-detail.tsx#L200-L215 for reference.

After moving it there shouldn't be any need for new CSS and it should consistent with similar pages.

ZitaNemeckova (changes requested):

awcrosby:

ShaiahWren:
Issue: AAH-903

himdel:
(backport failed because of a conflict with (missing) #1259 , backporting that one first)

PR #1258, opened 2021-11-18
add tasking system  permissions to the UI

ShaiahWren:
Issue: https://issues.redhat.com/browse/AAH-955

Assignment: Add tasking system permissions so that all Users can view and manage tasks.

Before:
Screen Shot 2021-11-19 at 10 48 07 AM

After:
Screen Shot 2021-11-19 at 10 55 04 AM

### DIFF
--- a/src/constants.tsx
+++ b/src/constants.tsx
@@ -99,6 +99,15 @@ export class Constants {
         'galaxy.delete_containerregistryremote', // model_permissions.delete_containerregistry
       ],
     },
+    {
+      name: 'task_management',
+      label: t`Task Management`,
+      object_permissions: [
+        'core.change_task',
+        'core.delete_task',
+        'core.view_task',
+      ],
+    },
 
     // These aren't currently used. Removing them to reduce confusion in the UI
     // {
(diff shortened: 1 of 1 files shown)


newswangerd (changes requested):
These should go under a new category called "Tasking System", not Users

src/constants.tsx
```diff
@@ -132,6 +132,12 @@ export class Constants {
     // },
   ];
 
+  static TASKING_PERMISSIONS = [
```
newswangerd, line 135: I believe this has to go in the `PERMISSIONS` variable. You'll know it's set up correctly when you see the permissions appear on the group permissions screen

newswangerd (changes requested):

src/constants.tsx
ShaiahWren: @newswangerd The new Tasking category shows up in the UI, _however_, the dropdown options do not appear (see image)
The format is identical to the others categories under PERMISSIONS. Any insights would be appreciated.:)

Dropdown options are not visible:
Screen Shot 2021-11-19 at 10 35 13 AM

newswangerd (approved):

PR #1260, opened 2021-11-19
EE Detail page - Edit button in the header is primary

MilanPospisil:
Change edit button from primary to secondary.
Issue - https://issues.redhat.com/browse/AAH-1114


himdel (approved):

himdel:
Backporting as a dependency of #1553

PR #1263, opened 2021-11-20
Update react router to 6

himdel:
This updates react-router to v6, specifically ^6.6.1

[Upgrading from v5](https://reactrouter.com/en/v6/upgrading/v5)
[cloud arch for v6](http://front-end-docs-insights.apps.ocp4.prod.psi.redhat.com/blog/router-v6)

(related to https://github.com/RedHatInsights/tower-analytics-frontend/pull/908)

---

Main changes:

router v6 changes `Route` from using `component={Foo}` to `element={<Foo /}`, and no longer provides `RouteComponentProps` - `history`, `match`, `location`
=> implemented our own `withRouter` to wrap class components
(changes from react-router-dom to src/utilities; props renamed:
`history.push` -> `navigate`
`match.params` -> `routeParams`
`match.path` -> `routePath`
(`location` -> `location`))
(`routePath` is now no longer available from the router, so we pass it as a prop from route element defs; the rest comes from hooks)

..but App also needs location and lives above the routes
=> used the useLocation hook in a wrapper componet (standalone), added useLocation (insights, already functional)

..and repo-selector also needs it
=> converted to functional, use hook

Then there is no `Redirect`, => changed to `<Navigate to=...`,
changed `history.push` and `redirect` calls to `navigate` calls (from props or `useNavigate`).

Optional route segments are no longer supported - affects containers
=> created Route + RouteWithNamespace pairs, `/ee/:container+` becomes `/ee/:container` and `/ee/:namespace/:container`,
=> added `formatEEPath` to chose the long or short path and split container into namespace/container,
=> added `withContainerParamsFix` to convert `{namespace, container}` `routeParams` to `{container}`
=> added `encodeURIComponent` to `formatPath`, so any missing namespaced routes will just end up using `/ee/foo%2Fbar/` instead of `/ee/foo/bar/` and still work

Insights mode runs with different basename, v6 router sees basename=/beta (or /) with hub under a nested `ansible/automation-hub` route .. so `Link to=/` goes to `/beta/`. But the pathnames are still right for route definitions.
=> adding `formatPath` to every use of `Path.foo` except ones still to be used for `formatPath`, or for route definitions
=> changing `formatPath` to prefix `..` in insights mode - `formatPath('/')` now returns `../` which resolves to `/beta/ansible/automation-hub/` in links (as `..` means up one route nesting level in v6).
EDIT: Actually, cutting out the non-beta part of UI_BASE_PATH and using that .. the `../` works for Link and Navigate, but `matchPath` needs to match against `/ansible/automation-hub/repo/:repo/*`.
Also no `Router` in insights - it's provided by chrome.

`matchPath` changes order of args, but also matching algorithm  - no longer matching by prefix
=> using `formatPath('/repo/:repo/') + '*'` to construct a wildcard route and match against that

router no longer renders default route when outside basename - `/` is not in `/ui` - so no thing renders
=> adding a manual history.push to go to `UI_BASE_NAME` when outside (standalone only)

Also drops `redux` and `prop-types`, no longer needs either. 
And separated the standalone rendering logic, menu logic and context logic, converting into separate functional components, with context logic mostly similar to insights.

And then namespace deletion doesn't show success alert - because NamespaceDetail unmount now clears context alerts before NamespaceList can display anything.
=> clearing context alerts only after copying them to new component's state


himdel:
(waiting until #1373 is in)

himdel:
Relevant again, prod stable should be [ready for router updates] on January 5th 2023.

himdel:
The remaining screenshots.js diff is a good change - going to /ui/repo/published now shows the Collections > Collections menu item as active.

MilanPospisil (approved):
LGTM 👍 
I dont see anything wrong with this. But the code is to huge for me to be able ensure it is working correctly. I tested manually both standalone and insights and new insights tests are coming, so we will be able to auto test as much as possible.

PR #1284, opened 2021-11-22
EE Add/Edit form - add required to name, namespace name and registry

MilanPospisil:
Issue - https://issues.redhat.com/browse/AAH-1110


himdel:
Works just fine, doesn't prevent local repos from being validated, LGTM :+1:

PR #870, opened 2021-09-02
Rework of collection version select

jerabekjiri:
Issue: [AAH-907](https://issues.redhat.com/browse/AAH-907)

- Removed collection version select from collection-info.tsx 
- Used Patternfly Select instead of FormSelect for nicer styling

before:
Screenshot from 2021-09-02 16-48-17
after:
Screenshot from 2021-09-02 14-36-58

- Added modal for  full list of collections
Screenshot from 2021-09-02 14-37-25

Full content screen with removed version select inside install content and with new select component :)
Screenshot from 2021-09-02 14-37-11

@ZitaNemeckova  please take a look :)

PR #1289, opened 2021-11-22
[stable-4.2] Collection detail - make both version selects affect the download tarball version

himdel:
Fixes AAH-1067 for stable-4.3 and stable-4.2, by making the "Download tarball" button default to `params.version` instead of `latest_version.version`.

(while for stable-4.4+, this was fixed in https://github.com/ansible/ansible-hub-ui/pull/870/, by removing the non-header selector, and latest_version is actually the selected version)

And adding a `setVersion` prop to `CollectionInfo`, `CollectionHeader` & having `CollectionDetail` share `setVersion` between both allows the bottom version selector to update the top one when changed.

### DIFF
--- a/src/components/collection-detail/collection-info.tsx
+++ b/src/components/collection-detail/collection-info.tsx
@@ -27,7 +27,7 @@ interface IProps extends CollectionDetailType {
   params: {
     version?: string;
   };
-  updateParams: (params) => void;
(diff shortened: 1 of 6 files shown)


himdel:
(this "backports" cleanly to 4.3, adding the label to do so)

newswangerd (approved):

PR #1290, opened 2021-11-23
Fix Cli Configuration tab on Namespace detail page

MilanPospisil:
https://issues.redhat.com/browse/AAH-825


himdel:
before:
                          

after:

himdel (approved):

PR #1295, opened 2021-11-24
Fix translation in repo selector 

jerabekjiri:
Issue: [AAH-1070](https://issues.redhat.com/browse/AAH-1070)

Strings that are marked for translation don't seem to be working if imported from a different file. This is fixed by returning the value from a function and then evoking the function. Not sure this is the right way, I couldn't find any mention about this in the Lingui documentation, wdyt @himdel ?

before:
Screenshot from 2021-11-24 16-26-49

after:
Screenshot from 2021-11-24 16-11-32


jerabekjiri:
This is the same case happening in Task Management, the tooltip should be translated, but it is imported from `constants.tsx` and therefore not being translated at the moment.
Screenshot from 2021-11-24 16-32-24

https://github.com/ansible/ansible-hub-ui/blob/2acbd97a3fb466781f8bf80a6b34ec04751bb899/src/constants.tsx#L278-L279

himdel:
Aah @jerabekjiri this is not really about different files, this is about the order in which things run..

Constants in modules are evaluated right away, before lingui switches the language to the right one .. so thay stay in English.

Normally this is solved by a 2 step translation... using `N_('Foo')` instead of `'_('Foo')` causes the string to get indexed into the translation catalog but returns the original string... so in traditional gettext, you'd have `REPOSITORYNAMES = { published: N_("Published") };` and then use `_(REPOSITORYNAMES.published)`  - we know the string is in the catalog thanks to the `N_` so we can safely pass it to `_` inside a variable.

Except lingui doesn't have an `N_` equivalent that I can see - `i18n._(str)` can be used for the second part, but I see no way to index the strings without translating them right away, without doing something weird like ``{ published: (t`My message`, "My message), }``.

---

So.. I think the approach of using functions instead of constants is valid, definitely for this case, but I'm less inclined to wrap every single constant with a human name in a function, that doesn't sound right.

So I think we should not reinvent the wheel, and come up with a way to achieve `N_` with lingui.

Do you want to look into that or should I?

himdel:
One more note.. Looking at [gettext docs](https://www.gnu.org/software/gettext/manual/gettext.html), `N_` is also called `gettext_noop`... and looking for that + lingui led me to https://github.com/lingui/js-lingui/issues/60 which adds a `i18nMark` to an ancient lingui ... aand it's no longer there :(.

Maybe we can reintroduce it? (Or, failing that, we *could* write our own babel `N_` macro that translates `` N`foo` `` to `` t`foo`, `foo` ``, so they get indexed but return the original.)

jerabekjiri:
We could theoretically use `defineMessage`. From docs: `defineMessage is transformed into message descriptor. In other words, the message isn’t translated directly and can be used anytime later`. That would mean we would still have to mark all constants `community: defineMessage({message: 'Community' })`. And then still use `i18n._(community.id)` to translate it.

himdel:
Aah, sounds like you've found it, yeah, that sounds about right :+1:  (docs - [lazy translations](https://lingui.js.org/tutorials/react-patterns.html?highlight=definemessage#lazy-translations), [defineMessage](https://lingui.js.org/ref/macro.html?highlight=definemessage#jsmacro-defineMessage))

jerabekjiri:
Reworked it to the `defineMessage`. We still have to do `i18n._(repo)` in the component though.

Also found out this `t({ message: 'Published' })` works as well.

src/constants.tsx
```diff
@@ -246,9 +247,9 @@ export class Constants {
   ];
 
   static REPOSITORYNAMES = {
-    published: t`Published`,
-    'rh-certified': t`Red Hat Certified`,
-    community: t`Community`,
+    published: t({ message: `Published` }),
```
himdel, line 250: `defineMessage` so this doesn't get double translated if the language is already set here

src/containers/task-management/task_detail.tsx
```diff
@@ -201,8 +203,10 @@ class TaskDetail extends React.Component<RouteComponentProps, IState> {
                               task: parentTaskId,
                             })}
                           >
-                            {Constants.TASK_NAMES[parentTask.name] ||
-                              parentTask.name}
+                            {i18n._(
+                              Constants.TASK_NAMES[parentTask.name] ||
+                                parentTask.name,
+                            )}
```
himdel, line 209:
```suggestion
                            {i18n._(
                              Constants.TASK_NAMES[parentTask.name]) ||
                                parentTask.name
                            }
```

(+- prettier anyway :))

src/containers/task-management/task_detail.tsx
```diff
@@ -467,8 +473,9 @@ class TaskDetail extends React.Component<RouteComponentProps, IState> {
             childTasks,
             parentTask,
             loading: false,
-            taskName:
+            taskName: i18n._(
               Constants.TASK_NAMES[result.data.name] || result.data.name,
+            ),
```
himdel, line 478:
```suggestion
            taskName: i18n._(
              Constants.TASK_NAMES[result.data.name]) || result.data.name,
```

src/containers/task-management/task_detail.tsx
```diff
@@ -225,8 +229,10 @@ class TaskDetail extends React.Component<RouteComponentProps, IState> {
                                       task: childTaskId,
                                     })}
                                   >
-                                    {Constants.TASK_NAMES[childTask.name] ||
-                                      childTask.name}
+                                    {i18n._(
+                                      Constants.TASK_NAMES[childTask.name] ||
+                                        childTask.name,
+                                    )}
```
himdel, line 235:
```suggestion
                                    {i18n._(
                                      Constants.TASK_NAMES[childTask.name]) ||
                                        childTask.name
                                    }
```

himdel:
> Also found out this t({ message: 'Published' }) works as well.

Works but it tries to translate the message when a language is set. Though, I suspect that means `t(Constants.TASK_NAMES[childTask.name])` would work as well. (But it makes sense to stil use `i18n._`  to keep them distinct.)

---

Verified that `defineMessage` still causes the string to end up in the catalog with `npm run gettext:extract` :+1:

himdel:
Works in the UI, I think the repo selector changes make sense as well :+1: 

So, just that one `defineMessage`, and `i18n._(x || y)` -> `i18n._(x) || y` and it's ready :)

src/containers/task-management/task-list-view.tsx
```diff
@@ -271,7 +273,7 @@ export class TaskListView extends React.Component<RouteComponentProps, IState> {
       <tr aria-labelledby={name} key={index}>
         <td>
           <Link to={formatPath(Paths.taskDetail, { task: taskId })}>
-            <Tooltip content={Constants.TASK_NAMES[name] || name}>
+            <Tooltip content={i18n._(Constants.TASK_NAMES[name] || name)}>
```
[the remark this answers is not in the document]
jerabekjiri: @himdel This will crash if `Constants.TASK_NAMES[name]` is `undefined`, `i18n._(...)` will throw error. 

```Constants.TASK_NAMES[name] ? (i18n._(Constants.TASK_NAMES[name]) || name) : name``` something like this would work. We first check if `Constants.TASK_NAME[name]` exists and then try to translate with `i18n._(...)`, if not, return `name`. The code seems a bit longer though, that's why I used that first variant.

jerabekjiri:
@himdel As I mentioned here [comment](https://github.com/ansible/ansible-hub-ui/pull/1295#discussion_r769620037), we can't do `i18n._(undefined)`, which will result in error. 

So using it just like `i18n._(x) || y` isn't possible. We would have to do something like `!!x ? (i18n._(x) || x) : x` or stick to  `i18n._(x || y)`.

himdel:
Well, we *could* write that as `x && i18n._(x) || y`, which may be slightly shorter, and maybe we can wrap it in some kind of wrapper functions, but I guess no way around it. Luckily it's not in that many places :)

himdel (approved):
Works for me :)

PR #1317, opened 2021-12-02
Remove extra padding from EE Detail Images (No Issue)

MilanPospisil:
The same problem as #1222


src/containers/execution-environment-detail/execution-environment-detail_images.scss
```diff
@@ -1,3 +1,9 @@
 .tags-column {
   max-width: 250px;
 }
+
+.body .toolbar {
```
himdel, line 5: this would be a global style, we need some kind of class to limit this to a specific screen

himdel (approved):

PR #1319, opened 2021-12-02
Remove extra padding from Approval Dashboard (No issue)

MilanPospisil:
The same problem as #1222

### DIFF
--- a/src/containers/certification-dashboard/certification-dashboard.scss
+++ b/src/containers/certification-dashboard/certification-dashboard.scss
@@ -1,4 +1,4 @@
-.body .toolbar {
+.certification-dashboard-toolbar {
   display: flex;
   justify-content: space-between;
   align-items: center;
--- a/src/containers/certification-dashboard/certification-dashboard.scss
+++ b/src/containers/certification-dashboard/certification-dashboard.scss
@@ -1,10 +1,10 @@
-.certification-dasboard {
-  .toolbar {
-    padding-bottom: 16px;
-    display: flex;
-    justify-content: space-between;
-  }
+.body .toolbar {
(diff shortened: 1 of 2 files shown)


src/containers/certification-dashboard/certification-dashboard.scss
```diff
@@ -1,10 +1,10 @@
-.certification-dasboard {
-  .toolbar {
-    padding-bottom: 16px;
-    display: flex;
-    justify-content: space-between;
-  }
+.body .toolbar {
```
himdel, line 1: this would affect every single toolbar in the ui, not only certification dashboard, the `.certification-dashboard {` wrapper is needed

himdel:
Also, can you add a before/after screenshot? I don't think Approval Dashboard even has this padding problem

MilanPospisil:
> Also, can you add a before/after screenshot? I don't think Approval Dashboard even has this padding problem

Before:
ApprovalDashboardPaddingBefore
After:
ApprovalDashboardPaddingAfter

src/containers/certification-dashboard/certification-dashboard.scss
MilanPospisil: Time to learn more about css :-D Makes sense, this is somewhat single page application, so when css are loaded, they are "alive" for the whole application, whereas in old apps, whole page was rerendered, so css were valid only for the one page and they dont have to be that specific, it is right?

himdel:
Aah, thanks, I was looking at the bottom bits, yeah, looks good :+1:
And so does the css now :)

himdel (approved):

PR #1323, opened 2021-12-05
workflows: add cancel-in-progress to cypress, deploy/release, update-manifest

himdel:
Workflows - using named concurrency groups allows us to automatically cancel a github action if the same one is running on the same branch - this should just kill double builds on double push and fast subsequent merges.

For cypress this just saves cpu time,
for deploy/release/update-manifest it should prevent ocasional failures trying to push to an updated branch.

(ref https://yonatankra.com/7-github-actions-tricks-i-wish-i-knew-before-i-started/, https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#concurrency)

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -11,6 +11,10 @@ on:
   schedule:
     - cron:  '30 5 * * *'
 
+concurrency:
+  group: cypress-${{ github.ref }}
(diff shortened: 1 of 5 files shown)


ZitaNemeckova (approved):
LGTM 👍

PR #1342, opened 2021-12-07
webpack.base: update for new frontend component config - bundlePfModules and htmlPlugin.inject

himdel:
Replaces #1188

As of `@redhat-cloud-services/frontend-components-config` 4.5.0+, patternfly is removed from the build so as not to clash with other patternfly versions in cloud.
This is not desirable for standalone/standalone-keycloak => disabling for non-insights builds.
But the option only affects `config.resolve.alias`, which we override in `webpack-ts-overrides.js`.
So merging the overrides together with base so we can merge the config properly and get both sets of aliases.

As of `@redhat-cloud-services/frontend-components-config` 4.5.4+, `HtmlWebpackPlugin` is configured to not inject js & css to html by default.
This is not desirable for standalone/standalone-keycloak => setting inject to true for non-insights builds.

Also added a couple of FIXME comments with things we may want to resolve for cloud config, while things still work without it :)

Cc @ZitaNemeckova , we may want to consider merging with #920. (Or #1373.)


himdel:
backports

4.4: TODO (via #1341 ?)
4.3: TODO
4.2: no, newer versions need newer webpack, https://github.com/ansible/ansible-hub-ui/pull/776

config/webpack.base.config.js
```diff
@@ -71,12 +77,63 @@ module.exports = (inputConfigs) => {
...
+
+    // override from empty
+    devtool: 'source-map',
+
+    module: {
+      ...webpackConfig.module,
+
+      // override to drop ts-loader
+      rules: [
+        // FIXME(cloud): add chrome-render-loader here?
```
himdel, line 97: ah, not needed because we removed the obsolete entry.js files in #920

himdel:
Ah, this only passed cypress before updating from 4.5.6 to 4.5.8 :)...
the difference is harmless, an extra `App.(hash).js` becomes `App.(timestamp).(hash).js` but fails the `curl` check .. and I failed to recognize the error and assumed cypress failed downloading :).

So...
* updated the cypress task to give these checks individual steps, and output more info there (expected vs reality)
* fixed BUILD_HASH parsing to read both timestamp and hash

It should go green this time @ZitaNemeckova :)

ZitaNemeckova (approved):
LGTM 👍

PR #1343, opened 2021-12-07
Cypress test for imports filter

MilanPospisil:
Cypress test for imports filter. 

Test partial match search for name and exact match search for status.

https://issues.redhat.com/browse/AAH-788

### DIFF
--- a/src/components/my-imports/import-list.tsx
+++ b/src/components/my-imports/import-list.tsx
@@ -189,7 +189,7 @@ export class ImportList extends React.Component<IProps, IState> {
     return (
       <div>
         <div>
-          <span className='item-name'> {item.name} </span>{' '}
+          <span data-cy='item-name'> {item.name} </span>{' '}
           {item.version ? 'v' + item.version : ''}
         </div>
         <div className='sub-text'>
--- a/src/components/my-imports/import-list.tsx
+++ b/src/components/my-imports/import-list.tsx
@@ -118,7 +118,7 @@ export class ImportList extends React.Component<IProps, IState> {
           }}
         />
 
-        <div className='import-list-data'>
+        <div data-cy='import-list-data'>
           {this.renderList(selectImport, importList, selectedImport, loading)}
         </div>
         <Pagination
--- a/src/components/my-imports/import-list.tsx
+++ b/src/components/my-imports/import-list.tsx
@@ -103,6 +103,7 @@ export class ImportList extends React.Component<IProps, IState> {
             ]}
           />
         </Toolbar>
+
         <AppliedFilters
           updateParams={(p) => {
             p['page'] = 1;
@@ -117,7 +118,7 @@ export class ImportList extends React.Component<IProps, IState> {
           }}
         />
 
-        <div>
+        <div className='import-list-data'>
           {this.renderList(selectImport, importList, selectedImport, loading)}
         </div>
         <Pagination
@@ -188,7 +189,8 @@ export class ImportList extends React.Component<IProps, IState> {
     return (
       <div>
         <div>
-          {item.name} {item.version ? 'v' + item.version : ''}
+          <span className='item-name'> {item.name} </span>{' '}
+          {item.version ? 'v' + item.version : ''}
         </div>
         <div className='sub-text'>
           Status: {item.state}{' '}
--- a/src/components/my-imports/import-list.tsx
+++ b/src/components/my-imports/import-list.tsx
@@ -189,7 +189,7 @@ export class ImportList extends React.Component<IProps, IState> {
     return (
       <div>
         <div>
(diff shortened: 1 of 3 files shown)


src/components/my-imports/import-list.tsx
```diff
@@ -117,7 +118,7 @@ export class ImportList extends React.Component<IProps, IState> {
           }}
         />
 
-        <div>
+        <div className='import-list-data'>
```
ShaiahWren, line 121: Check in with @himdel and/or @ZitaNemeckova about this.  Cypress documentation recommends using the data-cy attribute for selectors; classNames are brittle selectors and should be reserved for styling. See this chart: https://docs.cypress.io/guides/references/best-practices#:~:text=Selector-,recommended,-Notes

src/components/my-imports/import-list.tsx
```diff
@@ -188,7 +189,8 @@ export class ImportList extends React.Component<IProps, IState> {
     return (
       <div>
         <div>
-          {item.name} {item.version ? 'v' + item.version : ''}
+          <span className='item-name'> {item.name} </span>{' '}
```
ShaiahWren, line 192: See above comment and https://docs.cypress.io/guides/references/best-practices#:~:text=Selector-,recommended,-Notes

ShaiahWren (commented):
I see two more ways to test this: 
1) apply filters for partial name match _and_ status
2) apply filters for exact name match _and_ status

test/cypress/integration/imports_filter.js
```diff
@@ -0,0 +1,56 @@
+import { range, sortBy } from 'lodash';
+
+describe('Imports filter test', () => {
+  var adminUsername = Cypress.env('username');
+  var adminPassword = Cypress.env('password');
+
+  before(() => {
+    cy.login(adminUsername, adminPassword);
+
```
ShaiahWren, line 9: would it be important to delete all test collections before creating them with galaxykit? I think this may be why cypress isn't passing

src/components/my-imports/import-list.tsx
MilanPospisil: You are right, I used this in my previous test and I forgot :-)

test/cypress/integration/imports_filter.js
MilanPospisil: That would be great, but I dont know if the galaxykit has this possibility. I had plenty of problems with previous tests including collections and namespaces, which had to be deleted manualy using UI and thats a bit of work :-(

MilanPospisil: It seems test is working without changes, there had to be some unrelated problem - it failed when loading the page. I rerun the tests twice and no error.

src/components/my-imports/import-list.tsx
ShaiahWren: missed one :)

ShaiahWren:
https://github.com/ansible/ansible-hub-ui/pull/1343#:~:text=left%20a%20comment-,i%20see%20two%20more,-ways%20to%20test
Will these tests be added?

MilanPospisil:
> Waiting on the two more tests to be added?

What tests do you mean?

MilanPospisil:
> > Waiting on the two more tests to be added?
> 
> What tests do you mean?

Ah, I see, you want to test namespace and other filters together, why not :-) I will do it, but I rather first wait for the galaxykit support for deletion, because there is so much garbage data, so it is hard to do proper tests without clean database.

src/components/my-imports/import-list.tsx
```diff
@@ -188,7 +189,8 @@ export class ImportList extends React.Component<IProps, IState> {
     return (
       <div>
         <div>
-          {item.name} {item.version ? 'v' + item.version : ''}
+          <span data-cy='item-name'> {item.name} </span>{' '}
```
himdel, line 192:
```suggestion
          <span data-cy='item-name'>{item.name}</span>{' '}
```

otherwise the space would offset the padding

test/cypress/integration/imports_filter.js
```diff
@@ -0,0 +1,132 @@
+describe('Imports filter test', () => {
+  var adminUsername = Cypress.env('username');
```
himdel, line 2: we no longer need these, you can use `cy.login()` for admin (see https://github.com/ansible/ansible-hub-ui/pull/1346)

MilanPospisil: Good idea :-)

ShaiahWren (approved):
This pr LGTM. Nice work testing both filters at the same time:)
I approve, but you might want to let Martin give the final say:)

himdel (approved):
What Shaiah said :+1: 
LGTM as well :)

PR #1346, opened 2021-12-08
cy.login() - default to admin

himdel:
we seem to read the admin username/password config in every test, only to login
defaulting cy.login() to admin, and cleaning up the tests

(kept uses of adminUsername/adminPassword for apiLogin, manualLogin, cookieLogin, contains and attemptToDelete)


ZitaNemeckova (approved):
LGTM 👍

PR #1368, opened 2021-12-13
Execution environments - add Delete to detail screen

MilanPospisil:
AAH-1089

From Issue:
problem statement: Execution environments options do not match between list and detail screens:

    ee list has Edit, Sync, Use, Delete
    ee detail has Edit, Sync, Use (no Delete)

acceptance criteria:

    EE list and detail have the same actions

PR Description:
There is need to implement delete in both list and detail. But delete is in list only. First step was to refactor the list and create standalone DeleteExecutionEviromentComponent that will replace delete functionality in list container. Now the delete component can be use also in the detail container.

### DIFF
--- a/src/containers/execution-environment-detail/base.tsx
+++ b/src/containers/execution-environment-detail/base.tsx
@@ -6,6 +6,7 @@ import {
   ContainerRepositoryType,
   ExecutionEnvironmentAPI,
   ExecutionEnvironmentRemoteAPI,
+  ExecutionEnvironmentType,
 } from 'src/api';
 import { formatPath, Paths } from '../../paths';
 import { Button, DropdownItem } from '@patternfly/react-core';
@@ -22,6 +23,8 @@ import {
 } from 'src/components';
 import { parsePulpIDFromURL, waitForTask } from 'src/utilities';
 
+import { DeleteExecutionEnviromentModal } from 'src/containers/execution-environment-detail/delete-execution-enviroment-modal';
+
 interface IState {
   publishToController: { digest?: string; image: string; tag?: string };
   repo: ContainerRepositoryType;
@@ -29,6 +32,7 @@ interface IState {
   redirect: string;
   editing: boolean;
   alerts: AlertType[];
+  showDeleteModal: boolean;
 }
 
 export interface IDetailSharedProps extends RouteComponentProps {
@@ -48,6 +52,7 @@ export function withContainerRepo(WrappedComponent) {
         redirect: undefined,
         editing: false,
         alerts: [],
+        showDeleteModal: false,
       };
     }
 
@@ -56,6 +61,11 @@ export function withContainerRepo(WrappedComponent) {
     }
 
     render() {
+      if (this.state.redirect === 'list') {
+        return (
+          <Redirect push to={formatPath(Paths.executionEnvironments, {})} />
+        );
+      }
       if (this.state.redirect === 'activity') {
         return (
           <Redirect
@@ -119,9 +129,19 @@ export function withContainerRepo(WrappedComponent) {
         >
           {t`Use in Controller`}
         </DropdownItem>,
+        <DropdownItem
+          key='delete'
+          onClick={() => {
+            this.setState({ showDeleteModal: true });
+          }}
+        >
+          {t`Delete`}
+        </DropdownItem>,
       ].filter((truthy) => truthy);
 
-      const { publishToController } = this.state;
+      const { alerts, repo, publishToController, showDeleteModal } = this.state;
+      let selectedItem = new ExecutionEnvironmentType();
+      selectedItem.name = repo.name;
 
       return (
         <React.Fragment>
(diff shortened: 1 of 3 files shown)


himdel:
@MilanPospisil I updated the title to actually say what the PR does, can you fix the description too please?
Be specific about what this pr, is actually changing add a screenshot, and mention the Jira issue.

MilanPospisil:
> @MilanPospisil I updated the title to actually say what the PR does, can you fix the description too please? Be specific about what this pr is actually changing, add a screenshot, and mention the Jira issue.

This is from Jira :-) So we better fix the description in Jira to match the PR exactly?

himdel:
No need, the Jira description describes the problem (ie. the list and detail screen don't have the same actions available),
but the PR description should also describe the solution (ie. this is adding a missing Delete action to the detail screen).

Hope it helps :)

src/containers/execution-environment-detail/base.tsx
```diff
@@ -119,9 +129,19 @@ export function withContainerRepo(WrappedComponent) {
...
+            this.setState({ showDeleteModal: true });
+          }}
+        >
+          {t`Delete`}
+        </DropdownItem>,
       ].filter((truthy) => truthy);
 
-      const { publishToController } = this.state;
+      const { alerts, repo, publishToController, showDeleteModal } = this.state;
+      let selectedItem = new ExecutionEnvironmentType();
```
himdel, line 143: this is surprising .. instatiating a Type doesn't sound right
.. ah, it only works becuse the type is defined as a class, not a type for some reason

Either way, I don't think you should ever create a new one .. can you pass the existing item instead?
Or is this supposed to replace the `null` value?

himdel: (is this just `selectedItem={repo}`?)

src/containers/execution-environment-detail/delete-execution-enviroment-modal.tsx
```diff
@@ -0,0 +1,97 @@
...
+import * as React from 'react';
+import { ExecutionEnvironmentType, ExecutionEnvironmentAPI } from 'src/api';
+import { waitForTask } from 'src/utilities';
+import { AppContext } from 'src/loaders/app-context';
+import { DeleteModal } from 'src/components/delete-modal/delete-modal';
+
+import { Checkbox } from '@patternfly/react-core';
+
+interface IState {
+  isWaitingForResponse: boolean;
```
himdel, line 11: dead one (replaced by `isDeletionPending`?)

src/containers/execution-environment-detail/base.tsx
```diff
@@ -119,9 +129,19 @@ export function withContainerRepo(WrappedComponent) {
         >
           {t`Use in Controller`}
         </DropdownItem>,
+        <DropdownItem
```
himdel, line 132: missing rbac.. the other screen is using `this.context.user.model_permissions.delete_containerrepository &&`
￼

src/containers/execution-environment-detail/base.tsx
MilanPospisil: Those are two different classes, I create one instance and then I pass the name to it. Maybe selectedItem should be only string, because only name is used. But, actualy, 'selectedItem = repo' works, compiler does not throws error... Oh, now application crashed :-D

MilanPospisil: So I changed selectedItem to string, in my custom modal component, only attribute name is used, not any other info, so string is better than class.

src/containers/execution-environment-detail/delete-execution-enviroment-modal.tsx
MilanPospisil: Right. This is some dead code, because I copied DeleteUserModal and start impementing DeleteExecutionEnviroment modal from it and I forgot to remove this.

himdel:
you seem to be adding a 98M `core` file.. remove please :)

src/containers/execution-environment-detail/base.tsx
```diff
@@ -37,7 +42,10 @@ export interface IDetailSharedProps extends RouteComponentProps {
 
 // A higher order component to wrap individual detail pages
 export function withContainerRepo(WrappedComponent) {
-  return class extends React.Component<RouteComponentProps, IState> {
+  let wrapper_class = class extends React.Component<
+    RouteComponentProps,
+    IState
+  > {
```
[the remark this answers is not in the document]
himdel: (but the current approach is fine if putting this inside the class fails for some reason)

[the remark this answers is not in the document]
MilanPospisil: It works :-)

himdel:
I think this is ready :) :+1: 

You may also want to remove these unused remnants before merging though:

```
src/containers/execution-environment-list/execution_environment_list.tsx
62:  confirmDelete: boolean;
63:  isDeletionPending: boolean;
99:      confirmDelete: false,
100:      isDeletionPending: false,
```

Tested in the UI, works for me :)

himdel (approved):

PR #1373, opened 2021-12-14
Remove legacy routing code

Hyperkid123:
Fixes initialization issues in the latest chrome release.

More details in slack conversation

@himdel @ZitaNemeckova commit `185e202bd2a45e2754b2e55c615bf03ca256fc5a` is the one that fixes the issue and should be cherry-picked all the way to prod-stable build. The other one is just some clean-up.


src/loaders/insights/Routes.js
```diff
@@ -153,150 +122,62 @@ InsightsRoute.propTypes = {
...
-          some(Paths, (p) => p === path) ? null : (
-            <Redirect push to={Paths.notFound} />
-          )
-        }
-      />
-    </Switch>
+    /**
+     * TODO: Add global loader component to the fallback
+     */
+    <Suspense fallback={null}>
```
himdel, line 130: shouldn't we still fallback to `LoadingPageWithHeader`?

Hyperkid123: @himdel if that is your loader then yes.

himdel: yeah, that's the one hardcoded in the old asyncComponent :)

himdel:
(cypress failure unrelated, pr-checks is failing on `src/loaders/insights/Routes.js` vs prettier)

himdel:
Cc @ZitaNemeckova any chance your insights mode is working? 

I think this PR is ready, would approve and merge :+1: ... but when testing, `qa.foo`, `ci.foo` and `stage.foo` come completely blank, while `prod.foo` has the menu and top nav while only the automation hub parts are blank :) .. and I get the same for master as well.

ZitaNemeckova:
@himdel Insights mode gives me 403 on every page :/

himdel:
Yeah, agreed, we'll need to fix the local setup, but not blocking this PR for it :) We can test on stage.

PR #1195, opened 2021-11-04
Empty search correct

MilanPospisil:
Compound filter had problem that it accepted blank string (string with only empty spaces). This PR disables filter button in this situation. This corrects the issue for many containers that uses compound filter.

PR #1376, opened 2021-12-14
[stable-4.4] Fix empty string filter (#1195)

himdel:
Backports #1195 - Disable CompoundFilter search for empty-looking strings, as well as just "".

Fixes AAH-820

manual backport because state vs props - changed in #832 which is not backported yet


himdel:
(waiting for https://github.com/ansible/ansible-hub-ui/pull/1195#issuecomment-994247185 .. this may need more fixing)

PR #1380, opened 2021-12-16
validate commit has a jira issue

jctanner:
Per https://issues.redhat.com/browse/AAH-1106 all commits must have a JIRA issue and every commit must have a CHANGELOG entry.

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -19,6 +19,7 @@ jobs:
       - name: Run script to validate commits for both pull request and a push
         env:
           GITHUB_PR_COMMITS_URL: ${{ github.event.pull_request.commits_url }}
+          GITHUB_USER: ${{ github.event.pull_request.user.login }}
           START_COMMIT: ${{ github.event.before }}
           END_COMMIT: ${{ github.event.after }}
         run: |
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -5,6 +5,33 @@ on:
     branches: [ 'master', 'stable-*' ]
 
 jobs:
+
+  check_commit:
+
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.after }}  # for PR avoids checking out merge commit
+          fetch-depth: 0  # include all history
+
+      - name: Run script to validate commits for both pull request and a push
+        env:
+          PY_COLORS: '1' 
+          ANSIBLE_FORCE_COLOR: '1' 
+          GITHUB_PULL_REQUEST: ${{ github.event.number }}
+          GITHUB_PULL_REQUEST_BODY: ${{ github.event.pull_request.body }}
+          GITHUB_BRANCH: ${{ github.head_ref }}
+          GITHUB_REPO_SLUG: ${{ github.repository }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


awcrosby:
When we added the `galaxy_ng` `validate_commit_message_custom.py` to `galaxy-importer` - in order to automate the need to remember to keep them in sync, I added a comparison script - if you think that is useful it can be done here: 
https://github.com/ansible/galaxy-importer/blob/master/.github/workflows/galaxy_ng_commit_check.yml#L25-L35

jctanner:
@awcrosby we could also simply wget the script from the galaxy_ng repo instead of having to maintain it in multiple spots. What do you think about that?

awcrosby:
@jctanner yeah that sounds good!

.github/workflows/pr-checks.yml
```diff
@@ -5,6 +5,33 @@ on:
...
+      - name: Run script to validate commits for both pull request and a push
+        env:
+          PY_COLORS: '1' 
+          ANSIBLE_FORCE_COLOR: '1' 
+          GITHUB_PULL_REQUEST: ${{ github.event.number }}
+          GITHUB_PULL_REQUEST_BODY: ${{ github.event.pull_request.body }}
+          GITHUB_BRANCH: ${{ github.head_ref }}
+          GITHUB_REPO_SLUG: ${{ github.repository }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_CONTEXT: ${{ github.event.pull_request.commits_url }}
```
himdel, line 28: none of these env vars seem to be used? I'm only seeing references to `GITHUB_PR_COMMITS_URL`, `START_COMMIT` and `END_COMMIT` in the code

.github/workflows/pr-checks.yml
```diff
@@ -5,6 +5,33 @@ on:
...
+          GITHUB_PR_COMMITS_URL: ${{ github.event.pull_request.commits_url }}
+          START_COMMIT: ${{ github.event.before }}
+          END_COMMIT: ${{ github.event.after }}
+        run: |
+          curl https://raw.githubusercontent.com/ansible/galaxy_ng/master/.ci/scripts/validate_commit_message_custom.py | python
```
himdel, line 33: this needs addiitional logic to ignore pull requests made by dependabot , those will not have `Fixes:` nor `No-Issue`

himdel: And I think we also need the same treatmeant for patchback .. example https://github.com/ansible/ansible-hub-ui/pull/1369 ... the PR does link to the issue, but the commit message doesn't

himdel: I think something like

```
if: ${{ github.event.pull_request.user.login }} != "dependabot[bot]" && ${{ github.event.pull_request.user.login }} != "patchback[bot]"
```

should fix it, I think :)

himdel:
hm, @awcrosby I think the script has a bug now - it requires the commit with `Fixes: AAH-123" and the one with `CHANGES/123.misc` to be the same commit *even* in PR checks.

Can we fix that? They become a single commit anyway on squash, so let's allow the message in separate commits.

.github/workflows/pr-checks.yml
jctanner: https://github.com/ansible/galaxy_ng/pull/1088

jctanner:
waiting for https://github.com/ansible/galaxy_ng/pull/1088 to get merged.

himdel (approved):
https://github.com/ansible/galaxy_ng/pull/1088 is in,
I think this is ready :+1: :)

PR #1402, opened 2021-12-21
Disable enter on submit empty filter string, disable empty filter from params URL

MilanPospisil:
This PR corrects previous PR, which was incomplete:

https://github.com/ansible/ansible-hub-ui/pull/1195
AAH-820

It disables enter (submit) when user fills only empty filter text. It also does not add empty filter parameters from URL to container component state.


himdel (approved):
LGTM, works for me :+1: :)

When adding `?keywords=%20` manually, it no longer shows up in applied filter, and it no longer gets passed to the API :+1:

himdel:
(will backport together with #1195, in https://github.com/ansible/ansible-hub-ui/pull/1376)

PR #1459, opened 2022-01-03
Collections list test delete

MilanPospisil:
Added deletion in Collections list test using galaxykit.

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -26,7 +26,7 @@ jobs:
 
     - name: "Install galaxykit dependency"
       run: |
-        pip install galaxykit
+        pip install git+https://github.com/ansible/galaxykit.git
 
     - name: "Set env.SHORT_BRANCH, env.GALAXY_NG_COMMIT"
       run: |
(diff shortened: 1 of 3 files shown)


himdel:
Code looks good, but `galaxykit namespace delete my_namespace` seems to be failing for some reason.

If the new version of galaxykit is not released yet you may need this change:

```diff
diff --git a/.github/workflows/cypress.yml b/.github/workflows/cypress.yml
index 60b95ee..e3bd0f5 100644
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -26,7 +26,7 @@ jobs:
 
     - name: "Install galaxykit dependency"
       run: |
-        pip install galaxykit
+        pip install git+https://github.com/ansible/galaxykit.git
 
     - name: "Set env.SHORT_BRANCH, env.GALAXY_NG_COMMIT"
       run: |
```

(looks like https://github.com/ansible/galaxykit/pull/20 was merged Dec 23, but the [latest release](https://pypi.org/project/galaxykit/) is from Dec 1st)

himdel (approved):
LGTM :+1: 

And I think we can keep pip installing from github on master without problems,
only `stable-*` branches should probably depend on released versions.

PR #1463, opened 2022-01-04
eslintrc: enable more default rules

himdel:
Enabling these default rules: `@typescript-eslint/no-explicit-any`, `@typescript-eslint/no-var-requires`, `react/display-name`, `react/jsx-key`, `react/no-unescaped-entities`

`no-explicit-any` warns about 25 more uses of `any`, but those shouldn't make PRs go red (https://github.com/ansible/ansible-hub-ui/runs/4698268100?check_suite_focus=true#step:4:92), and can be fixed separately

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -41,7 +41,6 @@
 
     /// Rules we're disabling until they can be fixed, or disabled permanently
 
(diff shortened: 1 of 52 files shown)


MilanPospisil (approved):
LGTM 👍

PR #1465, opened 2022-01-05
typescript-eslint/no-explicit-any: add eslint-disable-next-line comments

himdel:
Follow-up to #1463 

We have a bunch of quite valid uses of explicit any, we can either go through the effort of explicitly typing these, or disable the whole rule, or mark the remaining ones as ok, while still warning about new ones.

Going with the last option for now.

### DIFF
--- a/src/api/active-user.ts
+++ b/src/api/active-user.ts
@@ -8,6 +8,7 @@ class API extends HubAPI {
     super();
   }
 
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   getUser(): Promise<any> {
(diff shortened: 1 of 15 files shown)


brumik:
I feel like some of these could be typed pretty easily, or maybe marked with unknown - like in the promises from the API. However I am not too familiar with the codebase yet. Maybe we could try to type the easier ones before ignoring them?

himdel:
I agree we should type whatever we can, but  the easy ones were already done in https://github.com/ansible/ansible-hub-ui/pull/1463/commits/773cb47d53af7b43d49f4201f33b0029fb353811 :)

If you know how to type any of these, please feel free to share :)

himdel:
Closing for #1493, thanks :)

PR #1489, opened 2022-01-10
Fix indent of the groupped toolbar

brumik:
Issue: https://issues.redhat.com/browse/AAH-822

### Concern about the change
If we edit these pages (merge this PR), then the toolbar indent is not matching to the "Approval" and "User access" page's toolbar (Where they have this indent).

### Before
image

### Result
Screenshot from 2022-01-10 10-43-08
Screenshot from 2022-01-10 10-43-00

### DIFF
--- a/src/components/patternfly-wrappers/toolbar.tsx
+++ b/src/components/patternfly-wrappers/toolbar.tsx
@@ -68,7 +68,7 @@ export class Toolbar extends React.Component<IProps, IState> {
     return (
       <ToolbarPF>
         <ToolbarContent>
-          <ToolbarGroup>
+          <ToolbarGroup style={{ marginLeft: 0 }}>
             <ToolbarItem>
               <InputGroup>
                 <TextInput
--- a/CHANGES/822.misc
+++ b/CHANGES/822.misc
@@ -1,0 +1,1 @@
+Removed the extra spacing from the headers
--- a/src/components/collection-list/collection-filter.tsx
+++ b/src/components/collection-list/collection-filter.tsx
@@ -64,7 +64,7 @@ export class CollectionFilter extends React.Component<IProps, IState> {
     return (
       <Toolbar>
         <ToolbarContent>
-          <ToolbarGroup>
+          <ToolbarGroup style={{ marginLeft: 0 }}>
             <ToolbarItem>
               <CompoundFilter
                 inputText={this.state.inputText}
--- a/src/components/repo-selector/repo-selector.tsx
+++ b/src/components/repo-selector/repo-selector.tsx
@@ -47,6 +47,7 @@ export class RepoSelector extends React.Component<IProps, IState> {
         <FlexItem>
           <InputGroup>
             <InputGroupText
+              style={{ paddingLeft: 0 }}
               variant='plain'
               className='hub-input-group-text-no-wrap'
             >


brumik:
>If we edit these pages (merge this PR), then the toolbar indent is not matching to the "Approval" and "User access" page's toolbar (Where they have this indent).

Taufique Rahman approved and we may raise this issue to the PF team.

Good to merge

src/components/patternfly-wrappers/toolbar.tsx
```diff
@@ -3,7 +3,6 @@ import { t } from '@lingui/macro';
 
 import {
   Toolbar as ToolbarPF,
-  ToolbarGroup,
```
himdel, line 6: This may be the wrong way of fixing this one, as we do have multiple toolbar groups here.

I believe the other places were fixed by using 

```css
.pf-c-toolbar__content {
  padding-left: 0px;
}
```

himdel:
> If we edit these pages (merge this PR), then the toolbar indent is not matching to the "Approval" and "User access" page's toolbar (Where they have this indent).

Those should also be fixed, I believe the goal was to fix all toolbars to align with the header.
(The issue is from July, we didn't have Execution Environments nor Task Management back then)

But up to you if you want to keep this one short and deal with the rest separately :)

src/components/patternfly-wrappers/toolbar.tsx
brumik: Not here. I fixed all the toolbar problems (the pf class was overwritten in a random scss file). This one I removed because they were wrapping one toolbar item. So either we put those items in one group (which I don't think is in our intention) or the group is not needed.

brumik:
@himdel I looked into that issue and found the global solution for it. It was a random pf class overwrite in a scss affecting every page. Ideally we should not overwrite pf classes in scss. (In fact I am a big fan of styled-components since it encapsulates the styling and makes these kinds of problems disapear :)

Anyway: I feel like your concerns on this PR are answered. Another round of review?

himdel:
OK, looks like we need to keep the class override for now, this now breaks the alignment on those screens...

brumik:
I modified most of the toolbars to:
1. Keep the current looks and the fixed paddings
2. Use minimal scss and custom styling and follow the PF library as close as possible.

***All the modified toolbars are visible in the screenshots under.***


Followup task could be:
1. make the chips under the toolbar use the default patternfly builtin functions instead of hacking it under the toolbar by ourselves. (If I get the time I would like to work on this.)

Screenshot from 2022-01-13 11-32-25
Screenshot from 2022-01-13 11-32-15
Screenshot from 2022-01-13 11-31-59
Screenshot from 2022-01-13 11-31-45
Screenshot from 2022-01-13 11-31-29
Screenshot from 2022-01-13 11-31-15

brumik:
If this is good I fix the tests next monday.

src/components/patternfly-wrappers/toolbar.tsx
brumik: I don't think this is an UX breaking change. That's like a 5px diff and there is still enough space between the two toolbar items. Overwriting a pf class is not worth it in my opinion, especially that keeping to pf makes us more in line with the rest of the platform.

himdel: https://www.patternfly.org/v4/components/toolbar/#adjusting-item-spacers should help :)

himdel:
So, the original change was really close, this now seems overly ambitious for that particular bug, especially since we may want to backport the fix.

Can you split this into 2 separate PRs please? Or just revert back to the original fix for Collections, without affecting the rest of the appliaction.

---

More specifically, this now seems to be introducing quite some extra whitespace both layouts...

| |master|here|
|-|-|-|
|empty|||
|filtering|||

And additionally breaking `AppliedFilters` alignment in content layouts... (screenshots from Remote Registries)

| |master|here|
|-|-|-|
|empty|||
|filtering|||

brumik:
@himdel Ill do a quick fix then and leave this on the side. 

The filter spacing can be not so easily fixed by using the built in filters in the PF Toolbar component. 
When I do these changes I can redo that part too, but that's a bit bigger refactor (I am happy to undertake but may take up more time so I need somebody approval that I can work on it)

src/components/patternfly-wrappers/toolbar.tsx
brumik: I'm all in using the PF spacer and component composition to achieve the same look :)

brumik:
For the current state look at the PR description.

himdel (approved):

PR #1493, opened 2022-01-10
Types some of the any types and ignores rest

brumik:
Improves upon #1465 with adding some easily type-able types

### DIFF
--- a/src/api/active-user.ts
+++ b/src/api/active-user.ts
@@ -8,6 +8,7 @@ class API extends HubAPI {
     super();
   }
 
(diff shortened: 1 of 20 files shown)


brumik:
The others could be typed too probably, but my understanding of the product is too low for it. And I set a 40 min time-frame on this to not to waste time.

himdel (approved):
Works for me :+1: 

Some of the typing info may be too specific now, reflecting how we use those functions instead how we could, but that's something we can fix when we need to :)

PR #1494, opened 2022-01-11
Added l10n translation cypress tests

brumik:
Issue: AAH-1189
https://issues.redhat.com/browse/AAH-1189

### DIFF
--- a/test/cypress/integration/l10n.js
+++ b/test/cypress/integration/l10n.js
@@ -1,0 +1,59 @@
+// Note up to date there is no translation for any string in 'es' and 'nl'.
+const languageCheckHelper = (url, selector) => (language, message) => {
+  cy.visit(url, {
+    onBeforeLoad(win) {
+      if (language) {
+        Object.defineProperty(win.navigator, 'languages', {
+          value: [language],
+        });
+      }
+    },
+  });
+
+  cy.get(selector).should('contain.text', message);
+};
+
+describe('Localization tests with the t`String` format', () => {
+  const helper = languageCheckHelper('/ui/repositories', 'h1');
+
+  beforeEach(() => {
+    cy.login();
+  });
+
+  const translations = {
+    en: 'Repo Management',
+    fr: 'Gestion de référentiel',
+    ja: 'リポジトリー管理',
+    zh: '仓库管理',
+  };
+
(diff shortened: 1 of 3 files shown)


test/cypress/integration/l10n.js
```diff
@@ -0,0 +1,61 @@
...
+  beforeEach(() => {
+    cy.login();
+  });
+
+  const translations = [
+    'Repo Management',
+    'Gestion de référentiel',
+    'リポジトリー管理',
+    '仓库管理',
+  ];
```
himdel, line 30:
```suggestion
  const translations = {
    en: 'Repo Management',
    fr: 'Gestion de référentiel',
    ja: 'リポジトリー管理',
    zh: '仓库管理',
  };
```

so that we don't have to synchronize the list of languages between 3 arrays in different parts of the file?

(and `helper(key, translations[key])` is probably a bit clearer too)

himdel:
Overall looks great :+1: 

I wonder if we should also add a test for Namespaces (once https://github.com/ansible/ansible-hub-ui/pull/1495 is in), given these would also cover testing delayed strings.

test/cypress/integration/l10n.js
brumik: I think in tests, especially this straightforward one it does not really matter, but if you want it its a 1min update. Then I will remove the top level language list too as it will be "duplicate"

brumik:
@himdel We can merge this first I think and then we can add the tests in #1495 keeping the whole issue in one PR. (It seems like that PR is going to be merged slower than this could be)

himdel:
> We can merge this first I think and then we can add the tests in #1495 

I'm fine with that :+1: 

Just remember to keep AAH-1189 open until #1495 is merged with tests.

However that may be a moot point, this is also still missing the API stirng tests.

brumik:
The API tests are still WIP, still tracking down why I cannot see them properly in manual testing :)

brumik:
@himdel The API translations in POST request when catching an error is not working, and could not find any other places to test API translations with GET. 

I suggest we merge these tests in, then create separate tests for the API translations, which we could track in a separate ticket and all.

^ After discussing with @ jctanner

himdel (approved):

himdel:
Agreed, thanks, approved :)

PR #1495, opened 2022-01-11
Fixed the localization of the Namespaces string

brumik:
Issue: AAH-1069
https://issues.redhat.com/browse/AAH-1069

**Before**
Screenshot from 2022-01-11 14-40-40

**After**
Screenshot from 2022-01-14 10-16-51
Screenshot from 2022-01-14 10-16-41

### DIFF
--- a/src/containers/namespace-list/namespace-list.tsx
+++ b/src/containers/namespace-list/namespace-list.tsx
@@ -153,7 +153,7 @@ export class NamespaceList extends React.Component<IProps, IState> {
       );
     }
 
-    const title = namespaceBreadcrumb.name;
+    const title = namespaceBreadcrumb().name;
     const titleLowerCase = title.toLowerCase();
     const search = filterOwner
       ? t`Search my namespaces`
--- a/CHANGES/1069.misc
+++ b/CHANGES/1069.misc
@@ -1,0 +1,1 @@
+Fixed the namespace localization
--- a/src/containers/collection-detail/collection-content.tsx
+++ b/src/containers/collection-detail/collection-content.tsx
@@ -42,7 +42,7 @@ class CollectionContent extends React.Component<
     }
 
     const breadcrumbs = [
-      namespaceBreadcrumb,
+      namespaceBreadcrumb(),
       {
         url: formatPath(Paths.namespaceByRepo, {
           namespace: collection.namespace.name,
--- a/src/containers/collection-detail/collection-dependencies.tsx
+++ b/src/containers/collection-detail/collection-dependencies.tsx
@@ -86,7 +86,7 @@ class CollectionDependencies extends React.Component<
     }
 
     const breadcrumbs = [
-      namespaceBreadcrumb,
+      namespaceBreadcrumb(),
       {
         url: formatPath(Paths.namespaceByRepo, {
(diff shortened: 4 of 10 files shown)


src/containers/namespace-list/namespace-list.tsx
```diff
@@ -153,7 +153,7 @@ export class NamespaceList extends React.Component<IProps, IState> {
       );
     }
 
-    const title = namespaceBreadcrumb.name;
+    const title = t`Namespaces`;
```
himdel, line 156: This won't work. We need the partners vs namespaces logic.

Which means it's probably broken becuase https://github.com/ansible/ansible-hub-ui/pull/1295 is not merged yet.

brumik: Okey, I would like to understand this logic and how to test. Then let's wait for that PR and see if it fixes it.

himdel: Oh, sorry, it's the same **kind** of problem #1295 is fixing, but it's not actually addressing this specific problem.

The bug is this: constant expressions are evaluated before lingui gets a chance to run, so any ``export const s = t`Foo`;`` is never translated, because it always translates to english.

One possibly fix would be to make `namespaceBreadcrumb` into a function, so that this gets evaluated after lingui is initialized.

But the fix we've settled on is to separate string extraction from the actual translation (in proper gettext terms, instead of doing `string s = _("Foo"); whatever(s);` we do `string s = N_("Foo"); whatever(_(s));`) - it's a bit cumbersome with lingui, but:

```js
import { defineMessage } from '@lingui/macro';
import { i18n } from '@lingui/core';

const s = defineMessage({ message: `Published` }) // mark string for translation

whatever( i18n._(s) ); // translate later
```

brumik: const title = t`Namespaces`; seems to work here, so I am little confused about the problem with constants. Or is this only applicable to the `export const = NonFunctionObj` expressions? And even so: the `t(literal)` is a function.

himdel: This is about constants at top-level. Top-level code always runs sooner than lingui setup, and `namespaceBreadcrumb` is defined top-level.

`` t`Namespaces` `` only work for standalone mode, as I mentioned in the first comment, we do need to keep the `Namespaces` vs `Partners` logic here.

So this should be `const title = i18n._(namespaceBreadcrumb.name);`

himdel:
My apologies, but I asked you very specificically to do it as agreed in https://github.com/ansible/ansible-hub-ui/pull/1495#discussion_r782588312 .

This is not acceptable, I'm taking this fix, closed.
Cc @chouseknecht

himdel:
Any communication issues aside, this does fix the issue on all affected screens, and we *can* address any postponed translation issues separately, so maybe the best way forward is ..forward, approving :+1:

himdel (approved):

brumik:
@himdel I am sorry for the miscommunication, I though that's a general idea what you described in the comment as I did not understood it how it should work.

brumik:
Closing in favor of https://github.com/ansible/ansible-hub-ui/pull/1525

PR #1469, opened 2022-01-07
Unify success alerts

ShaiahWren:
See issue: https://issues.redhat.com/browse/AAH-1236

Assignment: Update success alerts to fit modular format outlined in the [documentation](https://docs.google.com/document/d/1MnTmoB64sQQTReynWdiUhtduRY-4WxokY22sdxKwLIk/edit#). Also see [additional documentation](https://docs.google.com/document/d/1CnC37CYwIafjIbkqSWTKOJSvnSSTrqUR6vTiK2leMXg/edit) for general messages. 

This pr updates all alerts messages across the application according to the above documentation. In some cases, missing alerts were added.

PR #1500, opened 2022-01-11
cypress workflow: ensure cypress is running with up-to-date translations

himdel:
Noticed in https://github.com/ansible/ansible-hub-ui/pull/1469,
if you change translations in a PR, github actions cypress will not pick up the updates until gettext extract & compile happens,
making sure this always happens before tests.


ShaiahWren (approved):
Looks like this should do the trick. Approved :)

PR #1502, opened 2022-01-13
Test approval dashboard

MilanPospisil:
Test for Approval Dashboard.

https://issues.redhat.com/browse/AAH-628


brumik (changes requested):
Except the log this LGTM.

Two notes: 
1. the tests names usually should spell something like `is('should do this and that')` so you can read it like a proper english (I see that in this repo its not used - yet :)
2. Instead of using class names we could use rather `data-cy` tags, that can futureproof the tests in case we decide to change the class name or change the component to something else (like the body is a custom implemented `Card` component from PF right now). I think if you could add those tags and in test use them it would be great. 

Only the left out log is a blocker here, everything else is a suggestion.

test/cypress/support/commands.js
```diff
@@ -395,6 +395,7 @@ Cypress.Commands.add('galaxykit', {}, (operation, ...args) => {
     args.length >= 1 && typeof args[args.length - 1] == 'object'
       ? args.splice(args.length - 1, 1)[0]
       : [];
+  cy.log('galaxykit ' + operation + ' ' + args);
```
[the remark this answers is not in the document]
brumik: https://github.com/ansible/ansible-hub-ui/pull/1502/files#diff-b09b4253c613a69992d44e32fbd39ffe4c8a219c17024fdf7c1296e887aa2662R66

These two logs too ^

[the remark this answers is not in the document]
MilanPospisil: Yes, the second log is mistake. I will repair the PR, but today I must solve other sporadic failing tests. But first log should be here for better logging, I added the info to the PR description. You are also right with data-cy, I will try to use them more :-)

test/cypress/integration/approval_dashboard_list.js
```diff
@@ -0,0 +1,176 @@
+import { range, sortBy } from 'lodash';
+
+describe('Approval Dashboard list tests for sorting, paging and filtering', () => {
+  let items = [];
```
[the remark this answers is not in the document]
MilanPospisil: I would actualy not rather goes this way. It seems it is not that easy - there is callback which sorts the array, and sort method returns new array, so there will be overhead of clearing orginaly passed array and filled it with the new sorted array. I think closure works also fine and it is readable, do you still want this change?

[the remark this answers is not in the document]
brumik: I read somewhere that every let can be converted to const in ES6. Anyways, I did not wanted you to change in the first place, it is still readable so feel free to leave this here.

test/cypress/integration/approval_dashboard_list.js
```diff
@@ -0,0 +1,180 @@
...
+      'approval_dashboard_collection_test_additional1',
+    );
+    cy.galaxykit(
+      '-i collection upload',
+      'approval_dashboard_namespace_test_additional_data',
+      'approval_dashboard_collection_test_additional2',
+    );
+  }
+
+  function loadData() {
```
brumik, line 28: Can you please replace the two `var` in this function with `const` or `let`? In this case it has no effect as both the `var` and `let` is encapsulated by the function's scope, but getting used to not using `var` is always a good idea :)

test/cypress/integration/approval_dashboard_list.js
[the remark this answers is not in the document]
MilanPospisil: Ok. But I dont think that every let can be converted to const, because you cant assign to const, but you can assign to let, this is what Im doing - assigning variable again to the result of sort operation.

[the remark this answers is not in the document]
brumik: You can replace `then` with `await` and `forEach` with `map` and your function can return a new variable instead of assinging to a higher scope variable. This has memory effects, but it is used in react since the rerender only happening when the object's reference is changing, which is not (always) happening if you are mutating an object (like pushing into an array).

Also in JS you actually can change arrays and objects even when they are constants but it is NOT recommended.
So this would work without any issue:
```js
const c = [];
c.push('s');
console.log(c); // ['s']
```

But as I said here is no real need to change it.

brumik (approved):

src/containers/certification-dashboard/certification-dashboard.tsx
```diff
@@ -126,7 +126,7 @@ class CertificationDashboard extends React.Component<
           <EmptyStateUnauthorized />
         ) : (
           <Main className='hub-certification-dashboard'>
-            <section className='body'>
+            <section className='body' data-cy='body'>
```
himdel, line 129: You *may* be misunderstanding the guideline about using `data-cy` here...

It's OK, no need to fix anything, but adding `data-cy=body` to a section which already has a class body seems ..useless.
The point is not to be adding classes just for cypress, and adding data-cy instead of relying on more complex selectors in tests.

The same goes for `<tr data-cy="table_row">`, again, no real harm, but that's exactly what `tr` already means, no need to add a second selector.

test/cypress/integration/collections_list.js
```diff
@@ -35,10 +30,6 @@ describe('Collections list Tests', () => {
     });
   });
 
-  after(() => {
```
himdel, line 38: plese don't remove these

test/cypress/integration/imports_filter.js
```diff
@@ -22,10 +15,6 @@ describe('Imports filter test', () => {
     cy.visit('/ui/my-imports?namespace=filter_test_namespace');
   });
 
-  after(() => {
```
himdel, line 25: same here

test/cypress/integration/imports_filter.js
```diff
@@ -89,12 +78,7 @@ describe('Imports filter test', () => {
     ).as('wait');
     cy.contains('a', 'Completed').click();
 
-    cy.get('[data-cy="import-list-data"]')
```
himdel, line 92: For future reference, this should definitely have been a second PR.

himdel:
This is also not ensuring we're running with `GALAXY_REQUIRE_CONTENT_APPROVAL: true`, as described in https://issues.redhat.com/browse/AAH-628?focusedCommentId=19226791&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-19226791 so this can't actually be testing approval - the main feature the screen exists for :(

src/containers/certification-dashboard/certification-dashboard.tsx
brumik: The body data-cy was suggested by me. It is essentially a `Card` from pf what the body implements and AFAIK it is good to not get dependent with the cypress on the implementation, but rather the UI. So in this case I thought that it is a good idea to add the `body` and a `data-cy` selector(s) since the element could be replaced with a component where we don't necessarily want to add the `classname="body"`. (Which you also recommend in your last sentence - adding data-cy to component roots :)

MilanPospisil: Classes are roughly the same as data-cy, because you can even use selectors in css over attributes, right? After all, cypress uses css selectors. So I also think it is a bit overhead to add data-class instead of classes. But I agree with Brumik that data-cy is better over element selector (tr) for example, it solves the situation when component is reworked and another elements are used, then they can use the same data-cy. But if we dont ever intent replace tr in code, we should be probably fine. But overall, I dont see any of these two solutions (Martins and Brumiks) as superior in our project (Brumiks solution is better for project that tends to change), they both have their pros and cons...

himdel: You're missing the point. `body` and `table_row` and `table_header` are a bad name for a selector, regardless of class vs cy, it's nowhere specific enough to tell us the body of what, what row, what table, which header?

So adding a new data-cy=body when there already is className=body is both a waste of time *AND* introducing a technical debt by having a second thing to maintain. I would not have objected if the new selector was more descriptive than just body, but if you're just using the same name, there is no advantage to cy over classNames, *except* that we should not introduce classnames for tests.

As for tr, again .. tr litrally means table row, having a separate attribute saying the same thing does not increase flexibility in any way. 

Still, just a minor annoyance, we can fix the names in the next pass, this is not the only place :)

PR #1504, opened 2022-01-13
Fix typescript errors related to `never` and `axios`

himdel:
Follow-up to #1493

editors can't seem to handle some of the typing bits, leading to `Property 'meta' does not exist on type 'never'.ts(2339)`,
turns out `Promise<never>` is superfluous here, removing :)

and also seeing

    Argument of type '(result: AxiosResponse<any>) => void' is not assignable to parameter of type '(value: AxiosResponse<any>) => AxiosResponse<any> | PromiseLike<AxiosResponse<any>>'.
    Type 'void' is not assignable to type 'AxiosResponse<any> | PromiseLike<AxiosResponse<any>>'.ts(2345)

=> changing `http` type back to `any`.

Cc @ShaiahWren does this help? :)

### DIFF
--- a/src/api/base.ts
+++ b/src/api/base.ts
@@ -1,11 +1,12 @@
-import axios, { AxiosInstance } from 'axios';
+import axios from 'axios';
 import { Constants } from 'src/constants';
 import { ParamHelper } from 'src/utilities';
 import Cookies from 'js-cookie';
 
 export class BaseAPI {
   apiPath: string;
-  http: AxiosInstance;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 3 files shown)


ShaiahWren (approved):
Works for me!
Do we know why cypress isn't passing?

himdel:
Ah, good to hear that :)

cypress.. `The Test Runner unexpectedly exited via a exit event with signal SIGTRAP` .. that's definitely not our fault :) but restarting, I think it should go away :)

himdel:
Yup, now we're back to the usual sporadic failure, Milan is fixing that :).

PR #1525, opened 2022-01-17
Fixed the localization of the Namespaces string

brumik:
Issue: AAH-1069

**Before**
Screenshot from 2022-01-17 10-51-15

**After**
Screenshot from 2022-01-17 10-50-11

### DIFF
--- a/CHANGES/1069.misc
+++ b/CHANGES/1069.misc
@@ -1,0 +1,1 @@
+Fixed the namespace localization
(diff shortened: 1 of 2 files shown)


himdel (approved):
Both strings are marked for translation, works for me :+1: 

I'm slightly worried that using `` t`..` `` + `i18n._` could lead to double translation if the `t` bit ever starts translating that, but we can use `defineMessage` once all callers are converted.

Thanks :)

PR #1526, opened 2022-01-19
Added insights dev setup standalone

brumik:
Issue: AAH-1262

This commit changes up the development setup for insights, now it
is using keycloak and insights-standalone to spin up the local
development mode.

I made a copy of the webpack file and removed the minimal amount of variables that were not used in the development mode, then added the insights standalone and some extra options to not to compile PF. 

***Important***: The insights URL is `http://localhost:8002/beta/ansible/automation-hub`. It does not work without the `beta` since the pendo tries to load then.

Tested:
**Local galaxy backend running on port 5001**
* `npm start` --> insights version
* `npm run start-standalone` --> standalone version
* local backend running the UI in a container (`compose up` with UI path) --> standalone version

**Production version**
* no changes to the production script

## Directions to the most stable setup I know yet
1. Pull this PR
2. UI: `sudo rm -rf dist` --> the docker is the owner of the compiled files if you used it from galaxy, if not deleted the build will fail on permissions
3. UI: `npm ci`
4. UI: `npm start` (this takes a while on first startup to download the images)
5. GALAXY: in the compose.env set `COMPOSE_PROFILE=insights` and comment out `#ANSIBLE_HUB_UI_PATH='~/Documents/ansible-hub-ui'`
6. GALAXY: general rebuild and cleaning: `make docker/build && make docker/resetdb && make docker/migrate && make docker/loaddata`
7. `compose up` - start the api backend
8. If all done visit: `http://localhost:8002/beta/ansible/automation-hub` --> note the beta
9. Login info to the keycloak is: admin/admin


config/webpack.dev.base.config.js
```diff
@@ -0,0 +1,180 @@
...
+
+  const isStandalone = customConfigs.DEPLOYMENT_MODE !== 'insights';
+
+  // config for HtmlWebpackPlugin
+  const htmlPluginConfig = {
+    // used by src/index.html
+    applicationName: customConfigs.APPLICATION_NAME,
+    targetEnv: customConfigs.DEPLOYMENT_MODE,
+
+    ...(isStandalone && {
```
brumik, line 58: Merged the htmlPluginConfig variables into one object.

Standalone mode overwrites favicon, insights not.

config/webpack.dev.base.config.js
```diff
@@ -0,0 +1,180 @@
...
+    htmlPlugin: htmlPluginConfig,
+    debug: true,
+    https: false,
+    port: 8002,
+    ...(isStandalone && {
+      // frontend-components-config 4.5.0+: don't remove patternfly from non-insights builds
+      bundlePfModules: true,
+    }),
+    ...(!isStandalone && {
+      appUrl: ['/beta/ansible/automation-hub/', '/ansible/automation-hub/'],
```
brumik, line 77: If it is an insights mode we need to define appUrl, deployment and we are using the `standalone` version to run it locally. It has inbuilt keycloak and proxies the API to to our `localhost:5001`. I could not use here the variable `API_HOST` without breaking the frontend.

config/webpack.dev.base.config.js
```diff
@@ -0,0 +1,180 @@
...
+  if (customConfigs.DEPLOYMENT_MODE === 'insights') {
+    /**
+     * Generates remote containers for chrome 2
+     */
+    plugins.push(
+      require('@redhat-cloud-services/frontend-components-config/federated-modules')(
+        {
+          root: resolve(__dirname, '../'),
+          useFileHash: false,
+          exclude: [
```
brumik, line 162: Exluding the frontend components and PF from compilation when on dev. By default insights ships React, PF, frontend-components already in production mode. 

This list does not contain React (it is more tricky than just include on this list), and this means that we can only see production react error messages when using the `insights` mode to develop.

himdel: This seems like a list that will evolve over time, any chance we take it from a default config?
(I assume this is also happening in non-dev insights mode, where it would have to be coming from the default config, right?)

package.json
```diff
@@ -90,7 +90,7 @@
         "prettier:check": "prettier -l 'src/**' 'config/**' 'test/**'",
         "prod": "NODE_ENV=production webpack serve --config custom.dev.config",
         "server:ctr": "node src/server/generateServerKey.js",
-        "start": "NODE_ENV=development webpack serve --host 0.0.0.0 --config config/insights.dev.webpack.config.js",
+        "start": "NODE_ENV=development webpack serve --config config/insights.dev.webpack.config.js",
```
himdel, line 93: Any reason for this change? Both start-pulp and start-standalone use it, I believe the container environment depends on it listening on other IP addresses.

profiles/galaxy-ng-compose.js
```diff
@@ -1,25 +0,0 @@
-/*global module*/
```
himdel, line 1: this file is referenced by `galaxy_ng/dev/insights/docker-compose-ui.yaml`, is it really dead? Do we need a correpsonding change there as well?

src/AppEntry.js
```diff
@@ -1,7 +1,6 @@
 import React from 'react';
 import { BrowserRouter as Router } from 'react-router-dom';
 import { Provider } from 'react-redux';
-import logger from 'redux-logger';
```
himdel, line 4: this would seem to remove redux-logger from all development setup, any reason to do that?

.ls-lint.yml
```diff
@@ -15,7 +15,7 @@ ls:
     .js: kebab-case | snake_case
 
   .d.ts: kebab-case
-  .js: kebab-case
+  .js: kebab-case | PascalCase
```
himdel, line 18: unless `AppEntry` is actually required by something external, let's keep the original names

config/webpack.dev.base.config.js
```diff
@@ -0,0 +1,180 @@
...
+      // frontend-components-config 4.5.0+: don't remove patternfly from non-insights builds
+      bundlePfModules: true,
+    }),
+    ...(!isStandalone && {
+      appUrl: ['/beta/ansible/automation-hub/', '/ansible/automation-hub/'],
+      deployment: 'beta/apps',
+      standalone: {
+        api: {
+          context: [customConfigs.API_BASE_PATH],
+          target: 'http://localhost:5001',
```
himdel, line 82: I don't think we can hardcode 5001 here, for the container env, this changes to 8000 (as per `profiles/galaxy-ng-compose.js`)

config/webpack.dev.base.config.js
himdel: > deployment and we are using the standalone version to run it locally.

Sorry, I'm not sure what you mean, is there a word missing?

`appUrl` seems sensible, though I wonder how this works in prod .. I was under the impression that the config is reading `package.json` `.insights.appname` to set these, do you know if that's still true? :)

Also, how does `API_HOST` break the frontend? The default value is empty? (I was under the impression that `API_HOST` is currently unused really, other than as a local debugging tool, so I wonder if I'm missing some use.)

himdel:
> I made a copy of the webpack file and removed the minimal amount of variables that were not used in the development mode, then added the insights standalone and some extra options to not to compile PF.

Looking at the diff, the change would really be just extra imports, adding

```js
    ...(!isStandalone && !isBuild && {
      appUrl: ['/beta/ansible/automation-hub/', '/ansible/automation-hub/'],
      deployment: 'beta/apps',
      standalone: {
        api: {
          context: [customConfigs.API_BASE_PATH],
          target: `http://localhost:${some api port var}`,
        },
        rbac,
        ...defaultServices,
      },
    }),
```

to config (+- the changes requested above), and ..a change to `federated-modules` config, right?

The shared config file is a good thing, as it ensures all our various configurations are still using mostly the same config, without us having to maintain multiple copies. 

(The file is also imported from the packaging repo, so the list of params needs to stay backwards compatible.)


> Important: The insights URL is http://localhost:8002/beta/ansible/automation-hub. It does not work without the beta since the pendo tries to load then.

What's a pendo?

---

Overall the change looks good once we can remove the extra bits :+1: (though I haven't tested this yet, I'm pretty sure this is not working in the container configuration given the port issues above)

.ls-lint.yml
brumik: It is required by the redhat config.

profiles/galaxy-ng-compose.js
brumik: It is dead on the platform for sure. Since chrome2 the proxy is not using this file. However I was not aware that in the galaxy-ng we are still using the spandx config.

I think it is safe to remove from there too, as the `npm start` should take care of the proxy now.

src/AppEntry.js
brumik: Could not find any redux in the app (except the `Main` component ofc) and unless you are trying to debug the app in the terminal only, the `redux extension` is a much more elaborate and powerful tool to debug redux. So for a sake of not complicating the setup more I felt like removing it is safe. We can still add back when we decide to use redux and somebody will miss it.

package.json
brumik: Can you please let me know where can this break stuff and how can I test it out?

profiles/galaxy-ng-compose.js
brumik: Once I get this working I can open a PR there too to remove the `insights-proxy` part from the docker setup

config/webpack.dev.base.config.js
brumik: Let's talk about it in person maybe?

brumik:
Pendo is a user interaction logging tool: https://www.pendo.io/ 

You can see how users interact with your page. I would say it is kinda similar to Google Analytics.

brumik:
Consolidating is doable, though I don't think its a great idea. Giving us chance to destroy prod deployment just with dev env changes. Both of them can also change without the need of the other change. And duplicating some changes I think is not a big deal in case of webpack setups.

profiles/galaxy-ng-compose.js
brumik: Okey, I looked around, the insights standalone uses docker, and running a docker inside a docker is not recommended as "it can cause low lever problems". How badly do we need it? There is one other way to go about insights dev setup and that is using the `prod` or `stage` environment, which will give us full compiled code and to develop you will need to be able to log in onto `console.redhat.com` or `console.stage.redhat.com`.

package.json
himdel: I think the steps here are .. get a Mac, then try to do same thing you're doing now :)

(though it should be common to docker-machine on any platform, I just don't think any other platform needs it, and not sure if docker desktop for win/mac has the same issue, though I wouldn't worry too much about this as long as we don't need to remove the host setting :))

.ls-lint.yml
brumik: Change the entry point instead of the name.

brumik:
TODO: Cosolidate somehow the prod/dev setup
Change the AppEntry filename back

brumik:
@himdel I could not change the `AppEntry` for some reason in devel mode, no matter if I overwrote it in the config. I managed get one base webpack config though.

himdel:
I think you may have forgotten to push the new version? Still seeing 2 base configs.

brumik:
Yes, I did forget :D

himdel:
Getting closer, but I'd still like to minimize the extra changes...

* base config changes of course,
* the README change also makes sense,
* profiles/ as well (pending a backend update)
* merging the entry files makes sense, we agreed to remove the logger, and we can live with AppEntry if needed :)

* I suppose we need the changes in insights.dev.config.js but maybe not all of them? (API_HOST shouild stay)
* we should not really be changing the standalone config here (though I like the extra comment about namespaces)

brumik:
I made those changes because it is weird to have in the config file variables that are ignored for that config. (The `API_HOST` is used in UI, but in insights you should not create a new URL but use a relative link and proxy in dev so this is why it is left to use the default value from the config - which is empty string). 

Anyways, your castle so I'll backroll the changes.

himdel:
This is mostly LGTM :+1:,
I just want to make sure this doesn't break anything (can you rebase please? - there's a new webpack config check that will fail, but it will tell us what changes in which config),
and I want to look into what it would take to keep running this from the insights mode compose script, I suspect nothing is stopping us from running the containers side by side instead of inside one another, and if we do need to run matrioska containers, [this](https://www.redhat.com/sysadmin/podman-inside-container) might help.

.ls-lint.yml
brumik: Cannot change, comment below

brumik:
>I just want to make sure this doesn't break anything (can you rebase please? - there's a new webpack config check that will fail, but it will tell us what changes in which config),

It did fail indeed, but it seems okey for me, though I am not sure what you are looking for exactly in it.

>I want to look into what it would take to keep running [this](https://www.redhat.com/sysadmin/podman-inside-container) from the insights mode compose script, I suspect nothing is stopping us from running the containers side by side instead of inside one another, and if we do need to run Matryoshka containers, this might help.

I checked before and matryoska containers are not too much recommended, so maybe I would go with the side by side. Is this something that is blocking this PR? Is this also something I should look into and work on (I believe it could take some time for me to figure out but I could be capable without too much outside help)?

There is one last outstanding issue, that backend is using the `profiles` files in compose, merging this will break probably that, but was it working fine with them before? This is anyways connected to the previous paragraph.

I tested it and the `npm run start` and `npm run start-standalone` commands both work as expected. Just to make sure from the description:

**Important**: The insights URL is http://localhost:8002/beta/ansible/automation-hub. It does not work without the beta since the pendo tries to load then.

himdel:
>>  there's a new webpack config check that will fail, but it will tell us what changes in which config),
>
> It did fail indeed, but it seems okey for me, though I am not sure what you are looking for exactly in it.

:+1: No changes to the generated standalone mode configs,
:+1: insights dev changes the url to /beta/, proxies `/api/automation-hub/` to `localhost:5001`, exposes AppEntry, those sound reasonable

:-1: insights prod config changes at all (https://github.com/ansible/ansible-hub-ui/runs/5288431559?check_suite_focus=true#step:5:421) that can't happen

And losing both the timestamp and the hash from the generated filename sounds like something we'll want to fix,
adding a proxy target of `undefined:undefined` (insights prod only) also looks like a bug,
and `/api/rbac/v1` sounds like something we may want to actively skip including,
(localhost:4000 is presumably keycloak which would be needed).

(The standalone dev config change of using app/entry instead of entry-dev seems harmless and expected given the code changes.)

I also notice we're still getting the `Insights-proxy is deprecated in favor of "useProxy" or "standalone".` error message, is that something that could also be fixed?

himdel:
> I checked before and matryoska containers are not too much recommended, so maybe I would go with the side by side. Is this something that is blocking this PR? Is this also something I should look into and work on (I believe it could take some time for me to figure out but I could be capable without too much outside help)?

Yeah, until now the only way to run insights dev mode was by running `./compose up` with `COMPOSE_PROFILE=insights` and `ANSIBLE_HUB_UI_PATH` pointing to the UI dir. So this is what everybody will expect to use. (If it turns out impossible, oh well, our ways can change, but we should try to make sure first :))

So, if that's something you'd like to work on, :+1: go for it :)

Ideally I would like to be able to do both
* run the backend with COMPOSE_PROFILE=insights without the UI and run the UI outside independently
* the same *with* the UI inside the container


> There is one last outstanding issue, that backend is using the profiles files in compose, merging this will break probably that, but was it working fine with them before? This is anyways connected to the previous paragraph.

It was working, I think we can just remove the references to those files from galaxy_ng, my understanding is that this replaces that, right? (I don't think the files themselves were important, I don't think anyone ran them locally or outside the docker env, so this is just about ensuring all the API endpoints are proxied to the right place.)

brumik:
>insights prod config changes at all (https://github.com/ansible/ansible-hub-ui/runs/5288431559?check_suite_focus=true#step:5:421) that can't happen

This is not expected, if `NODE_ENV` is not set to `development` then that piece of code what generates this should not be there.

>And losing both the timestamp and the hash from the generated filename sounds like something we'll want to fix

It does seem like an error, though I am not sure if it is on our side. It does seem to work, I will verify if the file indeed does not have suffixes or just the watcher is working without them (I believe there was a webpack feature which allowed for defining just name.js and would take any of those files regardless of suffixes)

> adding a proxy target of undefined:undefined (insights prod only) also looks like a bug,
and /api/rbac/v1 sounds like something we may want to actively skip including (but harmless),
(localhost:4000 is presumably keycloak which would be needed).

This all is connected to the NODE_ENV issue I mentioned before: the `insights.prod` config does not have proxy fields set, as it does not need it.

>I also notice we're still getting the Insights-proxy is deprecated in favor of "useProxy" or "standalone". error message, is that something that could also be fixed?

This error is coming when using `onPremise` version and I believe it is because the config uses the `webpack proxy` to, well, proxy stuff. We could change the config in a way to use the proxy that the platform is giving us for the `onPremise` version, but it is probably not worth the refactoring.

brumik:
Regarding your other comment:

Yes I can work on setting that up again, but after merging this and before fixing that the `./compose up with COMPOSE_PROFILE=insights` will not work. The `onPremise` version should keep working as it is expected.

Everything else mentioned there is correct and the expected state is to get both the compose and separate version to work.

brumik:
I updated the PR packages, since it stopped working and repushed.

As it seems this is the only way right now to get the insights mode working, maybe we can push this PR trough. Last time I think we were pretty close to it.

To renew: this will break the (already broken) method of inisights mode from `galaxy_ng`. I can work on it after this is merged. This will NOT break nor change the standalone development setup. 

Asked @ShaiahWren to try it out on his local setup too.
@jerabekjiri @MilanPospisil If you have time I would appreciate if you would give it a go too.

I update the description with information needed to get the insights mode running.

jerabekjiri:
Tested and working fine for me :slightly_smiling_face:. Only thing is no `Partners` namespaces or collections are populated in localhost. And on creating my own namespace, then uploading my own collection gets completed fine, but I get redirected to 404 (collection not found). I think this has something to do with `staging` and `published` state, because I can visit the collection in URL by changing repo to `staging`.

I had some permission issues with docker on `npm start` and fixed it by running `sudo chmod 666 /var/run/docker.sock` FYI

Also, I get a lot of errors in `console.log`, but that has nothing to do with this PR I guess :slightly_smiling_face:

brumik:
@jerabekjiri Is this this issue: https://github.com/ansible/ansible-hub-ui/pull/1807?
Yes, the console log warnings are fine, errors should not appear only a few times for failed requests.

himdel:
As per the last team meeting we agreed we can have the insights mode env work this way (separate set of api containers + running ui outside with it's own set of containers; as opposed to having all the conainters in our composefile) :+1: 

So this will just need a corresponding galaxy_ng PR to remove the errors caused by the deletions here.

Apart from that, mostly LGTM, except I was recently debugging a cloud issue, started from here and had to tweak a couple of things, and some of the changes no longer seem necessary (plus it works with non-beta as well after that) - I have those in https://github.com/himdel/ansible-hub-ui/commit/3bf447317ebd3bc72f3f1e79309309fa56187c7f, will try to add suggestions here later today :)

brumik:
Isn't https://github.com/himdel/ansible-hub-ui/commit/3bf447317ebd3bc72f3f1e79309309fa56187c7f working ONLY on non beta? I wasn't able to get your commit to work when tried to help Shaiah, but I can give it another go if you think that's the way we need to go.

I preferred to have the beta since lots of platform features go into beta first and we are (okey I am guessing now here) deploying to beta first, so it would make more sense to use the app with them, or am I mistaken?

himdel:
> Isn't https://github.com/himdel/ansible-hub-ui/commit/3bf447317ebd3bc72f3f1e79309309fa56187c7f working ONLY on non beta?

Yeah, we'd have to add `BETA=true` to env to switch to beta in that config.

And you're right, having our dev point to beta by default makes sense :+1: 

---

> So this will just need a corresponding galaxy_ng PR to remove the errors caused by the deletions here.

Creating https://github.com/ansible/galaxy_ng/pull/1224 to address that.

---

In the mean time, can you try removing the changes in `src/`, and the custom exclude/exposes config?

brumik:
@himdel Thank you for the backend PR. I reverted the renaming changes in the `src`. I deemed some minor changes there sensible to keep in this PR, though if you meant to remove all the changes made there I can do it without a problem.

himdel:
Well, this is still deleting a file marked "Please do not delete", removing a redux logger from dev mode, and having the prod version entryfile import the dev version entryfile.

Also still removing an unrelated favicon config, and overriding a list of excluded packages.

brumik:
The `entry-dev` says that it is dependent on the `base.webpack.config` and I think we don't need it anymore.

About the logger, I remeber us agreeing that it is not used as we have zero use case for redux.

I revert them anyways.

brumik:
> and having the prod version entryfile import the dev version entryfile.

Where?

> Also still removing an unrelated favicon config

I thought this will simplify, since we are using a custom favicon only in standalone. Sorry if I misunderstood it but I thoght you are on board with this change. I roll back.


> overriding a list of excluded packages

I'm not sure I explained this: the exclusion makes the packages to not to be served in standalone mode as production versions. It is much better for development if we get a nice message with trace than just a simple error message and blank/broken page.

Also all of the changes you mentioned is scoped under the insights deployment mode:
https://github.com/ansible/ansible-hub-ui/pull/1526/files#diff-421cd61c232189aee852bdc0470301998e2509fe281662239092c220b87cf4efR179

brumik:
@himdel Going back and forth I think is unproductive, please make your changes to this PR what you see fit so we can merge it. We are pretty close where we want to be, so I have no problem with any old code you wanna keep in some form or another - and it's much faster if you just contribute to the PR.

himdel:
OK, no problem, updated with a smaller change :).
(Does this still work for you?)

Also added a bit more info in the README, and a warning message when trying to run the docker env with ui enabled.

config/webpack.base.config.js
```diff
@@ -174,6 +194,10 @@ module.exports = (inputConfigs) => {
               isBuild ? '../src/app-entry.js' : '../src/dev-entry.js',
             ),
           },
+          ...(!isBuild && {
+            // fixes "Shared module is not available for eager consumption"
+            exclude: ['@patternfly/react-core'],
```
brumik, line 199: I don't understand why we are excluding only this package, but I trust your decision.

docker/entrypoint.sh
```diff
@@ -1,7 +1,13 @@
 #!/bin/sh
 if [[ $DEPLOYMENT_MODE == 'insights' ]]
 then
-  npm run start
```
brumik, line 4: :+1:

brumik (commented):
LGTM (Cannot approve my PR)

PR #1528, opened 2022-01-19
Testing and searching for sporadic failures

MilanPospisil:
Issue: AAH-1282

This is meant to fix sporadic failures in `imports_filter.js`:

    "before each" hook for "Exact search for completed is working." / "before each" hook for "exact filter for name is working."
    
    Request failed with status code 403

example: https://github.com/ansible/ansible-hub-ui/runs/5100841966?check_suite_focus=true#step:17:601 , https://github.com/ansible/ansible-hub-ui/runs/5087051667?check_suite_focus=true#step:17:603

(this does not address failures in `collection_upload` or `view_only`)

### DIFF
--- a/test/cypress/integration/imports_filter.js
+++ b/test/cypress/integration/imports_filter.js
@@ -54,6 +54,7 @@ describe('Imports filter test', () => {
   });
 
   it('Exact search for completed is working.', () => {
+    cy.wait(10);
     cy.get('.import-list button:first').click();
     cy.contains('a', 'Status').click();
 
--- a/test/cypress/integration/imports_filter.js
+++ b/test/cypress/integration/imports_filter.js
@@ -51,10 +51,10 @@ describe('Imports filter test', () => {
       .contains('different_name')
       .should('not.exist');
     cy.get('[data-cy="import-list-data"]').contains('my_collection1');
+    cy.wait(10000);
   });
 
   it('Exact search for completed is working.', () => {
-    cy.wait(10000);
     cy.get('.import-list button:first').click();
     cy.contains('a', 'Status').click();
 
--- a/test/cypress/integration/imports_filter.js
+++ b/test/cypress/integration/imports_filter.js
@@ -10,6 +10,10 @@ describe('Imports filter test', () => {
     cy.galaxykit('-i collection upload filter_test_namespace different_name');
   });
 
+  after(() => {
+    deleteNamespacesAndCollections();
+  });
+
   beforeEach(() => {
     cy.login();
     cy.visit('/ui/my-imports?namespace=filter_test_namespace');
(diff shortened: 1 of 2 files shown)


himdel (approved):
Let's merge this one and address my comments in a later PR, looks like this does resolve the sporadic failure :+1: Thanks :)

test/cypress/integration/imports_filter.js
```diff
@@ -10,29 +10,48 @@ describe('Imports filter test', () => {
...
+    cy.deleteNamespacesAndCollections();
+  });
+
   beforeEach(() => {
     cy.login();
     cy.visit('/ui/my-imports?namespace=filter_test_namespace');
   });
 
   it('partial filter for name is working.', () => {
+    cy.intercept(
```
himdel, line 23: could we move the intercept to `beforeEach`? It seems to be used by all the tests now :)

And more, we could also wait for the first request (the one when loading the page) in `beforeEach`, so each tests starts in a state where the first unfiltered list has already loaded.

test/cypress/integration/imports_filter.js
```diff
@@ -10,29 +10,48 @@ describe('Imports filter test', () => {
...
+      'GET',
+      Cypress.env('prefix') + '_ui/v1/collection-versions/?namespace=*',
+    ).as('wait');
+
     cy.get('input[aria-label="keywords"').type('my_collection{enter}');
-    cy.get('[data-cy="import-list-data"]').contains('my_collection1');
-    cy.get('[data-cy="import-list-data"]').contains('my_collection2');
+    cy.wait('@wait');
+
     cy.get('[data-cy="import-list-data"]')
```
himdel, line 31: this ordering doesn't really work, I think...

wait for request doesn't necessarily wait for the render after that request,
so then waiting for contains and *then* for not.exist ensures that the screen has loaded before the test for nonexistence of something
but waiting for not.exist is likely to pass whenever the page hasn't rendered yet, so we're pretty much skipping the not.exist bits now

test/cypress/integration/imports_filter.js
```diff
@@ -10,29 +10,48 @@ describe('Imports filter test', () => {
...
       .contains('my_collection2')
       .should('not.exist');
     cy.get('[data-cy="import-list-data"]')
       .contains('different_name')
       .should('not.exist');
+    cy.get('[data-cy="import-list-data"]').contains('my_collection1');
+    cy.wait(10000);
```
himdel, line 54: I think the wait would not be needed after https://github.com/ansible/ansible-hub-ui/pull/1528/files#r803225726 ...

this may have been failing because the first wait actually catches the request when loading the page, instead of the one when filtering

test/cypress/integration/imports_filter.js
```diff
@@ -78,12 +97,12 @@ describe('Imports filter test', () => {
     ).as('wait');
     cy.contains('a', 'Completed').click();
 
-    cy.get('[data-cy="import-list-data"]').contains('my_collection1');
```
himdel, line 81: (same here, moving the wait makes sense, but contains should go before not.exists)

PR #1529, opened 2022-01-20
Adding Secure URL validation

brumik:
Issue: https://issues.redhat.com/browse/AAH-1195

**Before**
Screenshot from 2022-01-10 14-14-26
Screenshot from 2022-01-10 14-14-15

**After**
Screenshot from 2022-01-20 14-25-28
Screenshot from 2022-01-20 14-24-44
Screenshot from 2022-01-20 14-24-28

**Note**: The fields have the colored underline from the PF when they are not "focused" as you can see them focused in the screenshot.

### DIFF
--- a/src/components/repositories/remote-form.tsx
+++ b/src/components/repositories/remote-form.tsx
@@ -50,7 +50,9 @@ const validateURLHelper = (
     if (protocol === 'https:') {
       return { validated: 'default' };
     }
-  } catch (_) {}
+  } catch (_) {
+    // fallthrough
+  }
 
   return {
     validated: 'error',
--- a/src/components/repositories/remote-form.tsx
+++ b/src/components/repositories/remote-form.tsx
@@ -16,11 +16,47 @@ import {
 
 import { WriteOnlyField, HelperText, FileUpload } from 'src/components';
 
-import { DownloadIcon } from '@patternfly/react-icons';
+import {
+  DownloadIcon,
+  ExclamationTriangleIcon,
+  ExclamationCircleIcon,
+} from '@patternfly/react-icons';
 
 import { RemoteType, WriteOnlyFieldType } from 'src/api';
 import { Constants } from 'src/constants';
 import { isFieldSet, isWriteOnly, ErrorMessagesType } from 'src/utilities';
+
+const validateURLHelper = (
+  outsideError: string | undefined,
+  url: string,
+): {
+  validated: 'default' | 'warning' | 'error';
+  helperTextInvalid?: string;
+  helperText?: string;
+} => {
+  if (outsideError) {
+    return { validated: 'error', helperTextInvalid: outsideError };
+  }
+
+  try {
+    const { protocol } = new URL(url);
+    if (protocol === 'http:') {
+      return {
+        validated: 'warning',
+        helperText: t`Consider using a secure URL (https://).`,
+      };
+    }
+
+    if (protocol === 'https:') {
+      return { validated: 'default' };
+    }
+  } catch (_) {}
+
+  return {
+    validated: 'error',
+    helperTextInvalid: t`The URL needs to be in 'http(s)://' format.`,
+  };
+};
 
 interface IProps {
   allowEditName?: boolean;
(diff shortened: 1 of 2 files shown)


src/components/repositories/remote-form.tsx
```diff
@@ -16,12 +16,47 @@ import {
...
+
+  if (url && url.length > 4) {
+    if (url.slice(0, 5) === 'http:') {
+      return {
+        validated: 'warning',
+        helperText: t`Consider using a secure URL (https://).`,
+      };
+    }
+
+    if (url.slice(0, 4) !== 'http') {
```
himdel, line 49:
```suggestion
    if (url.slice(0, 6) !== 'https:') {
```

brumik: This change does not make sense for me. My logic without your suggestion:
1. If it is HTTP then show warning (consider using HTTPS)
2. If it is not even starting with HTTP then it cannot be HTTP nor HTTPS so show error that you need a HTTP(s) URL
3. Else the URL is HTTPS so don't show anything

I guess you would like to see this as testing for HTTPS then HTTP then another test for none of these (adds an extra if statement) but since we have 3 error types this makes more sense with returning the "success (default)" with the last return.

brumik: If it helps readibility I can switch up my two `HTTP` and `HTTP:` if cases

himdel: Right, but that's not what the code is doing, `httpie.io` is now a valid value, no warnings ;)

himdel: So.. thinking about this some more, even adding the colon doesn't quite resolve this right, maybe this should be closer to..

```javascript
const httpWarning = {
  validated: 'warning',
  helperText: t`Consider using a secure URL (https://).`,
};

const protocolError = {
  validated: 'error',
  helperTextInvalid: t`The URL needs to be in 'http(s)://' format.`,
};

const success = { validated: 'default' };

try {
  const { protocol } = new URL(url);
  if (protocol === 'https:') {
    return success;
  }
  if (protocol === 'http:') {
    return httpWarning;
  }
} catch (e) {};

return protocolError;
```

src/components/repositories/remote-form.tsx
```diff
@@ -201,16 +236,20 @@ export class RemoteForm extends React.Component<IProps, IState> {
...
+            validated={
+              validateURLHelper(errorMessages['url'], remote.url).validated
+            }
             isRequired={requiredFields.includes('url')}
             isDisabled={disabledFields.includes('url')}
             id='url'
             type='text'
-            value={remote.url || ''}
+            placeholder='https://'
+            value={remote.url}
```
himdel, line 252: any reason for this change? All the other form fields are using the `|| ''` pattern here.

src/components/repositories/remote-form.tsx
brumik: We are expecting the people to know how URLs are constructed and initially we just wanted to add a `Secure` URL warning. As users probably can use it with IP addresses too and there is no backend check for validity I am starting to think this would be better with two warnings:

`The URL should be in http(s) format` and `Consider using a sercure URL (https)`.  

Also I see why using the `protocol` from the `URL` object is superior, I just kinda did not expected from this change to from a no warning is still mostly working users we go to users which:
1. Will type `httpie.io`
2. They get the warning message about using secure https
3. They get confused :)

I'll update it anyway :)

src/components/repositories/remote-form.tsx
brumik: I think the UI has a really nice way of distingushing the placeholder and current text. I updated it anyway.

src/components/repositories/remote-form.tsx
```diff
@@ -16,12 +16,48 @@ import {
...
+      return {
+        validated: 'warning',
+        helperText: t`Consider using a secure URL (https://).`,
+      };
+    }
+
+    if (protocol === 'https:') {
+      return { validated: 'default' };
+    }
+  } catch (_) {}
```
himdel, line 53: Just this lint error, otherwise LGTM :+1: :)

```suggestion
  } catch (_) {
    // fallthrough
  }
```

fixes that :)

PR #1553, opened 2022-01-24
Test RBAC permissions

jerabekjiri:
Issue: [AAH-635](https://issues.redhat.com/browse/AAH-635)

This PR tests the basic user's access with no permissions. Then adds permissions and test the access again with full permissions.

Comments:
- Task permissions - Noticed we are not using tasks permissions anywhere, just checking this is purposeful and wasn't forgotten to implement it
- EE page - in list view, I'm able to edit the container even with no permissions (the `Edit` button is visible in toggled kebab) In the EE detail page this works fine. So I added it to this PR as well.


brumik (commented):
This looks good for me.

For future: 
The tests cases should describe what should happen there, for example:
`it('should not display edit and delete buttons when user does not have permissions')` (note that you can read that what should the tested piece of code do - **it should ...**)

And breaking up the tests to per page basis would make it much more clear which page breaks the tests. This would help debugging test failures much more convenient, instead of going trough the code looking for the specific line that failed in a test. (Also the tests fails at the first failure, so you won't know if it fails on multiple pages, or just one.

jerabekjiri:
- Added newline to `635.misc` file
- Rewritten `it(...)` names
- Split tests into smaller parts and tested it per each permission (page), followed a little unconventional style with two describes per file, hope it's okay? Wanted to keep tests independent and cleaner code.
@brumik followed your advice, wdyt? :)

brumik:
Looks sleek, thanks for taking the time rewording the tests.

If @himdel has no objections I think this is good to merge :+1:

himdel (approved):

src/containers/execution-environment-list/execution_environment_list.tsx
```diff
@@ -330,18 +330,25 @@ class ExecutionEnvironmentList extends React.Component<
 
   private renderTableRow(item, index: number) {
     const description = item.description;
+
+    const canEdit = item.namespace.my_permissions.includes(
```
himdel, line 334: :+1: this looks like a fix that should go in 4.4, adding :)

PR #1555, opened 2022-01-25
Adding signatures to UI

brumik:
## TODO
* [x] Approval page update after the meeting:
* [ ] Add synching signed only badge to repo as in https://issues.redhat.com/browse/AAH-1333

## Jira linked Issues: 
* https://issues.redhat.com/browse/AAH-1059
* https://issues.redhat.com/browse/AAH-1061
* https://issues.redhat.com/browse/AAH-1062
* https://issues.redhat.com/browse/AAH-1264
* https://issues.redhat.com/browse/AAH-1333

## Current UI looks (verified by UX ppl, thanks):
Screenshot from 2022-01-26 15-31-35
Screenshot from 2022-01-26 15-31-25
Screenshot from 2022-01-26 15-31-17
Screenshot from 2022-01-26 15-37-44
Screenshot from 2022-01-26 15-37-36
Screenshot from 2022-01-26 15-37-27


trahman73:
Hi @brumik These are looking great! Here are some changes/questions below...

- Change "This change will affect 1 versions." to "This change will affect 1 version(s).
- In the Namespaces/Collections tab, are we adding a Signed/Unsigned badge to this view?
   Here is the mockup
image

Here is the styling and placement for the badge itself... 
- We used the PF outline green and orange options for signed/unsigned here:
  https://www.patternfly.org/v4/components/label#outline
- We used the compact option for the Collections card view only here (the rest of the areas are the regular size):
  https://www.patternfly.org/v4/components/label#compact
- signed icon: fa-check-circle CheckCircleIcon (react)
- unsigned icon: fa-exclamation-triangle ExclamationTriangleIcon (react)
- placement of the badge for the card view is top right of the card
image

brumik:
@trahman73 Thank you for the help, I updated the description and the code with the new badges and spelling.

trahman73:
@brumik Thanks for making the changes, they look great! The only thing I notices is that the badge size in the details view (next to the Last updated text) can be regular size vs compact. The sizing everywhere else looks great. Thanks!

brumik:
@trahman73 The badge feels comically big next to the gray text in regular size, the compact version is much better for the eyes, except if we deem this badge as an important information for the user that should "pop".

trahman73:
@brumik Sounds good and I agree with you. Do you think we should go with the compact view for the Namespaces/Collections tab as well?

brumik:
I think they are good looking without being compact there since there they have their own line

src/api/sign-collections.ts
```diff
@@ -0,0 +1,36 @@
...
+  UI_API_VERSION = 'v3';
+  apiPath = this.getUIPath('sign/collections/');
+
+  constructor() {
+    super(API_HOST + API_BASE_PATH);
+  }
+
+  getUIPath(url: string) {
+    return `${this.UI_API_VERSION}/${url}`;
+  }
```
himdel, line 29:
```suggestion
  apiPath = 'v3/sign/collections/';
```

(there is no such thing as a UI API v3, and as a bonus, the code gets simpler :))

src/containers/namespace-detail/namespace-detail.tsx
```diff
@@ -330,6 +337,27 @@ export class NamespaceDetail extends React.Component<IProps, IState> {
...
+        <SignAllCertificatesModal
+          name={this.state.namespace.name}
+          numberOfAffected={this.state.itemCount}
+          isOpen={this.state.isOpenSignModal}
+          onSubmit={() => {
+            SignCollectionAPI.sign({
+              signing_service: 'ansible-default',
+              repository: this.context.selectedRepo,
+              namespace: namespace.name,
+            })
```
himdel, line 349: missing success/error alerts

src/components/headers/collection-header.tsx
```diff
@@ -166,7 +180,7 @@ export class CollectionHeader extends React.Component<IProps, IState> {
       noDependencies
         ? this.context.user.model_permissions.delete_collection && (
             <DropdownItem
-              key={1}
+              key='1-enabled'
```
himdel, line 183: let's give these meaningful keys as well if we're changing them anyway .. let's say `delete-collection-enabled`?

src/components/headers/collection-header.tsx
```diff
@@ -190,21 +205,72 @@ export class CollectionHeader extends React.Component<IProps, IState> {
...
-            onClick={() =>
-              this.openDeleteModalWithConfirm(collection.latest_version.version)
-            }
-          >
-            {t`Delete version ${collection.latest_version.version}`}
-          </DropdownItem>
-        </div>
+        <DropdownItem
+          data-cy='delete-version-dropdown'
+          key='2'
```
himdel, line 210: (and here)

src/api/sign-collections.ts
brumik: Oki, I started out like this but I am not certain why we need `API_HOST + API_BASE_PATH` in some urls and here why  we don't then?

src/containers/namespace-detail/namespace-detail.tsx
brumik: Can you provide please an example of how to do it meaningfully? We are getting back an task ID and Bruno told me we should check somehow how the task ended, but I find it hard to find an example of an API error reporting back to the User inside the UI. (see my morning PM).

I can also log of the `signing task creation in the api` fails (this call ^), but not sure what is the UX for that one either.

src/components/headers/collection-header.tsx
brumik: Cool, then I rename the sibling keys in these issues too. I added them since we had cases where two keys were the same and then react complained :)

src/containers/namespace-detail/namespace-detail.tsx
brumik: From Zita a template:
https://github.com/ansible/ansible-hub-ui/blob/master/src/containers/certification-dashboard/certification-dashboard.tsx#L505

https://github.com/ansible/ansible-hub-ui/blob/master/src/containers/execution-environment-list/execution_environment_list.tsx#L524

src/api/task.ts
```diff
@@ -1,11 +1,46 @@
...
+import { TaskType } from './response-types/task';
 
 export class API extends HubAPI {
   apiPath = 'v3/tasks/';
 
   constructor() {
     super();
   }
+
+  waitFor(interval: NodeJS.Timeout, taskId: string): Promise<TaskType> {
```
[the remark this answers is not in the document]
brumik: @himdel Oh, sure, I wasn't aware of the existence of that Utility. This is based on the implementation I got from Zita here:

https://github.com/ansible/ansible-hub-ui/blob/master/src/containers/certification-dashboard/certification-dashboard.tsx#L505

https://github.com/ansible/ansible-hub-ui/blob/master/src/containers/execution-environment-list/execution_environment_list.tsx#L524

Would you be okay to pull the `waitForTask` logic from the utilites to the SignatureAPI so we can create one `then ... catch` logic in the components using it or should I use it as it is in the `deleteCollection`

In the delete collection there is no failed task check as it is in the certification dashboard - is that okey or should I check for it on top of that?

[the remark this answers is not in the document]
brumik: @ZitaNemeckova No problem at all, nothing not fixable :) I sill appreciate your help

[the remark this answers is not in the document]
brumik: And also the example you sent me does more than the Utility alone

brumik:
Thoughts:
1. Waiting for 5 sec requery for the task to complete is a lot, added a info alert to tell the user at least  that something is happening.
2. Alert List could be global, maybe we can consider using the one from frontend-compoents, has same interface
3. We are not checking if the task actually completed successfully or not, same in the delete task.

Concern: the `waitForTask` is recursive, so in case the task for some reason does not resolve, we can potentially get stack overflow error - it is unlikely but possible I think.

src/api/sign-collections.ts
himdel: Well, we really use 3 different sets of APIs now:

* pulp apis .. `PulpApi` does `BaseApi('/pulp/api/v3/')` - this will become something like `API_HOST + PULP_API_BASE_PATH` (+-) when adding task management for cloud (#1589)
* hub apis .. `HubApi` does `BaseApi(API_HOST + API_BASE_PATH)`, `API_HOST` is usually empty, and `API_BASE_PATH` is either `/api/automation-hub/` or `/api/galaxy/`
  * UI apis .. are under hub, but use extra `getUIPath` to add `_ui/UI_API_VERSION/` between the base and `apiPath`

So.. you can drop the constructor as well, if you change from `extends BaseAPI` to `extends HubAPI` :)

himdel:
On Approvals Dashboard, I'm still seeing "Approved" status, not "Signed and approved":


On Collection detail, I'm *not* seeing the Sign kebabs .. the last I heard we're supporting multiple signatures so the buttons should be visible and active...



Same screen, the expand signature button works, but the signature is mistreating whitespace, I think you need `white-space: pre;` on that element: (there should be a visible newline after BEGIN PGP SIGNATURE-+, and before -+END)



I'm also not seeing the sign buttons on the namespace detail screen, presumably also because all collections are signed?

The "Signed" label has the same size on all the list screens, but is somehow scrunched on the collection detail screen - we should probably use the same size everywhere...




Namespace detail is missing the signedness filter, so is collection list...

himdel:
Testing further with an unsigned collection (I had to comment out `PULP_GALAXY_AUTO_SIGN_COLLECTIONS` from `dev/standalone/galaxy_ng.env` to be able to change it in `settings.py`)...

:+1: Seeing the Sign buttons on all relevant screens, as well as the unsigned indicator, and the modals.

But in Approvals Dashboard, the Approved icon should be blue, not green, when the collection is only approved, not signed.
(And as per above, the text should be different too.)

Switching between versions of a partially signed collections works well :+1:, except we're losing the kebab buttons when looking at signed versions.

But on the collections list screen, and namespace detail screens, "partial" is treated the same as unsigned...


The version selector on the collection detail screen is missing the signedness info...

src/components/headers/collection-header.tsx
```diff
@@ -512,6 +563,111 @@ export class CollectionHeader extends React.Component<IProps, IState> {
...
+      repository: this.context.selectedRepo,
+      namespace: this.props.collection.namespace.name,
+      collection: this.props.collection.name,
+    })
+      .then((result) => {
+        waitForTask(result.data.task_id)
+          .then(() => {
+            // Cannot requery the api as it comes as prop
+            // We should get the collection into state from api probably
+            window.location.reload();
```
himdel, line 596: We can and should requery the api - exactly the same way deletion does it... `props.updateParams` should work :)

himdel:
Overall looks like we're close :+1:, but given the current size of this PR and the amount of changes still needed, I would strongly suggest splitting this up into a bunch of smaller PRs, this already took a few hours to review and re-reviewing may take equally long if not split :)

I'd suggest:

* the `src/api` changes, type changes, shared components and utilities - we can merge these almost immediately, pending `Partial` support
* approval dashboard - will still need work with the labels, but probably not much work
* list screeens - list screen features, I think the card rewrite looks sane, but probably needs more thorough review than most of the other ui changes, and these are missing the filters
* collection detail, modals - seems like this will need some work for the reload code
* remote form changes (just because they're not related to the rest, also mergable I think)

himdel:
@ShaiahWren can you go through the alerts introduced in this PR and make sure they align with https://github.com/ansible/ansible-hub-ui/pull/1469 please? :)

(We can do this post-merge if needed, but it should be before the release :).)

src/components/cards/cards.scss
```diff
@@ -37,21 +28,6 @@
...
-    justify-content: space-between;
-  }
-
-  .logo-row {
-    display: flex;
-    justify-content: space-between;
-
-    .icon {
-      color: var(--pf-global--info-color--100);
-      margin-right: 3px;
```
himdel, line 51: something like this is probably still needed when a collection has a wide logo...



(I just changed these namespaces to use http://placekitten.com/48/48 , http://placekitten.com/480/48 , http://placekitten.com/48/480 respectively.)

src/components/headers/collection-header.tsx
```diff
@@ -512,6 +563,111 @@ export class CollectionHeader extends React.Component<IProps, IState> {
...
+          })
+          .catch((error) => {
+            this.setState({
+              alerts: [...this.state.alerts, errorAlert(error)],
+            });
+          })
+          .finally(() => {
+            this.setState({
+              alerts: this.state.alerts.filter(
+                (x) => x?.id !== 'loading-signing',
```
himdel, line 658: this is unlike any other alerts we have .. waiting for task is tradionally handled with a green alert linking to the specific task

(we don't necessarily need to handle this one now, we already have a task to unify the task alerts IIRC)

src/components/signing/signature-badge.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+  return (
+    <Label
+      variant='outline'
+      color={isSigned ? 'green' : 'orange'}
+      icon={isSigned ? <CheckCircleIcon /> : <ExclamationTriangleIcon />}
+      isCompact={isCompact}
+      {...props}
+    >
+      {isSigned ? t`Signed` : t`Unsigned`}
```
himdel, line 34: (missing Partial)

himdel:
Also tested with backend master (without https://github.com/ansible/galaxy_ng/pull/1076),
looks like most of the logic works correctly, treating it as the "no signing" scenario :+1: ... except when you try to open a collection detail screen, the whole UI goes away...

```
Uncaught TypeError: Cannot read properties of undefined (reading 'length')
    at CollectionHeader.render (App.1643929411392.0dc0310801a5970c5502.js:102862:75)
    at finishClassComponent (App.1643929411392.0dc0310801a5970c5502.js:79468:31)
    at updateClassComponent (App.1643929411392.0dc0310801a5970c5502.js:79418:24)
    at beginWork (App.1643929411392.0dc0310801a5970c5502.js:80928:16)
    at HTMLUnknownElement.callCallback (App.1643929411392.0dc0310801a5970c5502.js:62496:14)
    at Object.invokeGuardedCallbackDev (App.1643929411392.0dc0310801a5970c5502.js:62545:16)
    at invokeGuardedCallback (App.1643929411392.0dc0310801a5970c5502.js:62600:31)
    at beginWork$1 (App.1643929411392.0dc0310801a5970c5502.js:85511:7)
    at performUnitOfWork (App.1643929411392.0dc0310801a5970c5502.js:84462:12)
    at workLoopSync (App.1643929411392.0dc0310801a5970c5502.js:84438:22)
react-dom.development.js:19527 The above error occurred in the <CollectionHeader> component:
    in CollectionHeader (created by CollectionDetail)
    in CollectionDetail (created by Context.Consumer)
    in withRouter(CollectionDetail) (created by AuthHandler)
    in AuthHandler (created by Context.Consumer)
    in Route (created by Routes)
    in Switch (created by Routes)
    in Routes (created by App)
    in main (created by Page)
    in div (created by Page)
    in Page (created by App)
    in App (created by Context.Consumer)
    in withRouter(App)
    in I18nProvider
    in Router (created by BrowserRouter)
    in BrowserRouter
```

... you can fix that by...

```diff
diff --git a/src/components/collection-detail/download-signature-grid-item.tsx b/src/components/collection-detail/download-signature-grid-item.tsx
index 2b41a727..e7f8d07b 100644
--- a/src/components/collection-detail/download-signature-grid-item.tsx
+++ b/src/components/collection-detail/download-signature-grid-item.tsx
@@ -21,7 +21,7 @@ const DownloadSignatureGridItem: FC<Props> = ({ version }) => {
   const [show, setShow] = useState(false);
 
   // No signature object or the signatures is empty
-  if (!signingEnabled || version.metadata.signatures.length < 1) {
+  if (!signingEnabled || !version.metadata.signatures?.length) {
     return null;
   }
 
diff --git a/src/components/headers/collection-header.tsx b/src/components/headers/collection-header.tsx
index 78fcde5a..0afe9c0e 100644
--- a/src/components/headers/collection-header.tsx
+++ b/src/components/headers/collection-header.tsx
@@ -161,7 +161,7 @@ export class CollectionHeader extends React.Component<IProps, IState> {
 
     const latestVersion = collection.latest_version.created_at;
     const isVersionSigned =
-      collection.latest_version.metadata.signatures.length > 0;
+      !!collection.latest_version.metadata.signatures?.length;
     const isCollectionSigned = collection.sign_state === 'signed';
 
     const isLatestVersion = (v) =>
```

src/components/signing/signature-badge.tsx
brumik: When it s partial, it is unsigned.

src/components/cards/cards.scss
brumik: I'll add padding trough PF compoent props

src/components/headers/collection-header.tsx
brumik: So pass down prop to update a higher level params which requery the api?

src/components/headers/collection-header.tsx
brumik: Tried my best. Send me an example what is your idea about it, I can change it.

brumik:
* The different sized signed/unsigned badge comes from a mutual work with UX (earlier in this PR)
* The missing button after collection is signed is also coming from UX (my last info is we can do it but we don't show for now). Even though the user can sign multiple times, it has no real difference.
* Yes, the PR become big. I did not expected it and you were the one suggesting one PR (even for the signed_only issue, and that one is still getting updates in the mockups)
* The is no UI indicator in the drop-down for the versions, after selecting the version you can see on the right side of the version if it's signed or not.
* There is no ticket or description about the dropdown filters for signed collections.
* After the backed pr merged the API returns all the fields even when the signing is disabled so the breaking should not happen.
* I dont understand you approval dashboard issue, I thought I followed the mockup, and it has not talked about changing the sign we are displaying after approval. The signed status is already displayed everywhere (so many times that John Hardy was questioning if we give it too much importance). 


@himdel I can make all the changes you want. When we agree upon it I break up this pr and make the changes.

cc
@trahman73

ShaiahWren:
@himdel I can compare this to #1469, no problem.  :)

trahman73:
Regarding the different size labels, we are using the compact label found here https://www.patternfly.org/v4/components/label#compact for the card views. The regular sized label can be used everywhere else. We did do a comparison with @brumik on where the label is placed on the Collection detail screen. Visually it looked better as the compact version as it is placed next to the smaller last updated text. If others feel that the compact label version should not be used here in the Collection details screen, UX is fine with using the regular size.

I think the latest from my understanding is that the signing options in the kebab menu should still display as it can be signed multiple times or conditions might change to require to re-sign? I know that after completing the signing flow for the first time, the user will see a signed badge. I personally think it's confusing to still see the signing options valid in the kebab menu especially if I just completed the signing process. I would only expect to see it again if any of the signing conditions have changed and that I would need to sign again. If the easier option here is to just keep the signing options active at all time, UX is fine with this behavior.

src/components/headers/collection-header.tsx
```diff
@@ -512,6 +563,111 @@ export class CollectionHeader extends React.Component<IProps, IState> {
...
+      description: t`Failed to sign all versions in the collection.`,
+    });
+
+    this.setState({
+      alerts: [
+        ...this.state.alerts,
+        {
+          id: 'loading-signing',
+          variant: 'info',
+          title: t`Signing all version in the collection...`,
```
ShaiahWren, line 579: I recommend an alert title that includes the name of collection signed, e.g. `Signing started for all versions in collection &quot;{name}&quot;`

src/components/headers/collection-header.tsx
```diff
@@ -512,6 +563,111 @@ export class CollectionHeader extends React.Component<IProps, IState> {
...
+      description: t`Failed to sign the version.`,
+    });
+
+    this.setState({
+      alerts: [
+        ...this.state.alerts,
+        {
+          id: 'loading-signing',
+          variant: 'info',
+          title: t`Signing the version...`,
```
ShaiahWren, line 632:
```suggestion
          title: t`Signing started for collection quot;{collectionName} v{vcollectionVersions}quot;.`,
```

ShaiahWren: It would be good to include the actual collection name and version :)

src/containers/namespace-detail/namespace-detail.tsx
```diff
@@ -368,6 +394,56 @@ export class NamespaceDetail extends React.Component<IProps, IState> {
...
+      description: t`Failed to sign all collections.`,
+    });
+
+    this.setState({
+      alerts: [
+        ...this.state.alerts,
+        {
+          id: 'loading-signing',
+          variant: 'info',
+          title: t`Signing all collections...`,
```
ShaiahWren, line 410:
```suggestion
          title: t`Signing started for all collections in namespace quot;{namespace}quot;.`,
```

ShaiahWren: Is this for signing all collections in the entire system, or for a given namespace?  If it's the latter, the suggestion above.

src/components/headers/collection-header.tsx
ShaiahWren, line 579:
```suggestion
          title: t`Signing started for all versions in collection quot;{collection}quot;.`,
```

brumik:
> I would only expect to see it again if any of the signing conditions have changed and that I would need to sign again. If the easier option here is to just keep the signing options active at all time, UX is fine with this behavior.

To my understanding when there is a key rotation or other kind of action which invalidates the signature the api will return `sign_status: 'unsigned'` or `partial` and then (by the current logic) the button would be showing again. @rochacbruno @trahman73

src/components/headers/collection-header.tsx
```diff
@@ -512,6 +563,109 @@ export class CollectionHeader extends React.Component<IProps, IState> {
...
+      title: t`API Error: ${status}`,
+      description: t`Failed to sign all versions in the collection.`,
+    });
+
+    this.setState({
+      alerts: [
+        ...this.state.alerts,
+        {
+          id: 'loading-signing',
+          variant: 'info',
```
ShaiahWren, line 578:
```suggestion
          variant: 'success',
```
So far as I know, we use the green 'success' alert for the successful start to any action, so I would go with that (unless @himdel says otherwise) :)

src/components/headers/collection-header.tsx
```diff
@@ -512,6 +563,109 @@ export class CollectionHeader extends React.Component<IProps, IState> {
...
+      title: t`API Error: ${status}`,
+      description: t`Failed to sign the version.`,
+    });
+
+    this.setState({
+      alerts: [
+        ...this.state.alerts,
+        {
+          id: 'loading-signing',
+          variant: 'info',
```
ShaiahWren, line 629:
```suggestion
          variant: 'success',
```
Same as above

src/containers/namespace-detail/namespace-detail.tsx
```diff
@@ -368,6 +394,56 @@ export class NamespaceDetail extends React.Component<IProps, IState> {
...
+      title: t`API Error: ${status}`,
+      description: t`Failed to sign all collections.`,
+    });
+
+    this.setState({
+      alerts: [
+        ...this.state.alerts,
+        {
+          id: 'loading-signing',
+          variant: 'info',
```
ShaiahWren, line 409:
```suggestion
          variant: 'success',
```
Again :)

src/components/headers/collection-header.tsx
brumik: meeting.

himdel: Confirming, we're going with success

himdel:
> The different sized signed/unsigned badge comes from a mutual work with UX (earlier in this PR)

Understood, objection withdrawn :)


> The missing button after collection is signed is also coming from UX (my last info is we can do it but we don't show for now). Even though the user can sign multiple times, it has no real difference.

I'm pretty sure @rochacbruno said that we should not be hiding the actions at those times...

bruno

Unless anything changed there, let's go with that, I don't think this feature is stable enough yet to be hiding the UI without being sure it's the right thing to do.


> The is no UI indicator in the drop-down for the versions, after selecting the version you can see on the right side of the version if it's signed or not.

Yes, that's why I said we should add it :). Otherwise, searching for an unsigned version would mean going through each version until you find the right one.


> There is no ticket or description about the dropdown filters for signed collections.

Apart from AAH-1060, and the details in https://github.com/ansible/galaxy_ng/pull/1076#issuecomment-1016942811 ;).


> After the backed pr merged the API returns all the fields even when the signing is disabled so the breaking should not happen.

Nevertheless, let's make sure your code can work with unchanged APIs, as the change is literally already done for you.


> I dont understand you approval dashboard issue, I thought I followed the mockup, and it has not talked about changing the sign we are displaying after approval. The signed status is already displayed everywhere (so many times that John Hardy was questioning if we give it too much importance). 

Cutting out from the Approvals dashboard design...

approval-dashboard-sap

approvals dashboard doesn't distinguish between collections that were only signed and collections that were also approved, the design above solves that.

In a later discussion we settled on blue, instead of green, when only approved, otherwise as depicted above.

src/components/headers/collection-header.tsx
himdel: (this may be superseded by the specific alerts suggestions below, but for completenes, this is when starting sync for a remote registry...

https://github.com/ansible/ansible-hub-ui/blob/master/src/containers/execution-environment-detail/base.tsx#L326-L342
)

src/components/signing/signature-badge.tsx
himdel: I'm not seeing any designs with the Partial indicator on Jira, and Slack unfortunately forgets images in history, @trahman73 would you still happen to have a collection list screen view with a partially signed collection?

(It pretty much looks exactly like Unsigned, but saying Partial, I believe, and would be shown on a collection list screen, whenever only some versions of such a collections were signed.

Or did I miss a change where these were abandoned?

src/components/headers/collection-header.tsx
brumik: Here the user has to close the Sync started notification manually, and if errors out there will be two alerts?

src/components/signing/signature-badge.tsx
brumik: I never seen mockups for partial label. We just treated the partial as unsigned.

src/components/headers/collection-header.tsx
brumik: @himdel  I had a discussion with @ShaiahWren where we agreed on a meeting for today (which was pushed to next week), hence the meeting comment. I also made clear to @ShaiahWren that I don't mind using success.

src/components/signing/signature-badge.tsx
brumik: Just a text change is needed for the partial?

src/components/headers/collection-header.tsx
brumik: I do the same just removing the notification when the task is complete, so user does not have to close it. Should I remoe that logic?

brumik:
Closing in favor of #1624 #1625 #1626 #1627 #1628 #1629 #1630 

Same stuff with the above mentioned improvements broken up into processable pieces

PR #1557, opened 2022-01-25
Namespace list: use CompoundFilter & AppliedFilters instead of custom toolbar

himdel:
Fixes AAH-820

AAH-820 was fixed in #1195 and #1402 for all screens using CompoundFilter, but NamespaceList wasn't using CompoundFilter yet (happened in #509 for the collection list screens).

This updates namespace list to use CompoundFilter and AppliedFilters, fixing the empty search issue, removing a custom label,  and the page now also changes to 1 when filtering/sorting.

Before:

151068139-d7678430-ce7a-4fac-8f25-95d1905b8163
151068149-7ed63a9b-6397-4cfa-8cb7-19a8786639e1

After:


jerabekjiri (approved):
LGTM :+1:

PR #1559, opened 2022-01-26
AboutModal, html comment: show UI git commit hash

himdel:
This runs `git rev-parse HEAD` at build time to get the current git commit hash, and provide it as a `UI_COMMIT_HASH` config.

Then, in standalone mode, the About modal now has a new "UI Version" entry, and there's a html comment as well..

And in insights mode, there is a new html comment with the version, near the end...

TODO test & screenshot


himdel:
(webpack-config fail expected.. this is adding a new `UI_COMMIT_HASH` var to `definitions`)

himdel:
Merging to test insights mode in today's stage deploy.

PR #1561, opened 2022-01-26
Warning message for missing collection's documentation in Community repository

jerabekjiri:
Issue: [AAH-1207](https://issues.redhat.com/browse/AAH-1207)

- Add warning message to Documentation tab if no Documentation is present in Community collection

Before:
Screenshot from 2022-01-26 18-10-47

After:
Screenshot from 2022-01-26 17-54-03

- Should this warning message be displayed in the `Contents` tab as well?

### DIFF
--- a/src/components/collection-detail/collection-content-list.tsx
+++ b/src/components/collection-detail/collection-content-list.tsx
@@ -96,45 +96,43 @@ export class CollectionContentList extends React.Component<IProps> {
             </ToolbarGroup>
           </Toolbar>
         </div>
-        {toShow.length > 0 ? (
-          <table className='hub-c-table-content pf-c-table pf-m-compact'>
-            <thead>
-              <tr>
-                <th>{t`Name`}</th>
-                <th>{t`Type`}</th>
-                <th>{t`Description`}</th>
+        <table className='hub-c-table-content pf-c-table pf-m-compact'>
+          <thead>
+            <tr>
+              <th>{t`Name`}</th>
+              <th>{t`Type`}</th>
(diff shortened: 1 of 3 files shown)


jerabekjiri:
Also since the left sidemenu should always be empty, we could do a fullscreen warning in this case. 
Screenshot from 2022-01-26 17-56-48

himdel:
> Should this warning message be displayed in the Contents tab as well?

Agreed, it should, the content counts are also mentioned in the issue.

Otherwise LGTM :+1: 

> Also since the left sidemenu should always be empty, we could do a fullscreen warning in this case.

Let's stay with the design for now, the sidemenu gives us a confirmation that there's indeed nothing there at least :)

jerabekjiri:
- adding Contents tab screenshot

Screenshot from 2022-01-31 23-13-02

src/components/collection-detail/collection-content-list.tsx
```diff
@@ -126,7 +130,20 @@ export class CollectionContentList extends React.Component<IProps> {
             ))}
           </tbody>
         </table>
+        {toShow.length <= 0 &&
```
himdel, line 133: ooh this can't be `toShow.length` here, because that's after both keyword and type filtering

`summary.all === 0 &&` should work :)

himdel (approved):
LGTM once https://github.com/ansible/ansible-hub-ui/pull/1561/files#r797964024 is fixed :+1: :)

jerabekjiri:
Changed to `summary.all` and merging this. Thanks for the correction @himdel :)

PR #1577, opened 2022-01-31
CollectionContentList - fix by type counts being 0 when All not chosen

himdel:
go to Collection detail, tab Contents

the "Showing: All(23) Role(9) Module(3) Action(1) Filter(8) Test(2)" line only shows correct counts when All is selected,
otherwise only the current type is counted, leaving other types at 0,
because filtering is done by both keyword and type, before counting and showing the list

this changes the code to first filter by keyword, then count, then filter by type, then show

Cc @jerabekjiri I noticed this while reviewing #1561 , but it's also present on master, can you review please? :)

---

before: (only the last 2 pics differ)

after:

### DIFF
--- a/src/components/collection-detail/collection-content-list.tsx
+++ b/src/components/collection-detail/collection-content-list.tsx
@@ -41,15 +41,20 @@ export class CollectionContentList extends React.Component<IProps> {
     const keywords = params.keywords || '';
 
     for (const c of contents) {
-      const typeMatch = showing === 'all' ? true : c.content_type === showing;
-      if (!summary[c.content_type]) {
-        summary[c.content_type] = 0;
-      }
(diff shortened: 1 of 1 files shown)


jerabekjiri:
```
summary[c.content_type] ||= 0;
```
Never seen this before, pretty cool :sunglasses: 

Noticed weird recounting of `content_type` when switching between tabs, but this fixes it as well :+1: 

LGTM

jerabekjiri (approved):

himdel:
Thanks :)

[Logical assignment operators](https://2ality.com/2020/06/logical-assignment-operators.html) - they're quite recent :)

himdel:
Introduced in #507, the change only made it to 4.4, adding `backport-4.4`, and an issue... TODO

PR #1586, opened 2022-01-31
add env variable to dockerfile

ShaiahWren:
No-Issue

This pr adds a new ENV variable in the dockerfile to increase memory and minimize crashes due to limited space.


newswangerd (approved):

himdel (approved):
I can't reproduce the failures locally, but this looks like the right fix as long as you're running the UI this way in docker :+1: 

Only... https://nodejs.org/api/cli.html#--max-old-space-sizesize-in-megabytes mentiones `max-old-space-size` instead of `max_old_space_size`.

But if you're sure this works for you with underscores, no objections :)

PR #1588, opened 2022-02-01
Cypress - replace overused helpers with galaxykit

MilanPospisil:
Replace helpers like userCreate that uses GUI operations every time by galaxykit commands who do the same thing. This will speed up the tests, because we want to test creating user by GUI only once, not multiple times.

https://issues.redhat.com/browse/AAH-1192

Depends on: https://github.com/ansible/galaxykit/pull/31

### DIFF
--- a/test/cypress/integration/execution_environments.js
+++ b/test/cypress/integration/execution_environments.js
@@ -12,12 +12,12 @@ describe('execution environments', () => {
       `docker${num}`,
       'https://registry.hub.docker.com/',
     );
-    cy.addRemoteContainer({
-      name: `remotepine${num}`,
-      upstream_name: 'library/alpine',
-      registry: `docker${num}`,
-      include_tags: 'latest',
-    });
+    cy.galaxykit(
+      'container create',
+      `remotepine${num}`,
+      'library/alpine',
+      `docker${num}`,
+    );
   });
 
   beforeEach(() => {
--- a/test/cypress/integration/execution_environments.js
+++ b/test/cypress/integration/execution_environments.js
@@ -7,7 +7,11 @@ describe('execution environments', () => {
     cy.deleteRegistries();
     cy.deleteContainers();
 
-    cy.addRemoteRegistry(`docker${num}`, 'https://registry.hub.docker.com/');
+    cy.galaxykit(
+      'registry create',
+      `docker${num}`,
+      'https://registry.hub.docker.com/',
+    );
     cy.addRemoteContainer({
       name: `remotepine${num}`,
       upstream_name: 'library/alpine',
--- a/test/cypress/integration/execution_environments_use_in_controller.js
+++ b/test/cypress/integration/execution_environments_use_in_controller.js
@@ -13,10 +13,11 @@ describe('Execution Environments - Use in Controller', () => {
 
     cy.deleteRegistries();
     cy.deleteContainers();
-
-    // FIXME: add galaxykit support for remote registries
-    cy.addRemoteRegistry(`docker${num}`, 'https://registry.hub.docker.com/');
-
+    cy.galaxykit(
+      'registry create',
+      `docker${num}`,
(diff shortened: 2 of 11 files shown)


test/cypress/integration/execution_environments.js
```diff
@@ -7,13 +7,17 @@ describe('execution environments', () => {
     cy.deleteRegistries();
     cy.deleteContainers();
 
-    cy.addRemoteRegistry(`docker${num}`, 'https://registry.hub.docker.com/');
-    cy.addRemoteContainer({
-      name: `remotepine${num}`,
-      upstream_name: 'library/alpine',
-      registry: `docker${num}`,
-      include_tags: 'latest',
```
himdel, line 15: this seems problematic - have we lost the ability to pass include_tags latest to galaxykit?

MilanPospisil: I will look at it.

MilanPospisil: Im not completely sure, but when you create (using GUI) EE container and include latest tag, there is post request to /_ui/v1/execution-environments/remotes/ that has those fields: exclude_tags: [], include_tags: []. Galaxykit also sends those empty arrays. I remember that I did it on purpose, so I guess galaxykit creates equal EE container. And it cant create EE enviroment with different tags than latest. Is there any way we can check it?

himdel: we'll need to update galaxykit to either default to latest, or allow overriding

himdel: => https://github.com/ansible/galaxykit/pull/40

himdel: Resolved: galaxykit will now always include_tags=latest

test/cypress/integration/execution_environments_use_in_controller.js
```diff
@@ -10,11 +10,8 @@ describe('Execution Environments - Use in Controller', () => {
     });
 
     cy.login();
-
     cy.deleteRegistries();
     cy.deleteContainers();
-
-    // FIXME: add galaxykit support for remote registries
     cy.addRemoteRegistry(`docker${num}`, 'https://registry.hub.docker.com/');
```
ZitaNemeckova, line 15: Why is it not replaced by galaxykit?

test/cypress/integration/group_management.js
```diff
@@ -28,12 +28,12 @@ describe('Hub Group Management Tests', () => {
     let groupName = 'testGroup';
     let userName = 'testUser';
 
-    cy.createGroup(groupName);
-    cy.createUser(userName);
-    cy.addUserToGroup(groupName, userName);
+    cy.galaxykit('group create', groupName);
+    cy.galaxykit('user create', userName, userName + 'Password');
+    cy.galaxykit('user group add', userName, groupName);
```
ZitaNemeckova, line 33: `cy.addUserToGroup(groupName, userName);` should stay as a check that it can be done via UI.

test/cypress/integration/remote_registry.js
```diff
@@ -24,7 +24,12 @@ describe('Remote Registry Tests', () => {
 
   it('admin can add new remote registry', () => {
     cy.menuGo('Execution Environments > Remote Registries');
-    cy.addRemoteRegistry('New remote registry1', 'https://some url1');
```
ZitaNemeckova, line 27: This one should not be replaced as it is a check it can be done via UI.

test/cypress/integration/execution_environments_use_in_controller.js
MilanPospisil: Because it was replaced in other places (I think), only two uses remained, but with different number of arguments, so I left those here to test if command works correctly.

ZitaNemeckova (approved):

PR #1597, opened 2022-02-04
UI changelog

ShaiahWren:
See issue: https://issues.redhat.com/browse/AAH-1228

This pr adds a pyproject.toml file that provides the configuration for towncrier build.


himdel:
Yeah, the changelog looks about as expected, though we can't add "CHANGES.rst" in *this* PR - we'll have to wait for an actual release for that.

Other than that... `CHANGES/.gitignore` is not needed, especially if it is just unignoring itself,
but we should not be adding any `*.pyc` files .. may want to add `*.pyc` to the toplevel `.gitignore`.

(Or you can solve this locally by adding `export PYTHONDONTWRITEBYTECODE=true` to your environment (.bashrc, .bash_profile or the zsh equivalents, depending on what you're using.) - that would prevent python from creating these on your side, up to you :).)

---

Other than that, :+1: from me, leaving the rest up to @awcrosby :)
(I don't think the currently generated form is particularly useful, but I don't have a problem with it either :).)

awcrosby (approved):

ShaiahWren:
Everything looks good so far as the task at hand, but the test in `imports.js` is unhappy. Since this pr doesn't change any cypress and I saw the same fail in #1555, is the test buggy/ need to be changed? Was @MilanPospisil working on it? @himdel perhaps you know what's going on :)

MilanPospisil:
> Everything looks good so far as the task at hand, but the test in `imports.js` is unhappy. Since this pr doesn't change any cypress and I saw the same fail in #1555, is the test buggy/ need to be changed? Was @MilanPospisil working on it? @himdel perhaps you know what's going on :)

Yeah, I already finished PR that fixes it (althrough it is quick fix that it is not ideal).

ShaiahWren:
@MilanPospisil would it be this one?test/cypress/integration/imports_filter.js

himdel:
Oh @ShaiahWren feel free to merge despite that imports failure, you're correct, it's a sporadic failure Milan has been working on.

(Milan please do share the specific PR you had in mind :).)

ShaiahWren:
@himdel Looks like we're just waiting on the sporadic failures pr to get merged (https://github.com/ansible/ansible-hub-ui/pull/1528) but no pressure, I'm sure there are many other prs calling your attn.:)

ShaiahWren:
Merging pr despite sporadic failures at imports.js, which will be cleaned up in pr #1528 .

PR #1620, opened 2022-02-07
Bump stylelint to 14.3.0, switch to stylelint-config-standard-scss, fix

himdel:
This bumps stylelint to 14.3.0,
the new version of [`stylelint-config-recommended-scss`](https://github.com/stylelint-scss/stylelint-config-recommended-scss/blob/HEAD/index.js) changes some rules, but recommends using [`stylelint-config-standard-scss`](https://github.com/stylelint-scss/stylelint-config-standard-scss/blob/HEAD/index.js) for a more complete set of rules, doing that.

And then going by rule, details in individual commits :)

fix `alpha-value-notation`
fix `at-rule-empty-line-before`
fix `color-function-notation`
fix `color-hex-length`
fix `declaration-block-no-redundant-longhand-properties`
disable `declaration-colon-newline-after` - because `prettier` disagrees
fix `declaration-empty-line-before`
fix `length-zero-no-unit`
fix `rule-empty-line-before`
disable `scss/at-import-partial-extension` - because we resolve them at the webpack level
fix `scss/dollar-variable-empty-line-before`
fix `scss/double-slash-comment-empty-line-before`
disable `selector-class-pattern` (for now)
fix `shorthand-property-no-redundant-values`
override `string-quotes` to `single` - because `prettier` disagrees

### DIFF
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -2,6 +2,7 @@
   "extends": "stylelint-config-standard-scss",
   "rules": {
     "declaration-colon-newline-after": null,
-    "scss/at-import-partial-extension": null
+    "scss/at-import-partial-extension": null,
+    "selector-class-pattern": null
   }
 }
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -1,6 +1,7 @@
 {
   "extends": "stylelint-config-standard-scss",
   "rules": {
-    "declaration-colon-newline-after": null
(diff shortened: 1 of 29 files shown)


himdel:
Leaving stylelint on 13.* everywhere older than 4.5+:
stable-4.2: #2286
stable-4.3: (no PR yet)
stable-4.4: #2401
(stable-4.5+: this PR)

PR #1622, opened 2022-02-08
Adding links to namespaces has undesireable UX

MilanPospisil:
Issue: AAH-149

https://issues.redhat.com/browse/AAH-149

This issue is to track some follow-on work from [AAH-18](https://issues.redhat.com/browse/AAH-18). Now at least adding links works mostly as expected, but there is still some undesirable behavior:

    Links without http:// are valid and users will enter them. "Enter a valid URL" does not make it clear what is missing, either. Schema can be added if missing.
    Every time the namespace is saved, the order of the links is reversed.

### DIFF
--- a/src/components/namespace-form/namespace-form.scss
+++ b/src/components/namespace-form/namespace-form.scss
@@ -27,11 +27,13 @@
   margin-bottom: 10px;
 
   .link-name {
-    flex: 1;
+    flex-grow: 1;
+    flex-basis: 0;
   }
 
   .link-url {
-    flex: 2;
+    flex-grow: 2;
+    flex-basis: 0;
     margin-left: 20px;
   }
 
--- a/src/components/namespace-form/namespace-form.scss
+++ b/src/components/namespace-form/namespace-form.scss
@@ -27,18 +27,19 @@
   margin-bottom: 10px;
 
   .link-name {
-    flex-grow: 1;
+    flex: 1;
   }
 
   .link-url {
-    flex-grow: 2;
+    flex: 2;
     margin-left: 20px;
   }
 
   .link-button {
     display: flex;
-    align-items: center;
+    align-items: top;
     margin-left: 10px;
+    margin-top: 8px;
 
     .link-container {
       width: 20px;
--- a/src/components/namespace-form/namespace-form.tsx
+++ b/src/components/namespace-form/namespace-form.tsx
@@ -1,6 +1,7 @@
 import { t } from '@lingui/macro';
 import * as React from 'react';
 import './namespace-form.scss';
+import { validateURLHelper } from 'src/utilities';
 
 import {
   Form,
@@ -78,7 +79,6 @@ export class NamespaceForm extends React.Component<IProps, IState> {
             >
               <TextInput
                 validated={this.toError(!('company' in errorMessages))}
-                isRequired
                 id='company'
                 type='text'
                 value={namespace.company}
@@ -234,27 +234,68 @@ export class NamespaceForm extends React.Component<IProps, IState> {
     this.props.updateNamespace(update);
   }
 
+  public static validateName(link): {
+    validated: 'default' | 'error';
+    helperTextInvalid?: string;
+  } {
+    if (link.url) {
+      if (link.name) {
+        return { validated: 'default' };
+      } else {
+        return {
+          validated: 'error',
+          helperTextInvalid: t`Name must not be empty.`,
+        };
+      }
+    }
+
+    // if link url is empty, there is no need to insert name because the link data will be discarded
+    return { validated: 'default' };
+  }
+
+  public static validateUrl(link): ReturnType<typeof validateURLHelper> {
+    if (link.url) {
+      // only validate url if input is not blank, blank inputs are thrown away
+      return validateURLHelper(undefined, link.url);
+    }
+
+    if (link.name) {
+      return {
+        validated: 'error',
+        helperTextInvalid: t`URL must not be empty.`,
+      };
+    }
+
+    return { validated: 'default' };
+  }
+
   private renderLinkGroup(link, index) {
     const last = index === this.props.namespace.links.length - 1;
     return (
       <div className='useful-links' key={index}>
         <div className='link-name'>
-          <TextInput
-            id='name'
-            type='text'
-            placeholder={t`Link text`}
-            value={link.name}
-            onChange={(value, event) => this.updateLink(index, value, event)}
-          />
+          <FormGroup fieldId={'name'} {...NamespaceForm.validateName(link)}>
+            <TextInput
+              id='name'
+              type='text'
+              placeholder={t`Link text`}
+              value={link.name}
+              onChange={(value, event) => this.updateLink(index, value, event)}
+              validated={NamespaceForm.validateName(link).validated}
+            />
+          </FormGroup>
         </div>
         <div className='link-url'>
-          <TextInput
-            id='url'
-            type='text'
-            placeholder={t`Link URL`}
-            value={link.url}
-            onChange={(value, event) => this.updateLink(index, value, event)}
-          />
+          <FormGroup fieldId={'link'} {...NamespaceForm.validateUrl(link)}>
+            <TextInput
+              id='url'
+              type='text'
+              placeholder={t`Link URL`}
+              value={link.url}
+              onChange={(value, event) => this.updateLink(index, value, event)}
+              validated={NamespaceForm.validateUrl(link.url).validated}
+            />
+          </FormGroup>
         </div>
         <div className='link-button'>
           <div className='link-container'>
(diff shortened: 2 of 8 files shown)


MilanPospisil:
Before:
URLbefore
After:
URLAfter

MilanPospisil:
Note that Save button is still active (the previous functionality stays and button does not save invalid data and displays its own error message). Should it be disabled?

src/utilities/validateURLHelper.ts
```diff
@@ -0,0 +1,37 @@
+import { t } from '@lingui/macro';
+
+export function validateURLHelper() {
```
himdel, line 3: :+1: on extracting the helper, but why a function returning another function? Can you just export the inner function?

himdel:
> Note that Save button is still active (the previous functionality stays and button does not save invalid data and displays its own error message). Should it be disabled?

Ah, ideally yes, when a link fails UI validation, the Save button should be disabled.

Once UI validation is successful, the user can submit and then possibly get hit with the errors from the API from any other model validations.

But I think this is still missing the part where we take the errors from the server response, and show them by the right field as well. Let me know if you need any help with that, I think the error splitting logic is implemented in `mapErrorMessages`, hopefully it's also compatible with how this form is written :)

MilanPospisil:
> > Note that Save button is still active (the previous functionality stays and button does not save invalid data and displays its own error message). Should it be disabled?
> 
> Ah, ideally yes, when a link fails UI validation, the Save button should be disabled.
> 
> Once UI validation is successful, the user can submit and then possibly get hit with the errors from the API from any other model validations.
> 
> But I think this is still missing the part where we take the errors from the server response, and show them by the right field as well. Let me know if you need any help with that, I think the error splitting logic is implemented in `mapErrorMessages`, hopefully it's also compatible with how this form is written :)

The server does not return errors that could be easily assigned to particular items. For example for 3 invalid items (2 blank names and 3 invalid urls) it says:

0: {status: "400", code: "invalid", title: "Invalid input.", detail: "'asdsadads' is not a valid url.",…}
code: "invalid"
detail: "'asdsadads' is not a valid url."
source: {parameter: "links__url"}
status: "400"
title: "Invalid input."
1: {status: "400", code: "blank", title: "Invalid input.", detail: "This field may not be blank.",…}
code: "blank"
detail: "This field may not be blank."
source: {parameter: "links__name"}
status: "400"
title: "Invalid input."
2: {status: "400", code: "invalid", title: "Invalid input.",…}
code: "invalid"
detail: "'asddasdasb**-ui/pull/1490/files' is not a valid url."
source: {parameter: "links__url"}
status: "400"
title: "Invalid input."
3: {status: "400", code: "blank", title: "Invalid input.", detail: "This field may not be blank.",…}
code: "blank"
detail: "This field may not be blank."
source: {parameter: "links__name"}
status: "400"
title: "Invalid input."
4: {status: "400", code: "invalid", title: "Invalid input.", detail: "'gfhfhf' is not a valid url.",…}
code: "invalid"
detail: "'gfhfhf' is not a valid url."
source: {parameter: "links__url"}
status: "400"
title: "Invalid input."

I suggest that better way is to do good client side validation instead. What do you think?

MilanPospisil:
Another thing. Company name field has IsRequired in the textinput, but no error message is shown when it is empty, nor the server returns any errors. The IsRequired should not be here i guess?

src/components/namespace-form/namespace-form.tsx
```diff
@@ -1,6 +1,7 @@
 import { t } from '@lingui/macro';
 import * as React from 'react';
 import './namespace-form.scss';
+import validateURLHelper from './../../utilities/validateURLHelper';
```
himdel, line 4:
```suggestion
import validateURLHelper from 'src/utilities/validateURLHelper';
```

src/components/namespace-form/namespace-form.tsx
```diff
@@ -233,27 +234,76 @@ export class NamespaceForm extends React.Component<IProps, IState> {
...
+    validated: 'default' | 'error';
+    helperTextInvalid?: string;
+  } {
+    if (link.url) {
+      if (link.name) {
+        return { validated: 'default' };
+      } else {
+        return {
+          validated: 'error',
+          helperTextInvalid: 'Name must not be empty.',
```
himdel, line 247:
```suggestion
          helperTextInvalid: t`Name must not be empty.`,
```

src/components/namespace-form/namespace-form.tsx
```diff
@@ -233,27 +234,76 @@ export class NamespaceForm extends React.Component<IProps, IState> {
...
+  public static validateUrl(link): ReturnType<typeof validateURLHelper> {
+    if (link.url) {
+      // only validate url if input is not blank, blank inputs are thrown away
+      return validateURLHelper(undefined, link.url);
+    }
+
+    if (link.name) {
+      return {
+        validated: 'error',
+        helperTextInvalid: 'URL must not be empty.',
```
himdel, line 265:
```suggestion
        helperTextInvalid: t`URL must not be empty.`,
```

src/components/repositories/remote-form.tsx
```diff
@@ -26,39 +26,7 @@ import { RemoteType, WriteOnlyFieldType } from 'src/api';
...
-  } catch (_) {
-    // fallthrough
-  }
-
-  return {
-    validated: 'error',
-    helperTextInvalid: t`The URL needs to be in 'http(s)://' format.`,
-  };
-};
+import validateURLHelper from './../../utilities/validateURLHelper';
```
himdel, line 29:
```suggestion
import validateURLHelper from 'src/utilities/validateURLHelper';
```

src/components/namespace-form/namespace-form.tsx
```diff
@@ -233,27 +234,76 @@ export class NamespaceForm extends React.Component<IProps, IState> {
...
-          <TextInput
-            id='name'
-            type='text'
-            placeholder={t`Link text`}
-            value={link.name}
-            onChange={(value, event) => this.updateLink(index, value, event)}
-          />
+          <FormGroup
+            fieldId={'name'}
+            validated={NamespaceForm.validateName(link).validated}
```
himdel, line 279:
```suggestion
```

(the split on the next line already does that)

src/components/namespace-form/namespace-form.tsx
```diff
@@ -233,27 +234,76 @@ export class NamespaceForm extends React.Component<IProps, IState> {
...
-          <TextInput
-            id='url'
-            type='text'
-            placeholder={t`Link URL`}
-            value={link.url}
-            onChange={(value, event) => this.updateLink(index, value, event)}
-          />
+          <FormGroup
+            fieldId={'link'}
+            validated={NamespaceForm.validateUrl(link).validated}
```
himdel, line 295:
```suggestion
```

src/containers/edit-namespace/edit-namespace.tsx
```diff
@@ -212,6 +213,20 @@ class EditNamespace extends React.Component<RouteComponentProps, IState> {
...
+    let disabled = false;
+    namespace.links.forEach((link) => {
+      if (
+        NamespaceForm.validateName(link).validated == 'error' ||
+        NamespaceForm.validateUrl(link).validated == 'error'
+      ) {
+        disabled = true;
+      }
+    });
+    return disabled;
```
himdel, line 227:
```suggestion
    return namespace.links.some((link) =>
      NamespaceForm.validateName(link).validated == 'error' ||
      NamespaceForm.validateUrl(link).validated == 'error');
```

(+- prettier :))

[Array.some](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some)

himdel:
> The server does not return errors that could be easily assigned to particular items...
> I suggest that better way is to do good client side validation instead.

Well we still need to show server errors when they do happen, but agreed we can't really show server errors together with the right item, and clients side validation is enough to make it nice :)

himdel:
> Another thing. Company name field has IsRequired in the textinput, but no error message is shown when it is empty, nor the server returns any errors. The IsRequired should not be here i guess?

Yeah agreed, I'm pretty sure the company name is not required now. There used to be an `isRequired` on the corresponding `FormGroup` before #50 but I'm not sure this was ever required. (And it seems the `TextInput` `isRequired` prop only really sets the `<input required>` prop, which is ignored in `<form novalidate>` (html forms vs react forms).)

So, harmless, but feel free to remove it, I agree it doesn't belong there :)

himdel:
Thanks, the previous issue is fixed, just one more detail :) ...



the width of the right input apparently depends on the presence/length of the message under it, we should keep the 50:50 layout as before

src/components/namespace-form/namespace-form.scss
```diff
@@ -27,18 +27,19 @@
   margin-bottom: 10px;
 
   .link-name {
-    flex-grow: 1;
+    flex: 1;
```
himdel, line 30: can we keep flex-grow as is, and just add `flex-basis: 0`? (to both)
should have the same effect and makes it more clear what the problem was :)

(Either way LGTM :+1:, fixes the problem :))

PR #1625, opened 2022-02-09
[Signing 2] Add only signed options to remote form

brumik:
* [x] There is maybe a need to add a sign that repo is syncs only signed only 
  * Waiting for more info here: https://issues.redhat.com/browse/AAH-1333
  * Api does not supports it yet, by chat conversation we will not implement this now.

Screenshot from 2022-02-09 09-46-40

### DIFF
--- a/src/components/repositories/remote-form.tsx
+++ b/src/components/repositories/remote-form.tsx
@@ -12,9 +12,12 @@ import {
   Modal,
   Checkbox,
   ExpandableSection,
+  Switch,
 } from '@patternfly/react-core';
 
 import { WriteOnlyField, HelperText, FileUpload } from 'src/components';
+
+import { AppContext } from 'src/loaders/app-context';
 
 import {
   DownloadIcon,
@@ -87,6 +90,8 @@ interface IState {
 }
 
 export class RemoteForm extends React.Component<IProps, IState> {
+  static contextType = AppContext;
+
   constructor(props) {
     super(props);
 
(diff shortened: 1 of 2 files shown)


brumik:
@ZitaNemeckova this PR is ready. That particular feature can come later when api supports it.

ZitaNemeckova:
@brumik Right now I can say it looks good code-wise :+1:  

**BUT** it's impossible to test without API and definetely shouldn't be merged because it breaks the form rigth now if an user changes the value of `signed_only`. So that's why I would add some kind of warning to prevent anyone from merging it before API.

src/components/repositories/remote-form.tsx
```diff
@@ -256,6 +257,18 @@ export class RemoteForm extends React.Component<IProps, IState> {
           />
         </FormGroup>
 
+        <FormGroup
```
ZitaNemeckova, line 260: This needs to be hidden when there's no signing related feature flag.

src/components/repositories/remote-form.tsx
```diff
@@ -256,6 +257,20 @@ export class RemoteForm extends React.Component<IProps, IState> {
           />
         </FormGroup>
 
+        {this.context?.featureFlags?.collection_signing === true && (
```
himdel, line 260: `RemoteForm` doesn't seem to be using context .. perhaps this also needs these?

```
import { AppContext } from 'src/loaders/app-context';
...
  static contextType = AppContext;
```

himdel (approved):
LGTM :)

> Api does not supports it yet, by chat conversation we will not implement this now.

Is this still relevant?

brumik:
@himdel Yes, we will card it up probably now.

himdel:
> Yes, we will card it up probably now.

What does this mean?

That sounds like "yes, we're not doing this" but if so, why the merge?

brumik:
Thanks for the merge, yes, the missing part is moved to a separate issue.

PR #1626, opened 2022-02-09
[Signing 2] Add signed filter to the toolbar

brumik:
Screenshot from 2022-02-09 09-52-26

### DIFF
--- a/src/components/collection-list/collection-filter.tsx
+++ b/src/components/collection-list/collection-filter.tsx
@@ -59,7 +59,17 @@ export class CollectionFilter extends React.Component<IProps, IState> {
           title: tag,
         })),
       },
-    ];
+      this.context?.featureFlags?.collection_signing === true && {
+        id: 'sign_state',
+        title: t`Sign state`,
+        inputType: 'select' as const,
+        options: [
+          { id: 'signed', title: t`Signed` },
+          { id: 'unsigned', title: t`Unsigned` },
+          { id: 'partial', title: t`Partial` },
+        ],
+      },
+    ].filter(Boolean);
 
     return (
       <Toolbar>
@@ -75,6 +85,11 @@ export class CollectionFilter extends React.Component<IProps, IState> {
               />
               <ToolbarItem>
                 <AppliedFilters
+                  niceNames={{
+                    sign_state: t`sign state`,
+                    tags: t`tags`,
+                    keywords: t`keywords`,
+                  }}
                   style={{ marginTop: '16px' }}
                   updateParams={updateParams}
                   params={params}
--- a/CHANGES/1060.misc
+++ b/CHANGES/1060.misc
@@ -1,0 +1,1 @@
+Allow filtering collection by sign state
--- a/src/containers/search/search.tsx
+++ b/src/containers/search/search.tsx
@@ -85,7 +85,8 @@ class Search extends React.Component<RouteComponentProps, IState> {
   render() {
     const { loading, collections, params, numberOfResults } = this.state;
     const noData =
-      collections.length === 0 && !filterIsSet(params, ['keywords', 'tags']);
+      collections.length === 0 &&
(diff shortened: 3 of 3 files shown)


ZitaNemeckova (changes requested):
So I have clean DB and latest backend master. I uploaded arista cvp v.3.1.0 collection. It's auto approved. And when I try to filter by `Sign state` a get 400 error in console.

```
Request URL: http://localhost:8002/api/automation-hub/_ui/v1/repo/published/?sign_state=unsigned&deprecated=false&offset=0&limit=10

{"errors":[{"status":"400","code":null,"title":"Invalid input.","detail":"Invalid Filter: 'sign_state'","source":{"parameter":"errors"}}]}

```
Screenshot 2022-03-02 at 10 42 36

brumik:
@ZitaNemeckova  Well, not having the backend to filter by something is problematic. I would lie if I would say I still remember the details but I think these PRs supposed to work **only** with this backend branch https://github.com/ansible/galaxy_ng/pull/1145 which adds actually signing to it.

Though let me know if we need to make this whole feature to not cause problems on backends which are not supporting signing (but as a feature we are not backporting this so I don't know if there is a need for it)

ZitaNemeckova:
To make sure I do understand it correctly. Not having a sign feature flag means an user can see all signing related info like labels (because they can sync it from repo that has signed collections) but cannot sign it's own collections. So seeing a collection the user uploaded and approved as signed in this filter is correct, right?

@himdel @brumik Code-wise it looks good, it works well with the backend PR. I'm just not sure what should and shouldn't be hidden when sign feature flag is false.

himdel:
I think as long as things don't blow up, hiding Unsigned is not critical.. there were some discussions about what to do when signing is not enabled, but syncing from a repository where signing is enabled, and there is a case for still showing that information in such case.

However we may be able to tell these cases apart and hide it when `sign_state` / `signatures` is not present.

src/components/collection-list/collection-filter.tsx
```diff
@@ -59,7 +59,17 @@ export class CollectionFilter extends React.Component<IProps, IState> {
...
       },
-    ];
+      this.context?.featureFlags?.collection_signing === true && {
+        id: 'sign_state',
+        title: 'Sign state',
+        inputType: 'select' as const,
+        options: [
+          { id: 'signed', title: 'Signed' },
+          { id: 'unsigned', title: 'Unsigned' },
+          { id: 'partial', title: 'Partial' },
```
himdel, line 69:
```suggestion
        title: t`Sign state`,
        inputType: 'select' as const,
        options: [
          { id: 'signed', title: t`Signed` },
          { id: 'unsigned', title: t`Unsigned` },
          { id: 'partial', title: t`Partial` },
```

src/components/collection-list/collection-filter.tsx
```diff
@@ -75,6 +85,11 @@ export class CollectionFilter extends React.Component<IProps, IState> {
               />
               <ToolbarItem>
                 <AppliedFilters
+                  niceNames={{
+                    sign_state: 'sign state',
+                    tags: 'tags',
+                    keywords: 'keywords',
```
himdel, line 91:
```suggestion
                    sign_state: t`Sign state`,
                    tags: t`Tags`,
                    keywords: t`Keywords`,
```

src/components/collection-list/collection-filter.tsx
```diff
@@ -59,7 +59,17 @@ export class CollectionFilter extends React.Component<IProps, IState> {
           title: tag,
         })),
       },
-    ];
+      this.context?.featureFlags?.collection_signing === true && {
+        id: 'sign_state',
+        title: 'Sign state',
```
himdel, line 64:
```suggestion
        title: t`Sign state`,
```

himdel (approved):

PR #1627, opened 2022-02-09
[Signing 2] Add download signature item

brumik:
Screenshot from 2022-02-09 10-03-48
Screenshot from 2022-02-09 10-03-39

### DIFF
--- a/src/components/collection-detail/download-signature-grid-item.tsx
+++ b/src/components/collection-detail/download-signature-grid-item.tsx
@@ -1,0 +1,59 @@
+import React, { FC, useState } from 'react';
+import {
+  Button,
+  ButtonVariant,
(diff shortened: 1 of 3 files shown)


src/components/collection-detail/download-signature-grid-item.tsx
```diff
@@ -0,0 +1,57 @@
...
+        {show &&
+          version.metadata.signatures.map(({ signature }) => (
+            <code key={signature}>{signature}</code>
+          ))}
+      </GridItem>
+    </>
+  );
+};
+
+export default DownloadSignatureGridItem;
```
himdel, line 57: we don't really use default exports in src/components, export const ..?

himdel (approved):

PR #1628, opened 2022-02-09
[Signing 2] Add sign badge to collection card

brumik:
Screenshot from 2022-02-09 10-38-21
Screenshot from 2022-02-09 10-37-38

### DIFF
--- a/src/components/cards/cards.scss
+++ b/src/components/cards/cards.scss
@@ -14,14 +14,6 @@
 }
 
 .hub-c-card-collection-container {
-  width: 280px;
-
-  // For some reason the PF Card components overrides all the defaults for
-  // divs
-  div {
-    display: block;
-  }
-
   .name {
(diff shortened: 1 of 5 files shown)


himdel:
| here | before |
|-|-|
|  |  |
|  |  |
|  |  |

himdel:
^ Seems like we get inconsitent card sizing now, and title and company don't get cut off with ellipsis.

Is that something you can fix, or should we just add the badge to the existing cards without changing how they work?

himdel:
I think we can replace this by https://github.com/ansible/ansible-hub-ui/pull/1763 , looks like adding SignatureBadge to the existing CardHeader works fine by itself :)

brumik:
I probably just missed a PF prop on the card. But your PR seems smaller, we can roll with that.

Closing in favour of #1763

PR #1624, opened 2022-02-09
[Signing 1] Add types, api and components to prepare the UI

brumik:
No UI changes. Just types and API as @himdel recommended in a small PR.

PR #1629, opened 2022-02-09
[Signing 2] Collection and namespace signing

brumik:
Screenshot from 2022-02-09 13-05-33
Screenshot from 2022-02-09 13-05-26
Screenshot from 2022-02-09 13-05-10
Screenshot from 2022-02-09 13-05-00


brumik:
I'll rebase this when https://github.com/ansible/ansible-hub-ui/pull/1624 is merged.

trahman73:
This is minor, but the signed styling on the version dropdown itself is like below... same gray color test as the "Last updated..." message.
image

This helps to further separate the (latest) from the signed text.

brumik:
@trahman73 I never seen a design like this. The latest is the part of the dropdown item, but the `Signed` does not seem like it. Can you provide an example of an expanded dropdown please?

trahman73:
@brumik Here is the full dropdown...

image

If we can't do the above styling to the PF component, then this one will work... I just combined...

 (Latest) (Signed) to (Latest, Signed)

image

src/components/headers/collection-header.tsx
```diff
@@ -545,12 +700,7 @@ export class CollectionHeader extends React.Component<IProps, IState> {
                 ...this.state.alerts,
                 {
                   variant: 'success',
-                  title: (
-                    <Trans>
-                      Collection &quot;{name} v{collectionVersion}&quot; has
-                      been successfully deleted.
-                    </Trans>
-                  ),
+                  title: t`Successfully deleted collection version.`,
```
himdel, line 703: bad rebase?

brumik: Hm, I thought I fixed it in the last re base. This is the 3rd time it did not work, let me try again.

brumik: Okey, I rewrote it manually, now it should be fine.

src/components/headers/collection-header.tsx
```diff
@@ -513,6 +571,109 @@ export class CollectionHeader extends React.Component<IProps, IState> {
...
+      signing_service: this.context.settings.GALAXY_COLLECTION_SIGNING_SERVICE,
+      repository: this.context.selectedRepo,
+      namespace: this.props.collection.namespace.name,
+      collection: this.props.collection.name,
+      version: this.props.collection.latest_version.version,
+    })
+      .then((result) => {
+        waitForTask(result.data.task_id)
+          .then(() => {
+            window.location.reload();
```
himdel, line 654: why not `this.props.updateParams({});`?

brumik: I wasn't aware this would do the same thing. Thanks, I'll update it.

src/components/headers/collection-header.tsx
```diff
@@ -202,10 +220,46 @@ export class CollectionHeader extends React.Component<IProps, IState> {
           {t`Delete version ${collection.latest_version.version}`}
         </DropdownItem>
       ),
+      canSign(this.context) && (
+        <DropdownItem
+          key='sign-all'
+          onClick={() => this.setState({ isOpenSignAllModal: true })}
+        >
+          {`Sign entire collection`}
```
himdel, line 228:
```suggestion
          {t`Sign entire collection`}
```

himdel:
(Rebasing should fix the tests (https://github.com/ansible/ansible-hub-ui/pull/1827))

himdel (approved):
LGTM :+1: 

I wasn't able to sign, and `canSign` returned false as there's no `sign_collections_on_namespace` permission now, but that looks like it's being addressed in #1801

himdel:
Ah, @brumik is on PTO, all the view-only parts seem to be working, the non-working things are not visible, merging to unblock work on the rest.

PR #1631, opened 2022-02-10
Edit remote registry - Save button enabled when invalid data

MilanPospisil:
Issue: [AAH-1357](https://issues.redhat.com/browse/AAH-1357)

Edit remote registry - Save button enabled when invalid data
When URL is invalid, save button should be disabled.

### DIFF
--- a/src/components/repositories/remote-form.tsx
+++ b/src/components/repositories/remote-form.tsx
@@ -733,8 +733,10 @@ export class RemoteForm extends React.Component<IProps, IState> {
       }
     }
 
-    if (validateURLHelper(this.props.errorMessages['url'], remote.url).validated == 'error')
-    {
+    if (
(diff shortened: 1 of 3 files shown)


src/components/repositories/remote-form.tsx
```diff
@@ -733,6 +733,13 @@ export class RemoteForm extends React.Component<IProps, IState> {
       }
     }
 
+    if (
+      validateURLHelper(this.props.errorMessages['url'], remote.url)
```
himdel, line 737:
```suggestion
      validateURLHelper(null, remote.url)
```

Because `errorMessages` gets set when the server responds with an error. So if validation fails because the server responsed to the *previous* value with an error, it won't let us try again.

(Which will become relevant if the API ever fails for `url` with any extra validation.)

himdel (approved):

PR #1652, opened 2022-02-14
Unify fail notifications

ShaiahWren:
Issue: https://issues.redhat.com/browse/AAH-1354

This pr works towards unifying fail alert notifications across the application. The effort will be broken into several prs. 

Please reference this [working document](https://docs.google.com/document/d/1CnC37CYwIafjIbkqSWTKOJSvnSSTrqUR6vTiK2leMXg/edit) for a list of target notifications.

### DIFF
--- a/CHANGES/1354.task
+++ b/CHANGES/1354.task
@@ -1,0 +1,1 @@
+Unify fail alerts notifications across the application.
--- a/src/components/headers/collection-header.tsx
+++ b/src/components/headers/collection-header.tsx
@@ -523,7 +523,7 @@ export class CollectionHeader extends React.Component<IProps, IState> {
     )
       .then((res) => {
         const taskId = this.getIdFromTask(res.data.task);
-        // const name = deleteCollection.name;
+        const name = deleteCollection.name;
         waitForTask(taskId).then(() => {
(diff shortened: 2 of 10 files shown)


himdel:
Looks good, the one thing I'm not yet sure of is where there was an existing `err.message` and it isn't used anymore..

```diff
-                description: err?.message,
+                description: errorMessage(status, statusText),
```

I think most or all of those *should* be ok, but I wonder if there could be places where we're ignoring a specific server response to display a generic message instead. (Such as displaying "The server was unable to complete your request." instead of something like "Name is required" coming from the server.)

Would `err?.message || errorMessage(status, statusText)` work?
(If not I can check those places and see what kind of API responses we could be getting.)

ShaiahWren:
@himdel hmm, I _think_ the err?.message is even more simple and generic than the errorMessage, see console logs of each: 
Screen Shot 2022-02-22 at 12 07 00 PM

himdel:
Ooh, `err?.message` is always coming from the http response `statusText` then?
I was assuming there would be at least some cases where the message is coming from the actual API json response.

himdel (approved):

PR #1654, opened 2022-02-16
Part II: Unify fail alerts

ShaiahWren:
Issue: https://issues.redhat.com/browse/AAH-1354

This pr works towards unifying fail alert notifications across the application. The effort will be broken into several prs.

Please reference this [working document](https://docs.google.com/document/d/1CnC37CYwIafjIbkqSWTKOJSvnSSTrqUR6vTiK2leMXg/edit) for a list of target notifications.


src/containers/edit-namespace/edit-namespace.tsx
```diff
@@ -96,13 +101,13 @@ class EditNamespace extends React.Component<RouteComponentProps, IState> {
                 ...this.context.alerts,
                 {
                   variant: 'danger',
-                  title: t`Error loading active user.`,
-                  description: e?.message,
+                  title: t`Active user profile "${this.state.userName}" could not be displayed.`,
```
ShaiahWren, line 104: I'm debating whether type name (`this.state.userName`)  is useful in this alert given that it's in componentDidMount. If it fails before setState then userName will be an empty string

src/containers/edit-namespace/edit-namespace.tsx
```diff
@@ -282,8 +290,8 @@ class EditNamespace extends React.Component<RouteComponentProps, IState> {
             this.setState({
               alerts: this.state.alerts.concat({
                 variant: 'danger',
-                title: t`API Error: ${error.response.status}`,
-                description: t`You don't have permissions to update this namespace.`,
+                title: t`Changes to namespace "${this.state.namespace.name}" could not be saved.`,
```
ShaiahWren, line 293: I gave this the new errorMessage() for description so it's probably not an issue now, but I'm curious what status '404' had a description about permissions, was is intended to be 401?

ShaiahWren: Is there any error handling we can do if the action fails before the state of pertinent objects are set? Error handling the error handling :)

src/containers/certification-dashboard/certification-dashboard.tsx
```diff
@@ -531,8 +533,8 @@ class CertificationDashboard extends React.Component<
           updatingVersions: [],
           alerts: this.state.alerts.concat({
             variant: 'danger',
-            title: t`API Error: 500`,
-            description: t`Could not update the certification status for ${version.namespace}.${version.name}.${version.version}.`,
+            title: t`Changes to certification status for collection "${version.namespace} ${version.name} v${version.version}" could not be saved.`,
+            description: t`Error 500 - Internal Server Error: The server encountered an error and was unable to complete your request.`,
```
himdel, line 537: should this one use `errorMessage(500, t`Internal Server Error`)` to reuse the message?

src/containers/edit-namespace/edit-namespace.tsx
himdel: Ah, you're right that this won't be available in state, because the request that failed is the one that puts it there.

However, before we even display the screen (well, unless in view-only mode), we're setting the current user in `context`, so I think `this.context.user?.username` should work.

src/containers/edit-namespace/edit-namespace.tsx
himdel: This comes from https://github.com/ansible/ansible-hub-ui/pull/102, may be worth checking out if this can still happen..
But looks like this is a special enough situation that the generic message is good enough, as long as there's any error that it didn't work.

> 404 had a description about permissions, was it intended to be 401?

This could be the case of "we're not telling you if it doesn't exist just for you or at all", but I'm not 100% sure :). Might also have changed in the last 2 years.

> Is there any error handling we can do if the action fails before the state of pertinent objects are set? Error handling the error handling :)

Well, `this.state?.namespace?.name` is probably the best we could do for the most general case, maybe having a version of the message without the name.

But in this case we can probably do better, a) we always know the namespace when showing/editing because it's in the url (and thus props) as well, but b) this is `saveNamespace`, so the state has definitely loaded, or the form should not even let you submit.

ShaiahWren: Do I interpret correctly that we can leave my new code as is? :)

himdel: Heh, yeah, this should work :+1: :)

src/components/namespace-form/namespace-form.tsx
```diff
@@ -120,7 +120,7 @@ export class NamespaceForm extends React.Component<IProps, IState> {
                   formErrors: {
                     ...this.state.formErrors,
                     groups: {
-                      title: t`Error loading groups.`,
+                      title: t`Groups list could not be displayed.`,
                       description: err,
```
himdel, line 124: so, `err` in `ObjectPermissionField` `onError` is the same as `err?.message` elsewhere .. should this therefore also use `errorMessage`?

(You'll need to 

```diff
--- a/src/components/permissions/obect-permission-field.tsx
+++ b/src/components/permissions/obect-permission-field.tsx
@@ -123,2 +123,2 @@ export class ObjectPermissionField extends React.Component<IProps, IState> {
-      .catch((e) => this.props.onError(e?.message));
+      .catch((e) => this.props.onError(e));
```

to make it work.)

src/components/namespace-modal/namespace-modal.tsx
```diff
@@ -176,7 +176,7 @@ export class NamespaceModal extends React.Component<IProps, IState> {
                     formErrors: {
                       ...this.state.formErrors,
                       groups: {
-                        title: t`Error loading groups.`,
+                        title: t`Groups list could not be displayed.`,
                         description: err,
```
himdel, line 180: here as well.. https://github.com/ansible/ansible-hub-ui/pull/1654/files#r815554055

src/components/execution-environment/repository-form.tsx
```diff
@@ -404,18 +406,18 @@ export class RepositoryForm extends React.Component<IProps, IState> {
...
                 }
-                onError={(err) =>
+                onError={(err) => {
                   this.setState({
                     formErrors: {
                       ...this.state.formErrors,
                       groups: {
-                        title: t`Error loading groups.`,
+                        title: t`Groups list could not be displayed.`,
                         description: err,
```
himdel, line 415: here as well.. https://github.com/ansible/ansible-hub-ui/pull/1654/files#r815554055

himdel (commented):
LGTM as well, just the 3 `ObjectPermissionField` changes :)

ShaiahWren:
@himdel This works, though I had to change `onError` to take type `any` for it to work. Lmk if that works for you!

ZitaNemeckova (approved):
LGTM 👍  Cypress failures seem to be unrelated to the PR changes but I'm not 100% sure.

PR #1655, opened 2022-02-17
[Hub-4.4.0dev] Under 'Groups > admins > Edit group permissions', the lists for the chip group category {Namespaces, Collections,Users,Groups,Remotes,Containers} , need to be marked as translatable string

MilanPospisil:
Issue: AAH-1199
https://issues.redhat.com/browse/AAH-1199

While in Japanese google-chrome, and editing group permissions i.e. Groups > admins > Edit group permissions, the lists for the chip group category {Namespaces, Collections,Users,Groups,Remotes, Containers} appears in English, and need to marked as a translatable string.


src/components/permissions/permission-chip-selector.tsx
```diff
@@ -37,12 +40,30 @@ export class PermissionChipSelector extends React.Component<IProps, IState> {
...
+      this.props.selectedPermissions.forEach((item) => {
+        const item2 = {};
+        // orginal english value
+        item2['value'] = item;
+        item2.toString = function (): string {
+          // translated value
+          return i18n._(item);
+        };
+        selections.push(item2);
+      });
```
himdel, line 62: Tsk, this sounds like a use for `map` ;)

```suggestion
      selections = this.props.selectedPermissions.map((value) => ({
        // orginal english value
        value,
        // translated
        toString: () => i18n._(value),
      }));
```

(+- comments)

himdel:
Works for me, seems like both the readonly and read-write modes get translated correctly, 

However, the other use of PermissionChipSelector (Namespaces > detail > Edit namespace - Namespace owners) still doesn't translate the permissions:



(`bababa` was the group name and should not be translated, but the individual available and selected permissions should)

himdel (approved):
Works in both places now :+1: :)

PR #1656, opened 2022-02-20
lint:line - add linelint to ensure newlines before eof

himdel:
Adding new `npm run ` commands:

* `lint:line` - run linelint to check for missing newlines before end of file (for text files)
* `lint:line:fix` - run `linelint -a` - autofix
* `lint:line:setup` - install linelint via go get
* `lint:editorconfig` - run editorconfig-checker to check for other whitespace issues, indent, leading and trailing spaces
&nbsp;
* `lint-fix` - run all `lint:*:fix` rules
* `lint-setup` - run all `lint:*:setup` rules
* `lint:po:setup`, `lint:yaml:setup` - move `pip install lint-po yamllint` from `pr-checks.yml` to `package.json`, use `lint-setup` instead

And one commit runs `lint:line:fix` to fix any remaining missing newlines,
while another manually fixes errors produced by `lint:editorconfig` :)


himdel:
hm, breaks because of https://github.com/sabhiram/go-gitignore/commit/223ce5d391b0bfb8af397c4baf55ba4f0c9cd04c#diff-5d9c0f43886779d058bd6931863a1a9dc7965426e228927b4bfeb0160875e617R173 but uses the right version locally .. could be about go version maybe

if not, editorconfiglint can do the same it looks

himdel:
Closing, we can revisit later, but for now #2818 might be enough

PR #1382, opened 2021-12-17
Rescope class names

epwinchell:
Jira: https://issues.redhat.com/browse/RHCLOUD-16591

- Add "hub" prefix to classes
- Use PF/Platform class naming convention

PR #1657, opened 2022-02-20
Enable hub prefix css rule, dedup toolbar css, wrap globals

himdel:
Follow-up to #1620

The goal of this PR is to:

* ensure any css in `src/components/` and `src/containers/` is tied to a specific hub component by wrapping all of it in a `.hub-c-COMPONENT-NAME {` rule
* ensure all classnames use a `hub-` prefix (unless overriding a `pf-` style, and that never outside a specific component/container selector) (as per #1382)
* shared (=currently duplicated) css is pulled to a shared place - definitely all the identical toolbar classes
* ensure insights mode doesn't load possibly conflicting css - styles for standalone menu, body, or root
* remove any dead css

### DIFF
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -3,7 +3,7 @@
   "rules": {
     "declaration-colon-newline-after": null,
     "scss/at-import-partial-extension": null,
-    "selector-class-pattern": null,
+    "selector-class-pattern": "hub-",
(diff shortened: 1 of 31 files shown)


himdel:
created https://issues.redhat.com/browse/AAH-1581 for the top bits

PR #1677, opened 2022-02-21
Collections card view - fix author field overflow for long author names

himdel:
The title is already using ellipsis for long names, but the author field doesn't, relying on wrapping for now.

This breaks on names like `approval_dashboard_namespace_test_additional_data`...

20220221031757-allbad

Fixing...


MilanPospisil (approved):
The test failing seems to be unrelated (classic sporadic failure that needs to be repaired).

PR #1187, opened 2021-11-02
Localization: Update translated strings

Spredzy:
Signed-off-by: Christian M. Adams <chadams@redhat.com>

PR #1678, opened 2022-02-21
[Hub-4.4.0dev] Text over 'Log In' button, could be marked as translatable string

MilanPospisil:
Issue: AAH-1204

https://issues.redhat.com/browse/AAH-1204

While in Japanese google-chrome, text over 'Log In' button, could be marked as translatable string.

### DIFF
--- a/CHANGES/1204.misc
+++ b/CHANGES/1204.misc
@@ -1,0 +1,1 @@
+Added translation to Log In button. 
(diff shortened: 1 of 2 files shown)


MilanPospisil:
The text is not being translated yet, it needs also translation in locale.

himdel (approved):
:+1:  tested with `test_l10n`:

himdel:
> The text is not being translated yet, it needs also translation in locale.

Agreed, but that happens asynchronously, we don't have to worry about it here :)

(cypress on github actions is running `npm run gettext:extract ; npm run gettext:compile` before tests ([here](https://github.com/ansible/ansible-hub-ui/blob/master/.github/workflows/cypress.yml#L135-L136)),
there's a weekly github action that updates these files on master and stable-4.4 ([here](https://github.com/ansible/ansible-hub-ui/blob/master/.github/workflows/i18n.yml)),
and the actual translations get updated in PRs like https://github.com/ansible/ansible-hub-ui/pull/1187)

PR #1679, opened 2022-02-21
[Hub-4.4.0dev] Execution Environments > alpine > Activity, the (last) string 'was added' needs to be marked as translatable string

MilanPospisil:
Issue: AAH-1205

https://issues.redhat.com/browse/AAH-1205

While in Japanese google-chrome, do following:

    Add a new Execution Evironment say name is 'alpine'
    Upstream name say 'libpod/alpine'
    Registry 'Quay'
    Tags to include 'latest'
    Hit 'Save'
    Later do 'Sync from registry'
    Visit Execution Environments > alpine > Activity
    Observe the (last) string 'was added'

'was added' need to be marked as translatable string.

### DIFF
--- a/CHANGES/1205.misc
+++ b/CHANGES/1205.misc
@@ -1,0 +1,1 @@
+Added translation for execution eviroment -> activities '{repository} was added' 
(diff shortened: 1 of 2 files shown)


himdel (approved):

himdel:
Tested with `localStorage.test_l10n = true` :+1: :

PR #1707, opened 2022-03-01
Search for sporadic failures 2 (group_management, execution_environments_use_in_controller.)

MilanPospisil:
Issue: AAH-1410

Focus mostly on execution_enviroments_use_in_controller.js. It tries to repair it - it now waits for the delete task to be completed (delete user command).

And it repairs also group management - there is new error due to alert PR. New alerts now contains name of the user and the test had very bad selector (no selector) that searches the user name at the whole page (for not contains). It worked before, but now the alert also contains the user name and it wrongly finds it there and fail.

### DIFF
--- a/test/cypress/support/commands.js
+++ b/test/cypress/support/commands.js
@@ -625,6 +625,9 @@ Cypress.Commands.add('syncRemoteContainer', {}, (name) => {
     '.pf-c-alert__title',
     `Sync started for execution environment "${name}".`,
   );
+  // wait for finish
+  cy.contains('a', 'detail page').click();
+  cy.contains('.title-box h1', 'Completed', { timeout: 20000 });
 });
 
 Cypress.Commands.add('deleteRegistries', {}, () => {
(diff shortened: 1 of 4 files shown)


test/cypress/support/commands.js
```diff
@@ -625,6 +625,9 @@ Cypress.Commands.add('syncRemoteContainer', {}, (name) => {
     '.pf-c-alert__title',
     `Sync started for execution environment "${name}".`,
   );
+  // wait for finish
+  cy.contains('a', 'detail page').click();
+  cy.contains('.title-box h1', 'Completed', { timeout: 20000 });
```
himdel, line 630: :eyes: .. but I agree this fixes the problem for now :+1: 

Definitely a FIXME for later though, seems like now that we know sync sometimes takes longer than the usual timeouts, we can fix it properly too...



we can wait for the alert, click that "detail page" link, land on a task detail...


 
and that page auto-updates to say Complete when done.

(But that's for a separate PR)

himdel (approved):

himdel:
Restarting tests just to make sure :)

himdel:
Merging, this fixes the group_management failure that's now happening everywhere :+1: 

But, see the failure above, execution_environment_use_in_controller is still failing sporadically.
(And so is collection_upload)

PR #1709, opened 2022-03-01
Test sync action in EE Remote Registries

jerabekjiri:
Issue: [AAH-1000](https://issues.redhat.com/browse/AAH-1000)

Testing successful syncing of the remote registry. 

I noticed we moved to `data-cy` attributes, so added this attribute as a prop to the `AlertList` component as well.

### DIFF
--- a/src/components/patternfly-wrappers/alert-list.tsx
+++ b/src/components/patternfly-wrappers/alert-list.tsx
@@ -11,6 +11,8 @@ interface IProps {
 
   /** Callback to close the alert at the given index */
   closeAlert: (alertIndex) => void;
+
+  'data-cy'?: string;
 }
 
 export class AlertType {
@@ -32,6 +34,7 @@ export class AlertList extends React.Component<IProps> {
           display: 'flex',
           flexDirection: 'column',
         }}
+        data-cy={`AlertList-${this.props['data-cy']}`}
       >
         {alerts.map((alert, i) => (
           <Alert
--- a/src/components/patternfly-wrappers/alert-list.tsx
+++ b/src/components/patternfly-wrappers/alert-list.tsx
@@ -11,8 +11,6 @@ interface IProps {
 
   /** Callback to close the alert at the given index */
   closeAlert: (alertIndex) => void;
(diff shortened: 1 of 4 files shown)


src/components/patternfly-wrappers/alert-list.tsx
```diff
@@ -32,6 +33,7 @@ export class AlertList extends React.Component<IProps> {
           display: 'flex',
           flexDirection: 'column',
         }}
+        data-cy={this.props['data-cy']}
```
brumik, line 36: I wanted to say maybe since it is in the AlertList we can do an auto prepend with the component name for consistency:
```
data-cy={`AlertList-${this.props.data-cy}`}
```

But this may cause maybe minor confusion, so I am not sure if it would be better. I leave it up to you.

brumik (approved):
This looks good for me, tests are passing :+1:

src/components/patternfly-wrappers/alert-list.tsx
jerabekjiri: Makes sense to put the source component in the first place :+1:

src/components/patternfly-wrappers/alert-list.tsx
```diff
@@ -32,6 +34,7 @@ export class AlertList extends React.Component<IProps> {
           display: 'flex',
           flexDirection: 'column',
         }}
+        data-cy={`AlertList-${this.props['data-cy']}`}
```
himdel, line 37: Can we just do `data-cy='AlertList'` and then use just that in tests?  Or `[data-cy=ExecutionEnvironmentRegistryList] [data-cy=AlertList]` and put the first data-cy on a real element?

I'm trying to prevent `data-cy` meaning different things when on a component, right now if you see a `data-cy='foo'` in the code, you can assume `[data-cy=foo]` will work in tests, this breaks that.

ZitaNemeckova (approved):
himdel's comment has be resolved. LGTM :+1:

@jerabekjiri Feel free to merge.

PR #1711, opened 2022-03-02
Fix overflow content in ImportConsole on My imports Page

jerabekjiri:
Issue: [AAH-396](https://issues.redhat.com/browse/AAH-396)

This fixes content overflow in `ImportConsole` component when the text is too long in `message-list` element or when the error pops up in `title-bar` element.

before:
Screenshot from 2022-03-02 11-58-56

after:
Screenshot from 2022-03-02 12-29-11

### DIFF
--- a/CHANGES/396.misc
+++ b/CHANGES/396.misc
@@ -1,0 +1,1 @@
+Fixed content overflow in ImportConsole component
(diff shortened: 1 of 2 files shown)


ShaiahWren (approved):
Works for me!

himdel (approved):

PR #1713, opened 2022-03-03
Improve testing of my-imports page

jerabekjiri:
Issue: [AAH-625](https://issues.redhat.com/browse/AAH-625)

Adding/improving `imports_filter.js` tests:
- verifying `completed` state in `ImportConsole` component
- fail to import existing collection test 
- collection exists (and redirection works) after clicking on the collection link 
- switching between namespaces shows correct collections

### DIFF
--- a/src/components/my-imports/import-console.tsx
+++ b/src/components/my-imports/import-console.tsx
@@ -28,6 +28,7 @@ interface IProps {
   setFollowMessages: (follow: boolean) => void;
   hideCollectionName?: boolean;
   collectionVersion?: CollectionVersion;
+  'data-cy'?: string;
 }
 
 export class ImportConsole extends React.Component<IProps> {
@@ -67,7 +68,10 @@ export class ImportConsole extends React.Component<IProps> {
       selectedImport.state === PulpStatus.waiting;
 
     return (
-      <div className='hub-import-console pf-c-content'>
+      <div
+        className='hub-import-console pf-c-content'
+        data-cy={`ImportConsole-${this.props['data-cy']}`}
+      >
         {this.renderTitle(selectedImport)}
         <div className='message-list'>
           <div
--- a/src/components/my-imports/import-console.tsx
+++ b/src/components/my-imports/import-console.tsx
@@ -28,7 +28,6 @@ interface IProps {
   setFollowMessages: (follow: boolean) => void;
   hideCollectionName?: boolean;
   collectionVersion?: CollectionVersion;
-  'data-cy'?: string;
 }
 
 export class ImportConsole extends React.Component<IProps> {
@@ -70,7 +69,7 @@ export class ImportConsole extends React.Component<IProps> {
     return (
       <div
         className='hub-import-console pf-c-content'
-        data-cy={`ImportConsole-${this.props['data-cy']}`}
+        data-cy={'ImportConsole'}
       >
         {this.renderTitle(selectedImport)}
         <div className='message-list'>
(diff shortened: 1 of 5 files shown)


test/cypress/integration/imports_filter.js
```diff
@@ -19,6 +25,56 @@ describe('Imports filter test', () => {
...
+      .contains(`test_namespace.${testCollection}`)
+      .click();
+    cy.contains(testCollection);
+  });
+
+  it('should be able to switch between namespaces', () => {
+    cy.get('[aria-label="Select namespace"]').select('test_namespace');
+    cy.get(`[data-cy="ImportList-row-${testCollection}"]`).should('be.visible');
+    cy.get('[aria-label="Select namespace"]').select('filter_test_namespace');
+    cy.get('[data-cy="ImportList-row-my_collection1"]').should('be.visible');
```
himdel, line 75: `"before each" hook for "partial filter for name is working."` is failing with a 403...

that sounds like we need to wait for all requests to happen in the previous test

src/components/my-imports/import-console.tsx
```diff
@@ -67,7 +68,10 @@ export class ImportConsole extends React.Component<IProps> {
       selectedImport.state === PulpStatus.waiting;
 
     return (
-      <div className='hub-import-console pf-c-content'>
+      <div
+        className='hub-import-console pf-c-content'
+        data-cy={`ImportConsole-${this.props['data-cy']}`}
```
himdel, line 73: so, doing it like this means looking at import-list or my-imports, people will see `data-cy={'MyImports'}` and expect *that* to work in tests .. yet `cy.get('[data-cy=MyImports]')` will fail, because we're only using it to create a longer data-cy.

Can we just go with nesting `data-cy`?
(As in, `data-cy='ImportConsole'` in ImportConsole root, `data-cy='MyImports` in my-imports but on an actual html element, the row in import-list is fine as is, and in tests we can still do `[data-cy="ImportConsole"] [data-cy="MyImports"]'.

WDYT?

jerabekjiri: I think this is the right approach, this way we can reuse attributes, but need to keep them consistent in other tests

test/cypress/integration/imports_filter.js
```diff
@@ -19,6 +25,56 @@ describe('Imports filter test', () => {
...
+    cy.galaxykit(`-i collection upload test_namespace ${testCollection}`);
+
+    cy.visit('/ui/my-imports?namespace=test_namespace');
+    cy.get(`[data-cy="ImportList-row-${testCollection}"]`).click();
+    cy.get('[data-cy="ImportConsole-MyImports"]').contains(
+      `test_namespace.${testCollection}`,
+    );
+    cy.get('[data-cy="ImportConsole-MyImports"] .title-bar').contains(
+      'Completed',
+      { timeout: 10000 },
```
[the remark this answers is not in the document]
jerabekjiri: @himdel maybe after merging this, I could create a ticket and work on this, if this isn't part of the `replace overused commands` issue, which shouldn't be?

src/components/my-imports/import-console.tsx
himdel: Can we call it `dataCyPrefix` then?

src/components/my-imports/import-list.tsx
```diff
@@ -28,6 +28,7 @@ interface IProps {
     keyword?: string;
     namespace?: string;
   };
+  'data-cy'?: string;
```
ZitaNemeckova, line 31: This is never used.

ZitaNemeckova (approved):
One cleaning comment but otherwise LGTM :+1: .

PR #1714, opened 2022-03-04
Fix "Publish container images" documentation link version - 2.0-ea -> 2.1

himdel:
Fixes AAH-1364

Exection Environments list:
The "Publish container images" button currently links to 4.3 documentation, updating to link to 4.4. (even for master, since 4.5 docs don't exist yet).

2.0 = 4.3
2.1 = 4.4
(2.2 = 4.5, but no such link exists yet)

### DIFF
--- a/CHANGES/1364.bug
+++ b/CHANGES/1364.bug
@@ -1,0 +1,1 @@
+Fix "Publish container images" documentation link version - 2.0-ea -> 2.1
--- a/src/containers/execution-environment-list/execution_environment_list.tsx
+++ b/src/containers/execution-environment-list/execution_environment_list.tsx
@@ -128,7 +128,7 @@ class ExecutionEnvironmentList extends React.Component<
         variant='link'
         onClick={() =>
           window.open(
-            'https://access.redhat.com/documentation/en-us/red_hat_ansible_automation_platform/2.0-ea/html-single/managing_containers_in_private_automation_hub/index',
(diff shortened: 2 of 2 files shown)


jerabekjiri:
@himdel I've seen this issue before and wanted to ask. Ideally, in the future, we would like the version to be changed dynamically based on the AAP version, right? Which could be resolved with this issue [AAP-1855](https://issues.redhat.com/browse/AAP-1855)

himdel:
Well, if we always knew what the version is, it would probably make sense to use it, but:

* AAP-1855 proposes that we would get an optional file with that information, and hide it if not available .. that suggests to me that it won't always be available
* the link would not work on any branch that was not already officially released - the docs link starts to work after all our code freezes for a Y-release (and linking to an older version of docs is better than linking to a 404)
* for some reason, 4.3 uses `2.0-ea` instead of `2.0`, but AFAIK the release is still just called 2.0

jerabekjiri:
I see, and we wouldn't be able to get the version number in the community version as you mentioned, thanks for explaining :relaxed:

himdel:
No problem :) You're definitely right that we still need something that causes us not to forget on the next release.
I've put it in my notes for now, but if you have any ideas on automating this... :) (maybe we can add a release script that detects we're incrementing the Y version and creates a jira issue, maybe the URL exists by the time we're doing the actual build and we can upgrade it then; maybe a test that checks the branch is stable-, more than a month old, and using the same link as the previous branch... I guess we'll have to see :))

PR #1738, opened 2022-03-09
Fix missing collection version number in collection deletion alert

jerabekjiri:
Issue: [AAH-1431](https://issues.redhat.com/browse/AAH-1431)

Fixes a small issue with the missing version number in Alert on **successful deleting collection**. I had to slightly rework the collection deletion code, because before we didn't need to show the collection version in alert and therefore it was null on deleting entire collection. 

before:
Screenshot from 2022-03-09 11-48-37

after:
Screenshot from 2022-03-09 11-51-11

### DIFF
--- a/CHANGES/1431.misc
+++ b/CHANGES/1431.misc
@@ -1,0 +1,1 @@
+Fixed and reworked collection deletion to fix missing collection version in alert
--- a/src/components/headers/collection-header.tsx
+++ b/src/components/headers/collection-header.tsx
@@ -68,12 +68,14 @@ interface IState {
     pageSize: number;
   };
   deleteCollection: CollectionDetailType;
-  collectionVersion: string | null;
   confirmDelete: boolean;
   alerts: AlertType[];
   redirect: string;
   noDependencies: boolean;
   isDeletionPending: boolean;
+
+  // entire collection or collection version is deleted
+  deletedCollectionType: 'version' | 'entire' | null;
 }
 
 export class CollectionHeader extends React.Component<IProps, IState> {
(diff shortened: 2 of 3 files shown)


ShaiahWren (approved):
The changes in this pr look reasonable and work as expected. :)

ShaiahWren:
@jerabekjiri , Could you go back and sign the first commit on this pr before merging? Thank you :)

himdel:
Hm, @jerabekjiri what's the actual change here?

All I'm seeing is that `deletedCollectionType === 'version'` is being used wherever `collectionVersion` was checked previously,
so unless clearing `collectionVersion` in `closeModal` is the actual change, I don't see how this can fix anything?

IMO the only place this could affect is the alert in `deleteCollection` - and that one is actually wrong, since we should not be claiming to have deleted a specific version when deleting the entire collection.

If that's the case, can we just fix the message instead?

jerabekjiri:
> IMO the only place this could affect is the alert in deleteCollection - and that one is actually wrong, since we should not be  claiming to have deleted a specific version when deleting the entire collection.
>
> If that's the case, can we just fix the message instead?

I think this answers it, I'll close this and open new PR with just fixing the delete message. Thanks @himdel, I guess next time I should put more thinking about the actual fix and not jump straight into refactoring :laughing:

himdel:
> If that's the case, can we just fix the message instead?

Never mind, addressed in https://github.com/ansible/ansible-hub-ui/pull/1859

jerabekjiri:
@himdel Thanks for fixing this, I was planning on doing it, but haven't found a time for that :disappointed:

PR #1758, opened 2022-03-14
Test cypress failures

himdel:
Seeing a lot of test fail in galaxykit now (https://github.com/ansible/ansible-hub-ui/runs/5530725419?check_suite_focus=true)
we're still installing from master, not 0.7.0 so presumably something broke in the latest merges,
but cypress cuts stdout and stderr on 200 chars :(.

(TODO better cy.exec logging)


himdel:
Closing, fixed in https://github.com/ansible/galaxykit/pull/34

PR #435, opened 2021-05-12
[pending api] NamespaceList, NamespaceCard - show "My namespace" label, Edit on editable namespaces

himdel:
Follow up to #396
Related to https://issues.redhat.com/browse/AAH-487 and https://issues.redhat.com/browse/AAH-1087

Design: 
image-2021-04-08-08-50-00-925

This changes the Namespaces (& My namespaces) list screen to:

* make each collection title clickable
* add a `My namespace` label for editable namespaces
* TODO add a kebab menu linking to edit-namespace Edit

---

WIP: this doesn't work with paging, needs to wait for an API change

PR #1766, opened 2022-03-15
 Collections - Ensure feature parity between list/detail screens

MilanPospisil:
https://issues.redhat.com/browse/AAH-1025

problem statement: 

collections:

    collection list has no item actions
    collection list on namespace detail has Upload new version, Deprecate (no Deletes)
    collection detail has Delete collection, Delete version (but no upload or deprecate)

acceptance criteria:

    tests needed?
    Collection List and detail have matching actions - except Delete version? (not possible to change version on the list)
    Collection list has Deprecate, Delete Collection, Upload new version
    Collection list on Namespace has Upload new version, Deprecate, Delete collection
    Collection detail has Delete collection, Delete version, Deprecate, Upload new version

This PR contains:
- Deprecated
- Upload New Version

It also adds basic tests for this functionality.

### DIFF
--- a/src/components/cards/cards.scss
+++ b/src/components/cards/cards.scss
@@ -49,10 +49,16 @@
   .logo-row {
     display: flex;
     justify-content: space-between;
+    padding-left: 15px;
+    padding-right: 5px;
 
     .icon {
       color: var(--pf-global--info-color--100);
       margin-right: 3px;
+    }
+
+    .left-container {
+      display: flex;
     }
 
     .hub-signature-badge {
--- a/src/components/cards/cards.scss
+++ b/src/components/cards/cards.scss
@@ -49,16 +49,11 @@
   .logo-row {
     display: flex;
     justify-content: space-between;
-    padding-left: 15px;
-    padding-right: 5px;
+    padding-right: 8px;
 
     .icon {
       color: var(--pf-global--info-color--100);
       margin-right: 3px;
-    }
-
-    .left-container {
-      display: flex;
     }
 
     .hub-signature-badge {
--- a/src/components/cards/collection-card.tsx
+++ b/src/components/cards/collection-card.tsx
@@ -55,13 +55,11 @@ export class CollectionCard extends React.Component<IProps> {
             image={namespace.avatar_url}
             size='40px'
             unlockWidth
+            flexGrow
           />
-
-          <div className='left-container'>
-            <TextContent>{this.getCertification(repo)}</TextContent>
-            <SignatureBadge isCompact signState={sign_state} />
-            {menu}
-          </div>
+          <TextContent>{this.getCertification(repo)}</TextContent>
+          <SignatureBadge isCompact signState={sign_state} />
+          {menu}
         </CardHeader>
         <CardHeader>
           <div className='name'>
--- a/src/components/cards/collection-card.tsx
+++ b/src/components/cards/collection-card.tsx
@@ -25,6 +25,7 @@ interface IProps extends CollectionListType {
   className?: string;
   footer?: React.ReactNode;
   repo?: string;
+  menu?: React.ReactNode;
 }
 
 export class CollectionCard extends React.Component<IProps> {
@@ -39,6 +40,7 @@ export class CollectionCard extends React.Component<IProps> {
       footer,
       repo,
       sign_state,
+      menu,
     } = this.props;
 
     const company = namespace.company || namespace.name;
@@ -54,8 +56,12 @@ export class CollectionCard extends React.Component<IProps> {
             size='40px'
             unlockWidth
           />
-          <TextContent>{this.getCertification(repo)}</TextContent>
-          <SignatureBadge isCompact signState={sign_state} />
+
+          <div className='left-container'>
+            <TextContent>{this.getCertification(repo)}</TextContent>
+            <SignatureBadge isCompact signState={sign_state} />
+            {menu}
+          </div>
         </CardHeader>
         <CardHeader>
           <div className='name'>
--- a/src/components/headers/collection-header.tsx
+++ b/src/components/headers/collection-header.tsx
@@ -37,15 +37,22 @@ import {
   DeleteModal,
   SignSingleCertificateModal,
   SignAllCertificatesModal,
+  ImportModal,
 } from 'src/components';
 
 import {
   CollectionAPI,
   CollectionDetailType,
   SignCollectionAPI,
+  CollectionListType,
+  MyNamespaceAPI,
 } from 'src/api';
 import { Paths, formatPath } from 'src/paths';
-import { waitForTask, canSign as canSignNS } from 'src/utilities';
+import {
+  waitForTask,
+  canSign as canSignNS,
+  getIdFromTask,
+} from 'src/utilities';
 import { ParamHelper } from 'src/utilities/param-helper';
 import { DateComponent } from '../date-component/date-component';
 import { Constants } from 'src/constants';
@@ -65,6 +72,7 @@ interface IProps {
   activeTab: string;
   className?: string;
   repo?: string;
+  reload: () => void;
 }
 
 interface IState {
@@ -83,6 +91,8 @@ interface IState {
   redirect: string;
   noDependencies: boolean;
   isDeletionPending: boolean;
+  updateCollection: CollectionListType;
+  showImportModal: boolean;
 }
 
 export class CollectionHeader extends React.Component<IProps, IState> {
@@ -108,6 +118,8 @@ export class CollectionHeader extends React.Component<IProps, IState> {
       redirect: null,
       noDependencies: false,
       isDeletionPending: false,
+      updateCollection: null,
+      showImportModal: false,
     };
   }
 
@@ -135,6 +147,8 @@ export class CollectionHeader extends React.Component<IProps, IState> {
       deleteCollection,
       confirmDelete,
       isDeletionPending,
+      showImportModal,
+      updateCollection,
     } = this.state;
 
     const numOfshownVersions = 10;
@@ -239,10 +253,49 @@ export class CollectionHeader extends React.Component<IProps, IState> {
           {t`Sign version ${collection.latest_version.version}`}
         </DropdownItem>
       ),
+      <DropdownItem
+        onClick={() => this.deprecate(collection)}
+        key='deprecate'
+        isDisabled={DEPLOYMENT_MODE === Constants.INSIGHTS_DEPLOYMENT_MODE}
+        description={
+          DEPLOYMENT_MODE === Constants.INSIGHTS_DEPLOYMENT_MODE
+            ? t`Temporarily disabled due to sync issues. (AAH-1237)`
+            : null
+        }
+      >
+        {collection['deprecated'] ? t`Undeprecate` : t`Deprecate`}
+      </DropdownItem>,
+      <DropdownItem
+        key='upload-collection-version'
+        onClick={() => this.checkUploadPrivilleges(collection)}
+        data-cy='upload-collection-version-dropdown'
+      >
+        {t`Upload new version`}
+      </DropdownItem>,
     ].filter(Boolean);
 
     return (
       <React.Fragment>
+        {showImportModal && (
+          <ImportModal
+            isOpen={showImportModal}
+            onUploadSuccess={() =>
+              this.setState({
+                redirect: formatPath(
+                  Paths.myImports,
+                  {},
+                  {
+                    namespace: updateCollection.namespace.name,
+                  },
+                ),
+              })
+            }
+            // onCancel
+            setOpen={(isOpen, warn) => this.toggleImportModal(isOpen, warn)}
+            collection={updateCollection}
+            namespace={updateCollection.namespace.name}
+          />
+        )}
         {canSign && (
           <>
             <SignAllCertificatesModal
@@ -467,7 +520,7 @@ export class CollectionHeader extends React.Component<IProps, IState> {
             ) : null
           }
         >
-          {collection.deprecated && (
+          {collection['deprecated'] && (
             <Alert
               variant='danger'
               isInline
@@ -503,6 +556,39 @@ export class CollectionHeader extends React.Component<IProps, IState> {
         </BaseHeader>
       </React.Fragment>
     );
+  }
+
+  private checkUploadPrivilleges(collection) {
+    const addAlert = () => {
+      const alert = new AlertType();
+      alert.title = t`You dont have rights to do this operation.`;
+      this.state.alerts.push(alert);
+
+      this.setState({
+        alerts: this.state.alerts,
+      });
+    };
+
+    MyNamespaceAPI.get(collection.namespace.name, {
+      include_related: 'my_permissions',
+    })
+      .then((value) => {
+        if (
+          value.data.related_fields.my_permissions.includes(
+            'galaxy.upload_to_namespace',
+          )
+        ) {
+          this.setState({
+            updateCollection: collection,
+            showImportModal: true,
+          });
+        } else {
+          addAlert();
+        }
+      })
+      .catch(() => {
+        addAlert();
+      });
   }
 
   private renderTabs(active) {
@@ -678,6 +764,39 @@ export class CollectionHeader extends React.Component<IProps, IState> {
       });
   };
 
+  private deprecate(collection) {
+    const operation = !collection.deprecated;
+    const operation_text = operation ? 'deprecated' : 'undeprecated';
+
+    CollectionAPI.setDeprecation(
+      collection,
+      operation,
+      this.context.selectedRepo,
+    )
+      .then((res) => {
+        const taskId = getIdFromTask(res.data.task);
+        waitForTask(taskId).then(() => {
+          if (this.props.reload) {
+            this.props.reload();
+          }
+        });
+      })
+      .catch((err) => {
+        const { status, statusText } = err.response;
+        this.setState({
+          collectionVersion: null,
+          alerts: [
+            ...this.state.alerts,
+            {
+              variant: 'danger',
+              title: t`Collection "${collection.name}" could not be "${operation_text}".`,
+              description: errorMessage(status, statusText),
+            },
+          ],
+        });
+      });
+  }
+
   private deleteCollectionVersion = (collectionVersion) => {
     const {
       deleteCollection,
@@ -688,7 +807,7 @@ export class CollectionHeader extends React.Component<IProps, IState> {
       deleteCollection,
     )
       .then((res) => {
-        const taskId = this.getIdFromTask(res.data.task);
+        const taskId = getIdFromTask(res.data.task);
         const name = deleteCollection.name;
 
         waitForTask(taskId).then(() => {
@@ -793,6 +912,18 @@ export class CollectionHeader extends React.Component<IProps, IState> {
       });
   };
 
+  private toggleImportModal(isOpen: boolean, warning?: string) {
+    const newAlerts = this.state.alerts;
+    if (warning) {
+      const newAlert = new AlertType();
+      newAlert.title = warning;
+      newAlert.variant = 'warning';
+      newAlerts.push();
+      this.setState({ alerts: newAlerts });
+    }
+    this.setState({ showImportModal: isOpen });
+  }
+
   private deleteCollection = () => {
     const {
       deleteCollection,
@@ -800,7 +931,7 @@ export class CollectionHeader extends React.Component<IProps, IState> {
     } = this.state;
     CollectionAPI.deleteCollection(this.context.selectedRepo, deleteCollection)
       .then((res) => {
-        const taskId = this.getIdFromTask(res.data.task);
+        const taskId = getIdFromTask(res.data.task);
 
         waitForTask(taskId).then(() => {
           this.context.setAlerts([
@@ -868,9 +999,6 @@ export class CollectionHeader extends React.Component<IProps, IState> {
       });
   }
 
-  private getIdFromTask(task) {
-    return task.match(/tasks\/([a-zA-Z0-9-]+)/i)[1];
-  }
   private closeModal = () => {
     this.setState({ deleteCollection: null });
   };
--- a/src/components/headers/collection-header.tsx
+++ b/src/components/headers/collection-header.tsx
@@ -775,6 +775,18 @@ export class CollectionHeader extends React.Component<IProps, IState> {
       .then((res) => {
         const taskId = parsePulpIDFromURL(res.data.task);
         return waitForTask(taskId).then(() => {
+          const title = !collection.deprecated
+            ? t`The collection "${collection.name}" has been successfully deprecated.`
+            : t`The collection "${collection.name}" has been successfully undeprecated.`;
+          this.setState({
+            alerts: [
+              ...this.state.alerts,
+              {
+                title: title,
+                variant: 'success',
+              },
+            ],
+          });
           if (this.props.reload) {
             this.props.reload();
           }
--- a/src/components/headers/collection-header.tsx
+++ b/src/components/headers/collection-header.tsx
@@ -774,7 +774,7 @@ export class CollectionHeader extends React.Component<IProps, IState> {
     )
       .then((res) => {
         const taskId = parsePulpIDFromURL(res.data.task);
-        waitForTask(taskId).then(() => {
+        return waitForTask(taskId).then(() => {
           if (this.props.reload) {
             this.props.reload();
           }
--- a/src/components/headers/collection-header.tsx
+++ b/src/components/headers/collection-header.tsx
@@ -51,7 +51,7 @@ import { Paths, formatPath } from 'src/paths';
 import {
   waitForTask,
   canSign as canSignNS,
-  getIdFromTask,
+  parsePulpIDFromURL,
 } from 'src/utilities';
 import { ParamHelper } from 'src/utilities/param-helper';
 import { DateComponent } from '../date-component/date-component';
@@ -773,7 +773,7 @@ export class CollectionHeader extends React.Component<IProps, IState> {
       this.context.selectedRepo,
     )
       .then((res) => {
-        const taskId = getIdFromTask(res.data.task);
+        const taskId = parsePulpIDFromURL(res.data.task);
         waitForTask(taskId).then(() => {
           if (this.props.reload) {
             this.props.reload();
@@ -808,7 +808,7 @@ export class CollectionHeader extends React.Component<IProps, IState> {
       deleteCollection,
     )
       .then((res) => {
-        const taskId = getIdFromTask(res.data.task);
+        const taskId = parsePulpIDFromURL(res.data.task);
         const name = deleteCollection.name;
 
         waitForTask(taskId).then(() => {
@@ -929,7 +929,7 @@ export class CollectionHeader extends React.Component<IProps, IState> {
     } = this.state;
     CollectionAPI.deleteCollection(this.context.selectedRepo, deleteCollection)
       .then((res) => {
-        const taskId = getIdFromTask(res.data.task);
+        const taskId = parsePulpIDFromURL(res.data.task);
 
         waitForTask(taskId).then(() => {
           this.context.setAlerts([
--- a/src/components/headers/collection-header.tsx
+++ b/src/components/headers/collection-header.tsx
@@ -560,12 +560,14 @@ export class CollectionHeader extends React.Component<IProps, IState> {
 
   private checkUploadPrivilleges(collection) {
     const addAlert = () => {
-      const alert = new AlertType();
-      alert.title = t`You dont have rights to do this operation.`;
-      this.state.alerts.push(alert);
-
       this.setState({
-        alerts: this.state.alerts,
+        alerts: [
+          ...this.state.alerts,
+          {
+            title: t`You don't have rights to do this operation.`,
+            variant: 'warning',
+          },
+        ],
       });
     };
 
@@ -765,12 +767,9 @@ export class CollectionHeader extends React.Component<IProps, IState> {
   };
 
   private deprecate(collection) {
-    const operation = !collection.deprecated;
-    const operation_text = operation ? 'deprecated' : 'undeprecated';
-
     CollectionAPI.setDeprecation(
       collection,
-      operation,
+      !collection.deprecated,
       this.context.selectedRepo,
     )
       .then((res) => {
@@ -789,7 +788,9 @@ export class CollectionHeader extends React.Component<IProps, IState> {
             ...this.state.alerts,
             {
               variant: 'danger',
-              title: t`Collection "${collection.name}" could not be "${operation_text}".`,
+              title: !collection.deprecated
+                ? t`Collection "${collection.name}" could not be deprecated.`
+                : t`Collection "${collection.name}" could not be undeprecated.`,
               description: errorMessage(status, statusText),
             },
           ],
@@ -913,13 +914,10 @@ export class CollectionHeader extends React.Component<IProps, IState> {
   };
 
   private toggleImportModal(isOpen: boolean, warning?: string) {
-    const newAlerts = this.state.alerts;
     if (warning) {
-      const newAlert = new AlertType();
-      newAlert.title = warning;
-      newAlert.variant = 'warning';
-      newAlerts.push();
-      this.setState({ alerts: newAlerts });
+      this.setState({
+        alerts: [...this.state.alerts, { title: warning, variant: 'warning' }],
+      });
     }
     this.setState({ showImportModal: isOpen });
   }
--- a/src/components/headers/collection-header.tsx
+++ b/src/components/headers/collection-header.tsx
@@ -520,7 +520,7 @@ export class CollectionHeader extends React.Component<IProps, IState> {
             ) : null
           }
         >
-          {collection['deprecated'] && (
+          {collection.deprecated && (
             <Alert
               variant='danger'
               isInline
--- a/src/components/headers/collection-header.tsx
+++ b/src/components/headers/collection-header.tsx
@@ -263,7 +263,7 @@ export class CollectionHeader extends React.Component<IProps, IState> {
             : null
         }
       >
-        {collection['deprecated'] ? t`Undeprecate` : t`Deprecate`}
+        {collection.deprecated ? t`Undeprecate` : t`Deprecate`}
       </DropdownItem>,
       <DropdownItem
         key='upload-collection-version'
--- a/src/containers/collection-detail/base.ts
+++ b/src/containers/collection-detail/base.ts
@@ -25,7 +25,7 @@ export function loadCollection(
     forceReload,
   )
     .then((result) => {
-      CollectionAPI.list(
+      return CollectionAPI.list(
         {
           name: this.props.match.params['collection'],
         },
--- a/src/containers/collection-detail/base.ts
+++ b/src/containers/collection-detail/base.ts
@@ -25,7 +25,15 @@ export function loadCollection(
     forceReload,
   )
     .then((result) => {
-      this.setState({ collection: result }, callback);
+      CollectionAPI.list(
+        {
+          name: this.props.match.params['collection'],
+        },
+        this.context.selectedRepo,
+      ).then((collections) => {
+        result.deprecated = collections.data.data[0].deprecated;
+        this.setState({ collection: result }, callback);
+      });
     })
     .catch(() => {
       this.props.history.push(Paths.notFound);
--- a/src/containers/search/search.scss
+++ b/src/containers/search/search.scss
@@ -26,6 +26,24 @@
       flex-wrap: wrap;
       align-content: flex-start;
 
+      div[aria-label='collection-kebab'] {
+        button[aria-label='Actions'] {
+          transform: scaleY(0.7);
+          position: relative;
+          top: -10px;
+          padding-left: 8px;
+        }
+
+        ul {
+          position: absolute;
+          top: 20px;
+        }
+      }
+
+      .hub-signature-badge {
+        height: 16px;
+      }
+
       .card {
         margin-left: 24px;
         margin-bottom: 24px;
@@ -62,7 +80,7 @@
     flex-shrink: 0;
   }
 
-  .sync-toggle {
+  .list .sync-toggle {
     margin-right: 10px;
   }
 }
--- a/src/containers/search/search.scss
+++ b/src/containers/search/search.scss
@@ -21,28 +21,10 @@
     padding-top: 24px;
     flex-grow: 1;
 
-    .cards {
+    .hub-cards {
       display: flex;
       flex-wrap: wrap;
       align-content: flex-start;
-
-      div[aria-label='collection-kebab'] {
-        button[aria-label='Actions'] {
-          transform: scaleY(0.7);
-          position: relative;
-          top: -10px;
-          padding-left: 8px;
-        }
-
-        ul {
-          position: absolute;
-          top: 20px;
-        }
-      }
-
-      .hub-signature-badge {
-        height: 16px;
-      }
 
       .card {
         margin-left: 24px;
@@ -54,7 +36,7 @@
       width: 100%;
     }
 
-    .list {
+    .hub-list {
       margin-bottom: 24px;
       margin-left: 24px;
       margin-right: 24px;
@@ -68,6 +50,10 @@
         margin-top: -17px;
         margin-bottom: -17px;
       }
+
+      .sync-toggle {
+        margin-right: 10px;
+      }
     }
 
     .empty {
@@ -78,9 +64,5 @@
   .footer {
     border-top: 1px solid #d8d8d8;
     flex-shrink: 0;
-  }
-
-  .list .sync-toggle {
-    margin-right: 10px;
   }
 }
--- a/src/containers/search/search.scss
+++ b/src/containers/search/search.scss
@@ -61,4 +61,8 @@
     border-top: 1px solid #d8d8d8;
     flex-shrink: 0;
   }
+
+  .sync-toggle {
+    margin-right: 10px;
+  }
 }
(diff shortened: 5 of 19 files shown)


src/components/headers/collection-header.tsx
```diff
@@ -676,6 +762,39 @@ export class CollectionHeader extends React.Component<IProps, IState> {
       });
   };
 
+  private deprecate(collection) {
+    const operation = !collection.deprecated;
+    const operation_text = operation ? 'deprecated' : 'undeprecated';
```
ZitaNemeckova, line 767: Missing translations.

ZitaNemeckova:
Before:
Screenshot 2022-05-02 at 14 06 46
Screenshot 2022-05-02 at 14 39 47
(Old master without signing funcionality)


After:
Screenshot 2022-05-02 at 14 26 07
Screenshot 2022-05-02 at 14 44 52

src/components/headers/collection-header.tsx
```diff
@@ -791,14 +910,26 @@ export class CollectionHeader extends React.Component<IProps, IState> {
...
   };
 
+  private toggleImportModal(isOpen: boolean, warning?: string) {
+    const newAlerts = this.state.alerts;
+    if (warning) {
+      const newAlert = new AlertType();
+      newAlert.title = warning;
+      newAlert.variant = 'warning';
+      newAlerts.push();
+      this.setState({ alerts: newAlerts });
```
ZitaNemeckova, line 920: `this.setState({ alerts: this.state.alerts.push(newAlert) });` could save you few lines :)

src/containers/search/search.tsx
```diff
@@ -173,6 +231,18 @@ class Search extends React.Component<RouteComponentProps, IState> {
     );
   }
 
+  private toggleImportModal(isOpen: boolean, warning?: string) {
+    const newAlerts = this.state.alerts;
+    if (warning) {
+      const newAlert = new AlertType();
+      newAlert.title = warning;
+      newAlert.variant = 'warning';
+      newAlerts.push();
```
ZitaNemeckova, line 240: Nothing is pushed.

src/containers/search/search.tsx
```diff
@@ -228,6 +366,39 @@ class Search extends React.Component<RouteComponentProps, IState> {
     );
   }
 
+  private checkUploadPrivilleges(collection) {
+    const addAlert = () => {
+      const alert = new AlertType();
+      alert.title = t`You dont have rights to do this operation.`;
```
ZitaNemeckova, line 372: Typo `dont` => `don't`

src/components/headers/collection-header.tsx
```diff
@@ -503,6 +556,39 @@ export class CollectionHeader extends React.Component<IProps, IState> {
     );
   }
 
+  private checkUploadPrivilleges(collection) {
+    const addAlert = () => {
+      const alert = new AlertType();
+      alert.title = t`You dont have rights to do this operation.`;
```
ZitaNemeckova, line 562: Typo `dont` => `don't`

src/components/headers/collection-header.tsx
```diff
@@ -503,6 +556,39 @@ export class CollectionHeader extends React.Component<IProps, IState> {
...
+      this.state.alerts.push(alert);
```
ZitaNemeckova, line 563: Let's not change `state`  this way. Use `setState` instead.

src/containers/search/search.tsx
MilanPospisil: That was not typo, rather my bad english :-D

src/components/headers/collection-header.tsx
MilanPospisil: I know, that was very bad mistake.

src/components/headers/collection-header.tsx
```diff
@@ -503,6 +556,37 @@ export class CollectionHeader extends React.Component<IProps, IState> {
     );
   }
 
+  private checkUploadPrivilleges(collection) {
+    const addAlert = () => {
+      const alert = new AlertType();
```
himdel, line 561: alerts without variant should not be valid,
and `new AlertType` is not needed...


```suggestion
      const alert = { title: t`You don't have rights to do this operation.`, variant: 'error' };
```

src/components/headers/collection-header.tsx
```diff
@@ -791,14 +908,24 @@ export class CollectionHeader extends React.Component<IProps, IState> {
       });
   };
 
+  private toggleImportModal(isOpen: boolean, warning?: string) {
+    if (warning) {
+      const newAlert = new AlertType();
+      newAlert.title = warning;
+      newAlert.variant = 'warning';
+      this.setState({ alerts: [...this.state.alerts, newAlert] });
```
himdel, line 916:
```suggestion
      const newAlert = {
        title: warning,
        variant: 'warning',
      };
      this.setState({ alerts: [...this.state.alerts, newAlert] });
```

src/containers/search/search.tsx
```diff
@@ -173,6 +231,16 @@ class Search extends React.Component<RouteComponentProps, IState> {
     );
   }
 
+  private toggleImportModal(isOpen: boolean, warning?: string) {
+    if (warning) {
+      const newAlert = new AlertType();
+      newAlert.title = warning;
+      newAlert.variant = 'warning';
```
himdel, line 238: (and here)

src/containers/search/search.tsx
```diff
@@ -212,11 +280,79 @@ class Search extends React.Component<RouteComponentProps, IState> {
...
+            ariaLabel='collection-kebab'
+          />
+        </React.Fragment>
+      );
+    }
     if (!synclist) {
-      return null;
+      return menu;
     }
+
```
himdel, line 355: this skips rendering the menu when synclists are enabled

src/containers/search/search.tsx
```diff
@@ -228,6 +364,39 @@ class Search extends React.Component<RouteComponentProps, IState> {
     );
   }
 
+  private checkUploadPrivilleges(collection) {
+    const addAlert = () => {
+      const alert = new AlertType();
+      alert.title = t`You don't have rights to do this operation.`;
```
himdel, line 370: (and here)

test/cypress/integration/collection_upload.js
```diff
@@ -4,21 +4,55 @@ describe('Collection Upload Tests', () => {
 
   before(() => {
     cy.login();
-
+    //cy.deleteNamespacesAndCollections();
     cy.deleteTestGroups();
     cy.deleteTestUsers();
-    cy.galaxykit('collection delete', 'ansible', 'network');
-    cy.galaxykit('collection delete', 'testspace', 'testcollection');
```
himdel, line 11: no longer doing cleanup here?

src/components/headers/collection-header.tsx
```diff
@@ -676,6 +760,39 @@ export class CollectionHeader extends React.Component<IProps, IState> {
...
+      })
+      .catch((err) => {
+        const { status, statusText } = err.response;
+        this.setState({
+          collectionVersion: null,
+          alerts: [
+            ...this.state.alerts,
+            {
+              variant: 'danger',
+              title: t`Collection "${collection.name}" could not be "${operation_text}".`,
```
himdel, line 788:
```suggestion
              title: !collection.deprecated ? t`Collection "${collection.name}" could not be deprecated.` : t`Collection "${collection.name}" could not be undeprecated.`,
```

computing the verb of the sentence could be problematic for translations

src/containers/search/search.tsx
```diff
@@ -212,11 +280,79 @@ class Search extends React.Component<RouteComponentProps, IState> {
...
+        });
+      })
+      .catch((err) => {
+        const { status, statusText } = err.response;
+        this.setState({
+          alerts: [
+            ...this.state.alerts,
+            {
+              variant: 'danger',
+              title: t`Collection "${collection.name}" could not be "${operation_text}".`,
```
himdel, line 304: (and here .. computed verb)

src/components/headers/collection-header.tsx
MilanPospisil: But this causes error: Type '{ title: string; variant: string; }' is not assignable to type 'AlertType'
Also variant seems to be restricted to this: AlertType.variant: "default" | "success" | "danger" | "warning" | "info"

src/containers/search/search.tsx
MilanPospisil: I will need help with this. Tests are needed for this, and I don't know how this should behave.

test/cypress/integration/collection_upload.js
MilanPospisil: This was valid back then when we did not have ability to clear all collections. Now we are deleting all of them in cy.deleteNamespacesAndCollections(), so there is no need to call additional deletes.

src/components/headers/collection-header.tsx
himdel: oh, sorry, same issue .. if you change variant to one of those, the first error disappears as well

danger is the red one :)

test/cypress/integration/collection_upload.js
himdel: aah, makes sense :)

src/containers/search/search.tsx
himdel: Well, you can force the sync toggle enabled this way...

```diff
   private load() {
     this.queryCollections();
+    this.setState({ synclist: { collections: [] } as any }); // DEBUG
```

and then it looks like this...




----

now, first thing, we should probably not mix the synclist toggle `renderSyncToggle` with the dropdown menu, so what about something like...

```diff
@@ -271,7 +272,7 @@ class Search extends React.Component<RouteComponentProps, IState> {
               className='card'
               key={c.id}
               {...c}
-              footer={this.renderSyncToggle(c.name, c.namespace.name, false, c)}
+              footer={this.renderCardFooter(c)}
               repo={this.context.selectedRepo}
             />
           );
```

```diff
@@ -440,12 +443,7 @@ class Search extends React.Component<RouteComponentProps, IState> {
                 showNamespace={true}
                 key={c.id}
                 {...c}
-                controls={this.renderSyncToggle(
-                  c.name,
-                  c.namespace.name,
-                  true,
-                  c,
-                )}
+                controls={this.renderCollectionControls(c)}
                 repo={this.context.selectedRepo}
               />
             ))}
```

```diff
-  private renderSyncToggle(
-    name: string,
-    namespace: string,
-    list: boolean,
-    collection,
-  ): React.ReactNode {
+  private renderSyncToggle(name: string, namespace: string): React.ReactNode {
     const { synclist } = this.state;
 
-    let menu = null;
-    if (list) {
-      menu = (
-        <React.Fragment>
-          <Button
-            onClick={() => this.checkUploadPrivilleges(collection)}
-            variant='secondary'
-          >
-            {t`Upload new version`}
-          </Button>
-          <StatefulDropdown
-            items={[
-              <DropdownItem
-                onClick={() => this.handleControlClick(collection)}
-                key='deprecate'
-                isDisabled={
-                  DEPLOYMENT_MODE === Constants.INSIGHTS_DEPLOYMENT_MODE
-                }
-                description={
-                  DEPLOYMENT_MODE === Constants.INSIGHTS_DEPLOYMENT_MODE
-                    ? t`Temporarily disabled due to sync issues. (AAH-1237)`
-                    : null
-                }
-              >
-                {collection.deprecated ? t`Undeprecate` : t`Deprecate`}
-              </DropdownItem>,
-            ]}
-            ariaLabel='collection-kebab'
-          />
-        </React.Fragment>
-      );
-    }
-
     if (!synclist) {
-      return menu;
+      return null;
     }
 
     return (
@@ -365,6 +328,52 @@ class Search extends React.Component<RouteComponentProps, IState> {
     );
   }
 
+  private renderCardFooter(collection) {
+    return this.renderSyncToggle(collection.name, collection.namespace.name);
+  }
+
+  private renderCollectionControls(collection) {
+    const { synclist } = this.state;
+
+    // FIXME allow showing both
+    return synclist
+      ? this.renderSyncToggle(collection.name, collection.namespace.name)
+      : this.renderMenu(collection);
+  }
+
+  private renderMenu(collection) {
+    return (
+      <>
+        <Button
+          onClick={() => this.checkUploadPrivilleges(collection)}
+          variant='secondary'
+        >
+          {t`Upload new version`}
+        </Button>
+
+        <StatefulDropdown
+          items={[
+            <DropdownItem
+              onClick={() => this.handleControlClick(collection)}
+              key='deprecate'
+              isDisabled={
+                DEPLOYMENT_MODE === Constants.INSIGHTS_DEPLOYMENT_MODE
+              }
+              description={
+                DEPLOYMENT_MODE === Constants.INSIGHTS_DEPLOYMENT_MODE
+                  ? t`Temporarily disabled due to sync issues. (AAH-1237)`
+                  : null
+              }
+            >
+              {collection.deprecated ? t`Undeprecate` : t`Deprecate`}
+            </DropdownItem>,
+          ]}
+          ariaLabel='collection-kebab'
+        />
+      </>
+    );
+  }
+
```

himdel: after that, maybe we don't need to keep the original layout, and can put the button inside the dropdown .. and that allows us to display both .. with a bit of style tweaking even in the card view...

```diff
diff --git a/src/containers/search/search.tsx b/src/containers/search/search.tsx
index 1d4c3673..d107e958 100644
--- a/src/containers/search/search.tsx
+++ b/src/containers/search/search.tsx
@@ -329,48 +329,49 @@ class Search extends React.Component<RouteComponentProps, IState> {
   }
 
   private renderCardFooter(collection) {
-    return this.renderSyncToggle(collection.name, collection.namespace.name);
+    return (
+      <>
+        <div style={{ display: 'inline-block' }}>
+          {this.renderSyncToggle(collection.name, collection.namespace.name)}
+        </div>
+        <div style={{ display: 'inline-block' }}>
+          {this.renderMenu(collection)}
+        </div>
+      </>
+    );
   }
 
   private renderCollectionControls(collection) {
-    const { synclist } = this.state;
-
-    // FIXME allow showing both
-    return synclist
-      ? this.renderSyncToggle(collection.name, collection.namespace.name)
-      : this.renderMenu(collection);
+    return (
+      <>
+        {this.renderSyncToggle(collection.name, collection.namespace.name)}
+        {this.renderMenu(collection)}
+      </>
+    );
   }
 
   private renderMenu(collection) {
     return (
-      <>
-        <Button
-          onClick={() => this.checkUploadPrivilleges(collection)}
-          variant='secondary'
-        >
-          {t`Upload new version`}
-        </Button>
-
-        <StatefulDropdown
-          items={[
-            <DropdownItem
-              onClick={() => this.handleControlClick(collection)}
-              key='deprecate'
-              isDisabled={
-                DEPLOYMENT_MODE === Constants.INSIGHTS_DEPLOYMENT_MODE
-              }
-              description={
-                DEPLOYMENT_MODE === Constants.INSIGHTS_DEPLOYMENT_MODE
-                  ? t`Temporarily disabled due to sync issues. (AAH-1237)`
-                  : null
-              }
-            >
-              {collection.deprecated ? t`Undeprecate` : t`Deprecate`}
-            </DropdownItem>,
-          ]}
-          ariaLabel='collection-kebab'
-        />
-      </>
+      <StatefulDropdown
+        items={[
+          <DropdownItem
+            onClick={() => this.handleControlClick(collection)}
+            key='deprecate'
+            isDisabled={DEPLOYMENT_MODE === Constants.INSIGHTS_DEPLOYMENT_MODE}
+            description={
+              DEPLOYMENT_MODE === Constants.INSIGHTS_DEPLOYMENT_MODE
+                ? t`Temporarily disabled due to sync issues. (AAH-1237)`
+                : null
+            }
+          >
+            {collection.deprecated ? t`Undeprecate` : t`Deprecate`}
+          </DropdownItem>,
+          <DropdownItem onClick={() => this.checkUploadPrivilleges(collection)}>
+            {t`Upload new version`}
+          </DropdownItem>,
+        ]}
+        ariaLabel='collection-kebab'
+      />
     );
   }
 
```




(the card view needs a bit more work to right-align the kebab button though)

src/components/headers/collection-header.tsx
MilanPospisil: I tried yesterday and today again and still:
TS2322: Type '{ title: string; variant: string; }' is not assignable to type 'AlertType'.
ui_1           |     561 |       const alert = { title: t`You don't have rights to do this operation.`, variant: 'warning' };
ui_1           |     562 |       this.setState({
ui_1           |   > 563 |         alerts: [...this.state.alerts, alert],
ui_1           |         |                                        ^^^^^
ui_1           |     564 |       });

src/containers/search/search.tsx
MilanPospisil: In the list view, it looks like this and this look good enough for me. What is your opinion?
sync

MilanPospisil: And it seems, I forgot to add menu to the Card view :-D So you cant see it even without sync enabled...

MilanPospisil: Hmm... the menu looks weird when there is no sync list and with Upload new version buttion, there is not enough space and it is rendered with two rows (menu is under upload button and it looks weird). What about to insert somewhere directly into the card - for example next to the card title on the right?

himdel: the screenshot .. the ordering is wrong, we can't have sync after the kebab, the kebab must go last .. otherwise it could work :)

re menu looks weird.. that's the card view, right? You're right, sync and the kebab menu should be separate there, and the kebab should live next to the title .. we actually have designs for that for namespaces (https://github.com/ansible/ansible-hub-ui/pull/435), so hopefully it works for collection cards as well :)

src/components/headers/collection-header.tsx
himdel: ah, for some reason TS doesnt see the types as compatible when going through a variable... surprising :)

This worked for me...

```diff
diff --git a/src/components/headers/collection-header.tsx b/src/components/headers/collection-header.tsx
index d1c7ec78..f1fc2ccd 100644
--- a/src/components/headers/collection-header.tsx
+++ b/src/components/headers/collection-header.tsx
@@ -558,10 +558,14 @@ export class CollectionHeader extends React.Component<IProps, IState> {
 
   private checkUploadPrivilleges(collection) {
     const addAlert = () => {
-      const alert = new AlertType();
-      alert.title = t`You don't have rights to do this operation.`;
       this.setState({
-        alerts: [...this.state.alerts, alert],
+        alerts: [
+          ...this.state.alerts,
+          {
+            title: t`You don't have rights to do this operation.`,
+            variant: 'danger',
+          },
+        ],
       });
     };
 
@@ -909,10 +913,16 @@ export class CollectionHeader extends React.Component<IProps, IState> {
 
   private toggleImportModal(isOpen: boolean, warning?: string) {
     if (warning) {
-      const newAlert = new AlertType();
-      newAlert.title = warning;
-      newAlert.variant = 'warning';
-      this.setState({ alerts: [...this.state.alerts, newAlert] });
+      this.setState({
+        alerts: [
+          ...this.state.alerts,
+          {
+            title: warning,
+            variant: 'warning',
+          },
+        ],
+      });
     }
     this.setState({ showImportModal: isOpen });
   }
```

but also, this is an alternative...

```diff
diff --git a/src/components/headers/collection-header.tsx b/src/components/headers/collection-header.tsx
index d1c7ec78..75fb41ff 100644
--- a/src/components/headers/collection-header.tsx
+++ b/src/components/headers/collection-header.tsx
@@ -558,8 +558,10 @@ export class CollectionHeader extends React.Component<IProps, IState> {
 
   private checkUploadPrivilleges(collection) {
     const addAlert = () => {
-      const alert = new AlertType();
-      alert.title = t`You don't have rights to do this operation.`;
+      const alert: AlertType = {
+        title: t`You don't have rights to do this operation.`,
+        variant: 'danger',
+      };
       this.setState({
         alerts: [...this.state.alerts, alert],
       });
@@ -909,9 +911,10 @@ export class CollectionHeader extends React.Component<IProps, IState> {
 
   private toggleImportModal(isOpen: boolean, warning?: string) {
     if (warning) {
-      const newAlert = new AlertType();
-      newAlert.title = warning;
-      newAlert.variant = 'warning';
+      const newAlert: AlertType = {
+        title: warning,
+        variant: 'warning',
+      };
       this.setState({ alerts: [...this.state.alerts, newAlert] });
     }
     this.setState({ showImportModal: isOpen });
```

src/containers/search/search.tsx
MilanPospisil: And for list:
screenshoot

MilanPospisil:
And one question @himdel . I think it will be good to include also test for failed collection deprecation/undeprecation. I dont know how to test that alert, in what situations it will fail? The only solution that come to my mind is include typo into deprecate URL, but that is good only for manual testing.

src/containers/search/search.tsx
MilanPospisil: The Upload new version button is not present in Cards (no space), but this operation is included in the menu instead.

himdel:
It's enough to test that it worked - check for success alert only. (you can make the test a part of https://issues.redhat.com/browse/AAH-1578)

src/containers/search/search.tsx
himdel: the list view :+1: 

but for the cards view, that positioning conflicts with the Signed label, I'd put the kebab under it, next to title, not logo..

but this may be a good time to involve UX here...  (except I don't have any of their github handles)

tiyiprh:
Here is our proposed solution for the card view. The spacing between the "signed" label and kebab should be 8 px. We should also include ratio constraints to the logo, so that it doesn't bleed into the label and kebab area. 

collections-card view

src/components/headers/collection-header.tsx
MilanPospisil: Done.

src/containers/search/search.tsx
MilanPospisil: Is this ok? @tiyiprh @himdel 
screenshoot

MilanPospisil: Oh, I see now that signed logo is smaller. Tell me please at least if alignment to the left and right is ok.

MilanPospisil: What about this? I think it looks like you wanted. There should be 8px space, at least it is set in css. But I dont know if @himdel will approve how it is done in css :)
screenshoot

MilanPospisil: I fixed today some issues that were not visible in the screenshoots (menu items were also scaled to 70% and positioned badly and the scale also overrided list view).

MilanPospisil:
Do you please know @himdel why the eslint is throwing error at my css changes?

src/containers/search/search.tsx
himdel: Nope, you're not actually fixing the problem where the logo height affects the spacing of everything else, you're just scaling everything down, including a tiny kebab.

`scale` is never a good idea .. if you're trying to solve the label height, just add a flexitem so that it can size itself separately

src/components/cards/cards.scss
```diff
@@ -49,12 +49,18 @@
...
     justify-content: space-between;
+    padding-left: 15px;
+    padding-right: 5px;
 
     .icon {
       color: var(--pf-global--info-color--100);
       margin-right: 3px;
     }
 
+    .left-container {
```
himdel, line 60:
```suggestion
    .hub-left-container {
```

(let's not add new unprefix classnames)

src/components/cards/collection-card.tsx
```diff
@@ -54,8 +56,12 @@ export class CollectionCard extends React.Component<IProps> {
             size='40px'
             unlockWidth
           />
-          <TextContent>{this.getCertification(repo)}</TextContent>
-          <SignatureBadge isCompact signState={sign_state} />
+
+          <div className='left-container'>
```
himdel, line 60:
```suggestion
          <div className='hub-left-container'>
```

src/components/cards/cards.scss
```diff
@@ -49,12 +49,18 @@
   .logo-row {
     display: flex;
     justify-content: space-between;
+    padding-left: 15px;
```
himdel, line 52: What's up with this? Are we changing the *left* padding for the logo as well?

src/components/cards/collection-card.tsx
himdel: though given the css, this should really be
```suggestion
          <Flex>
```

src/components/headers/collection-header.tsx
```diff
@@ -676,6 +764,38 @@ export class CollectionHeader extends React.Component<IProps, IState> {
...
 
+  private deprecate(collection) {
+    CollectionAPI.setDeprecation(
+      collection,
+      !collection.deprecated,
+      this.context.selectedRepo,
+    )
+      .then((res) => {
+        const taskId = getIdFromTask(res.data.task);
+        waitForTask(taskId).then(() => {
```
himdel, line 775: missing return?

src/containers/collection-detail/base.ts
```diff
@@ -25,7 +25,15 @@ export function loadCollection(
     forceReload,
   )
     .then((result) => {
-      this.setState({ collection: result }, callback);
+      CollectionAPI.list(
```
himdel, line 28: missing return?

src/containers/search/search.scss
```diff
@@ -26,6 +26,24 @@
       flex-wrap: wrap;
       align-content: flex-start;
 
+      div[aria-label='collection-kebab'] {
+        button[aria-label='Actions'] {
```
himdel, line 30: (this is probably going away when you remove the scale code, but if not...)

Don't use attribute values in css selectors, add a new classname instead

src/containers/search/search.scss
```diff
@@ -61,4 +79,8 @@
     border-top: 1px solid #d8d8d8;
     flex-shrink: 0;
   }
+
+  .list .sync-toggle {
```
himdel, line 83: can we use `.list-container .sync-toggle` here? I know `list` is an existing class used in search.tsx, but it's a terrible name :) I think `list-container` should work the same way, but is a bit more unique :)

(Or rename `list` to `hub-collections-list` or something.)

src/containers/search/search.tsx
```diff
@@ -71,19 +82,44 @@ class Search extends React.Component<RouteComponentProps, IState> {
...
+  }
+
+  private load() {
     this.queryCollections();
 
     if (DEPLOYMENT_MODE === Constants.INSIGHTS_DEPLOYMENT_MODE) {
       this.getSynclist();
     }
+
+    //this.setState({ synclist: { collections: [] } as any }); // DEBUG
```
himdel, line 103: (remember to remove this before merging)

src/containers/search/search.tsx
```diff
@@ -203,31 +272,149 @@ class Search extends React.Component<RouteComponentProps, IState> {
...
-  private renderSyncToggle(name: string, namespace: string): React.ReactNode {
+  private handleControlClick(collection) {
+    CollectionAPI.setDeprecation(
+      collection,
+      !collection.deprecated,
+      this.context.selectedRepo,
+    )
+      .then((res) => {
+        const taskId = getIdFromTask(res.data.task);
+        waitForTask(taskId).then(() => {
```
himdel, line 293: missing return

src/containers/search/search.tsx
```diff
@@ -203,31 +272,149 @@ class Search extends React.Component<RouteComponentProps, IState> {
...
+    name: string,
+    namespace: string,
+    list: boolean,
+    collection,
+  ): React.ReactNode {
     const { synclist } = this.state;
+
     if (!synclist) {
-      return null;
+      return list ? this.renderMenu(list, collection) : null;
```
himdel, line 366: why does renderSyncToggle also render a menu?

Sounds like the ordering is wrong, see https://github.com/ansible/ansible-hub-ui/pull/1766/files#r865533277

src/containers/search/search.tsx
```diff
@@ -203,31 +272,149 @@ class Search extends React.Component<RouteComponentProps, IState> {
...
-      />
+      <>
+        <Switch
+          id={namespace + '.' + name}
+          className='sync-toggle'
+          label={t`Sync`}
+          isChecked={this.isCollectionSynced(name, namespace)}
+          onChange={() => this.toggleCollectionSync(name, namespace)}
+        />
+        {list && this.renderMenu(list, collection)}
```
himdel, line 378: (and here)

test/cypress/integration/collection_detail.js
```diff
@@ -2,6 +2,20 @@ describe('Collection detail', () => {
...
+    cy.contains('Deprecate').click();
+    cy.contains('This collection has been deprecated.', { timeout: 10000 });
+  }
+
+  function undeprecate() {
+    cy.get('[aria-label=Actions]').click();
+    cy.contains('Undeprecate').click();
+    cy.contains('This collection has been deprecated.', {
+      timeout: 10000,
+    }).should('not.exist');
```
himdel, line 16: there should exist a different message

waiting for not exist is not useful if you want to wait until the undeprecation happened

test/cypress/integration/collection_upload.js
```diff
@@ -4,21 +4,73 @@ describe('Collection Upload Tests', () => {
...
-
     cy.createUser(userName, userPassword);
   });
 
   beforeEach(() => {
     cy.login(userName, userPassword);
   });
 
+  it('should not upload new collection version in collection list when user does not have permissions', () => {
+    cy.login(userName, userPassword);
```
himdel, line 19: the beforeEach already does this

test/cypress/integration/collection_upload.js
```diff
@@ -4,21 +4,73 @@ describe('Collection Upload Tests', () => {
...
+      '/ui/repo/published?page_size=10&view_type=card&keywords=testcollection',
+    );
+    cy.contains('testcollection');
+    cy.get('[aria-label=Actions]').click();
+    cy.contains('Upload new version').click();
+    cy.contains("You don't have rights to do this operation.");
+  });
+
+  it('should not upload new collection version in collection detail when user does not have permissions', () => {
+    cy.login(userName, userPassword);
```
himdel, line 38: (and here)

test/cypress/integration/collections_list.js
```diff
@@ -1,6 +1,46 @@
...
+  }
+
+  function undeprecate() {
+    cy.visit('/ui/repo/published/my_namespace/my_collection0');
+    cy.contains('This collection has been deprecated.');
+    cy.get('[aria-label=Actions]').click();
+    cy.contains('Undeprecate').click();
+    cy.contains('This collection has been deprecated.', {
+      timeout: 10000,
+    }).should('not.exist');
```
himdel, line 25: https://github.com/ansible/ansible-hub-ui/pull/1766/files#r871837582

src/containers/search/search.tsx
himdel: (github started hiding this convo because of the other threads.. moving this to a new comment)

himdel:
( continuing from https://github.com/ansible/ansible-hub-ui/pull/1766#discussion_r871814647 )

OK, the card situation is a bit more and less complicated than a new flexitem ... we're already in flexbox so everything is a flexitem a we just need the logo to be the one that grows, pushing the static ones to the right .. but the logo was set to center horizontally, instead of aligning left .. and we still end up with double right padding for the kebab ... I think this works...

(I started with css from master .. some of the removed css in this diff may be needed for other purposes)

```diff
diff --git a/src/components/cards/cards.scss b/src/components/cards/cards.scss
index 34ffbff2..867d000b 100644
--- a/src/components/cards/cards.scss
+++ b/src/components/cards/cards.scss
@@ -49,18 +49,15 @@
   .logo-row {
     display: flex;
     justify-content: space-between;
-    padding-left: 15px;
-    padding-right: 5px;
+
+    // CardHeader right padding 24px - Dropdown right pading 16px .. so the kebab has just single right padding, not double
+    padding-right: 8px;
 
     .icon {
       color: var(--pf-global--info-color--100);
       margin-right: 3px;
     }
 
-    .left-container {
-      display: flex;
-    }
-
     .hub-signature-badge {
       margin-left: 5px;
     }
diff --git a/src/components/cards/collection-card.tsx b/src/components/cards/collection-card.tsx
index 12e89744..93028982 100644
--- a/src/components/cards/collection-card.tsx
+++ b/src/components/cards/collection-card.tsx
@@ -2,14 +2,16 @@ import { t, Trans } from '@lingui/macro';
 import * as React from 'react';
 import cx from 'classnames';
 import {
+  Badge,
   Card,
-  CardHeader,
   CardBody,
   CardFooter,
-  TextContent,
+  CardHeader,
+  Flex,
+  FlexItem,
   Text,
+  TextContent,
   TextVariants,
-  Badge,
   Tooltip,
 } from '@patternfly/react-core';
 
@@ -55,13 +57,12 @@ export class CollectionCard extends React.Component<IProps> {
             image={namespace.avatar_url}
             size='40px'
             unlockWidth
+            flexGrow
           />
 
-          <div className='left-container'>
-            <TextContent>{this.getCertification(repo)}</TextContent>
-            <SignatureBadge isCompact signState={sign_state} />
-            {menu}
-          </div>
+          <TextContent>{this.getCertification(repo)}</TextContent>
+          <SignatureBadge isCompact signState={sign_state} />
+          {menu}
         </CardHeader>
         <CardHeader>
           <div className='name'>
diff --git a/src/components/logo/logo.tsx b/src/components/logo/logo.tsx
index 78494946..0b7ffe2f 100644
--- a/src/components/logo/logo.tsx
+++ b/src/components/logo/logo.tsx
@@ -10,6 +10,7 @@ interface IProps {
   alt: string;
   className?: string;
   unlockWidth?: boolean;
+  flexGrow?: boolean;
   fallbackToDefault?: boolean;
 }
 
@@ -31,6 +32,7 @@ export class Logo extends React.Component<IProps, IState> {
       image,
       size,
       unlockWidth,
+      flexGrow,
       width,
     } = this.props;
     const { failed } = this.state;
@@ -38,7 +40,7 @@ export class Logo extends React.Component<IProps, IState> {
     const style = {
       height: size,
       display: 'flex',
-      justifyContent: 'center',
+      justifyContent: 'flex-start',
       alignItems: 'center',
       width: width,
     };
@@ -49,6 +51,10 @@ export class Logo extends React.Component<IProps, IState> {
       style['width'] = size;
     }
 
+    if (flexGrow) {
+      style['flexGrow'] = 1;
+    }
+
     // use inline css so we can set size
     return (
       <div className={className} style={style}>
diff --git a/src/containers/search/search.scss b/src/containers/search/search.scss
index 0fe4d06b..15619e77 100644
--- a/src/containers/search/search.scss
+++ b/src/containers/search/search.scss
@@ -26,24 +26,6 @@
       flex-wrap: wrap;
       align-content: flex-start;
 
-      div[aria-label='collection-kebab'] {
-        button[aria-label='Actions'] {
-          transform: scaleY(0.7);
-          position: relative;
-          top: -10px;
-          padding-left: 8px;
-        }
-
-        ul {
-          position: absolute;
-          top: 20px;
-        }
-      }
-
-      .hub-signature-badge {
-        height: 16px;
-      }
-
       .card {
         margin-left: 24px;
         margin-bottom: 24px;
@@ -79,8 +61,4 @@
     border-top: 1px solid #d8d8d8;
     flex-shrink: 0;
   }
-
-  .list .sync-toggle {
-    margin-right: 10px;
-  }
 }
```

src/components/headers/collection-header.tsx
MilanPospisil: Yes and I suggest to create new Issue for that - I saw that this is wrong in more places in the code and many times, the parse task Id does not use utility function.

src/containers/search/search.scss
MilanPospisil: Í chose hub-list, because list-container is already div around it. I hope it will not break too much tests :-)

MilanPospisil: But we must also change .cards to hub-cards, otherwise its bad conventions. Is it really necessary to do this in this PR (its already pretty big)? I have to redo this in tests as well. This naming did not came from me PR, it was already there.

src/containers/search/search.scss
MilanPospisil: Ok. Is there some issue with that in css? In cypress we use this a lot.

src/containers/search/search.tsx
MilanPospisil: So lets rename the function to better match what it is doing.

test/cypress/integration/collection_detail.js
MilanPospisil: Why do you think its wrong? The text is not alert, but it is part of the collection detail form data. So when the form is reloaded and this message disappear, it means the collection is no longer deprecated.

test/cypress/integration/collection_upload.js
MilanPospisil: Yeah, I missed this, usualy beforeEach contains admin login :-D

src/containers/search/search.scss
himdel: well, aria-label is human readable text .. we should not have any css that depends on the current locale to work

in general, if you need to write css for an element, it should have a classname

test/cypress/integration/collection_detail.js
himdel: well, `deprecate()` clicks Deprecate and then waits for it to be done
while `undeprecate()` clicks Undeprecate and then waits for deprecate not to have happened

I think both should just wait for the right success status message

src/components/headers/collection-header.tsx
```diff
@@ -465,7 +518,7 @@ export class CollectionHeader extends React.Component<IProps, IState> {
             ) : null
           }
         >
-          {collection.deprecated && (
+          {collection['deprecated'] && (
```
himdel, line 521:
```suggestion
          {collection.deprecated && (
```

src/components/headers/collection-header.tsx
```diff
@@ -238,10 +252,49 @@ export class CollectionHeader extends React.Component<IProps, IState> {
...
+        onClick={() => this.deprecate(collection)}
+        key='deprecate'
+        isDisabled={DEPLOYMENT_MODE === Constants.INSIGHTS_DEPLOYMENT_MODE}
+        description={
+          DEPLOYMENT_MODE === Constants.INSIGHTS_DEPLOYMENT_MODE
+            ? t`Temporarily disabled due to sync issues. (AAH-1237)`
+            : null
+        }
+      >
+        {collection['deprecated'] ? t`Undeprecate` : t`Deprecate`}
```
himdel, line 265:
```suggestion
        {collection.deprecated ? t`Undeprecate` : t`Deprecate`}
```

src/containers/search/search.tsx
```diff
@@ -71,19 +82,44 @@ class Search extends React.Component<RouteComponentProps, IState> {
...
+  }
+
+  private load() {
     this.queryCollections();
 
     if (DEPLOYMENT_MODE === Constants.INSIGHTS_DEPLOYMENT_MODE) {
       this.getSynclist();
     }
+
+    this.setState({ synclist: { collections: [] } as any }); // DEBUG
```
himdel, line 103:
```suggestion
```

src/containers/search/search.tsx
himdel: Sure, but the point was not to abuse the function that renders the sync toggle with extra code, when the differences can live outside the function...

and you're actually calling `renderMenu` twice now in the card view .. once in `menu={this.renderMenu` and the other from `renderCollectionControls`.

---

```diff

diff --git a/src/containers/search/search.tsx b/src/containers/search/search.tsx
index d6204884..5b41a563 100644
--- a/src/containers/search/search.tsx
+++ b/src/containers/search/search.tsx
@@ -272,12 +272,7 @@ class Search extends React.Component<RouteComponentProps, IState> {
               className='card'
               key={c.id}
               {...c}
-              footer={this.renderCollectionControls(
-                c.name,
-                c.namespace.name,
-                false,
-                c,
-              )}
+              footer={this.renderSyncToggle(c.name, c.namespace.name)}
               repo={this.context.selectedRepo}
               menu={this.renderMenu(false, c)}
             />
@@ -359,29 +354,21 @@ class Search extends React.Component<RouteComponentProps, IState> {
     );
   }
 
-  private renderCollectionControls(
-    name: string,
-    namespace: string,
-    list: boolean,
-    collection,
-  ): React.ReactNode {
+  private renderSyncToggle(name: string, namespace: string): React.ReactNode {
     const { synclist } = this.state;
 
     if (!synclist) {
-      return list ? this.renderMenu(list, collection) : null;
+      return null;
     }
 
     return (
-      <>
-        <Switch
-          id={namespace + '.' + name}
-          className='sync-toggle'
-          label={t`Sync`}
-          isChecked={this.isCollectionSynced(name, namespace)}
-          onChange={() => this.toggleCollectionSync(name, namespace)}
-        />
-        {list && this.renderMenu(list, collection)}
-      </>
+      <Switch
+        id={namespace + '.' + name}
+        className='sync-toggle'
+        label={t`Sync`}
+        isChecked={this.isCollectionSynced(name, namespace)}
+        onChange={() => this.toggleCollectionSync(name, namespace)}
+      />
     );
   }
 
@@ -462,11 +449,11 @@ class Search extends React.Component<RouteComponentProps, IState> {
                 showNamespace={true}
                 key={c.id}
                 {...c}
-                controls={this.renderCollectionControls(
-                  c.name,
-                  c.namespace.name,
-                  true,
-                  c,
+                controls={
+                  <>
+                    {this.renderSyncToggle(c.name, c.namespace.name)}
+                    {this.renderMenu(true, c)}
+                  </>
                 )}
                 repo={this.context.selectedRepo}
               />
```

(this keeps the original `renderSyncToggle` unchanged)

test/cypress/integration/collection_detail.js
MilanPospisil: So wee need to add alerts in the first place :-)

MilanPospisil: We should wait for https://github.com/ansible/ansible-hub-ui/pull/2062/. But additionaly, I will those alerts also to other pages that now (in this PR) uses deprecation button.

himdel:
I merged the open suggestions to fix the linter issues and cause the tests to rerun.. LGTM if it goes green :)

himdel:
Merging and addressing the rbac issue in a separate PR:

PR #1767, opened 2022-03-15
Missing cypress tests: Collections > Collections > List screen

brumik:
No-Issue

Connected to missing tests in this list:
https://github.com/himdel/ansible-hub-ui/issues/1

### DIFF
--- a/test/cypress/integration/collections_list.js
+++ b/test/cypress/integration/collections_list.js
@@ -1,33 +1,18 @@
 import { range } from 'lodash';
 
 describe('Collections list Tests', () => {
-  let items = [];
-
   before(() => {
-    cy.login();
     cy.deleteNamespacesAndCollections();
 
     cy.galaxykit('namespace create my_namespace');
     // insert test data
     range(21).forEach((i) => {
-      let item = { name: 'my_collection' + i };
-      items.push(item);
       cy.galaxykit('-i collection upload my_namespace my_collection' + i);
     });
+  });
 
-    // load items. Because not all test support cleaning yet,
-    // some other collections may be present from previous test, so we must load them, we can not expect
-    // that only our test data are in database.
-    cy.intercept(
-      'GET',
-      Cypress.env('prefix') +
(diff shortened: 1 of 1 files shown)


ShaiahWren (commented):
Excellent, good work so far! :)
Could we include a test that switches to the 'community' tab? 
Also, it would be great to do a 'collection cleanup' that removes test collections once the suite has ran.

MilanPospisil (approved):

MilanPospisil:
> Excellent, good work so far! :) Could we include a test that switches to the 'community' tab? Also, it would be great to do a 'collection cleanup' that removes test collections once the suite has ran.

LGTM, but you have point. We should include this.

brumik:
Thanks, sure I will add the extra test and the cleanup. But I had the feeling we are doing the cleanup `before` the suites in a lot of cases (at least the cases I checked out), in the future should I include in all of the tests a data cleanup `after`? @ShaiahWren

test/cypress/integration/collections_list.js
```diff
@@ -1,33 +1,18 @@
...
     cy.galaxykit('namespace create my_namespace');
     // insert test data
     range(21).forEach((i) => {
-      let item = { name: 'my_collection' + i };
-      items.push(item);
       cy.galaxykit('-i collection upload my_namespace my_collection' + i);
     });
+  });
 
-    // load items. Because not all test support cleaning yet,
```
brumik, line 18: When editing the after function I seen we already cleaning up the remaining collections (line 5), so it is safe to remove the double check intercept.

brumik:
After Milan edited his comment we had a talk:
1. We should wait for deletions in galaxykit
2. If it's possible we could introduce an API endpoint for deleting synchronously multiple data like `Namespaces.objects.all().delete()` style.

This is not concerning this PR though.

ShaiahWren (approved):
LGTM
Let's keep in mind the galaxykit deletion and possible a new endpoint for a later pr.

PR #1770, opened 2022-03-15
Add github actions for cron cloud stage deployment

himdel:
* `cloud-stage-enable`: triggered manually, just adds a `.cloud-stage-cron.enabled` to master-stable
* `cloud-stage-disable`: triggered manually, just removes .cloud-stage-cron.enabled` from master-stable

* `cloud-stage-cron`: triggered periodically every Tue & Thu at 13:00 UTC, fails without `.cloud-stage-cron.enabled`, pushes master to master-stable

Cc @awcrosby , WDYT? :)
Maybe we can do more calendar integration later, but for now, just having a button to enable/disable the cron action should work, I think.

The one thing I'm not sure of is whether the cron action can just push to master-stable and still trigger the deploy-cloud action, or if we have to call the `deploy-cloud` action explicitly from `cloud-stage-cron` because of the "actions don't trigger actions" loop prevention github-side.


awcrosby:
@himdel we thought about this in more detail...

We have a requirement that each commit to master in frontend and backend are auto-deployed to stage. And since each commit to ansible-hub-ui `master` branch is currently auto-deployed to console.stage.redhat.com/beta, I think we are satisfying that requirement. I created AAH-1447 to track the possible future step of deploying to console.stage.redhat.com rather than console.stage.redhat.com/beta. I think we can close this PR.

awcrosby:
Actually thinking more, since we still need to pause deploys to allow for QE test, this may be a good approach for now.

.github/workflows/cloud-stage-disable.yml
```diff
@@ -0,0 +1,26 @@
...
+      with:
+        ref: 'master'
+
+    - name: "Disable, push to master"
+      run: |
+        git config --global user.name 'GH Actions'
+        git config --global user.email 'gh_actions@users.noreply.github.com'
+
+        rm .cloud-stage-cron.enabled
+        git add .cloud-stage-cron.enabled
```
awcrosby, line 22: at the end of both "Disable periodic stage deploy" and "Enable periodic stage deploy" actions, there will be a `.cloud-stage-cron.enabled` file?

himdel: in this case we're adding the removal :)

himdel:
```
himdel@aule:~/ansible-hub-ui$ git status -sb
## test
himdel@aule:~/ansible-hub-ui$ ls -l .cloud-stage-cron.enabled 
-rw-r--r-- 1 himdel himdel 0 Mar 21 21:37 .cloud-stage-cron.enabled
himdel@aule:~/ansible-hub-ui$ rm .cloud-stage-cron.enabled
himdel@aule:~/ansible-hub-ui$ git add .cloud-stage-cron.enabled
himdel@aule:~/ansible-hub-ui$ git status -sb
## test
D  .cloud-stage-cron.enabled
himdel@aule:~/ansible-hub-ui$ ls -l .cloud-stage-cron.enabled 
ls: cannot access '.cloud-stage-cron.enabled': No such file or directory
himdel@aule:~/ansible-hub-ui$ git commit
...
# On branch test
# Changes to be committed:
#_deleted:    .cloud-stage-cron.enabled
```

I think it works .. but would it help to replace the `rm + git add` with `git rm -f`?

awcrosby: cool, lgtm

.github/workflows/cloud-stage-cron.yml
```diff
@@ -0,0 +1,25 @@
...
+      with:
+        ref: 'master'
+
+    - name: "Fail if paused"
+      run: |
+        [ -f .cloud-stage-cron.enabled ] || exit 1
+
+    - name: "Push to master-stable"
+      run: |
+        git remote set-url origin https://x-access-token:${{ secrets.GITHUB_TOKEN }}@github.com/ansible/ansible-hub-ui.git
```
himdel, line 24: https://github.com/ansible/ansible-hub-ui/runs/5645368756?check_suite_focus=true#step:4:7

fails to push, probably because we have to fetch master-stable first, will fix by Thu

PR #1775, opened 2022-03-16
use app context to display delete alerts

ShaiahWren:
Issue: AAH-1383

Assignment: Make visible the delete alerts on ex env detail page using global context alerts.

Though the console logs display `this.context.alerts` as intended, alerts _still_ do not appear on the UI. @newswangerd would you like to take a look?

### DIFF
--- a/CHANGES/1383.task
+++ b/CHANGES/1383.task
@@ -1,0 +1,1 @@
+Make visible the delete alerts on ex env detail page using global context alerts.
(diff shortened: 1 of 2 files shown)


brumik:
1. I believe the Jira task is turned around. There SHOULD be an alert when deleting right?
2. You need to push all the alerts to the context. This way when you are unmounting some components it will still remain there. Currently when unmounting the component you reset the alers context anyway so I think it is safe. Unfortunately I cannot upload diff files here, so I'll pm you a quick solution which works (but you need to double check it).
3. When all alerts are in the context you don't need local variables to duplicate alert data as well as you don't need to do willMount method copy of it.

ShaiahWren:
@brumik per 1), yes, the jira ticket is turned around, @newswangerd is the creator :)

himdel:
(replaced by https://github.com/ansible/ansible-hub-ui/pull/1810)

PR #1777, opened 2022-03-17
Missing cypress tests: Namespaces > Details

brumik:
No-Issue

Checking off few more boxes from this list: https://github.com/himdel/ansible-hub-ui/issues/1


test/cypress/integration/namespace_detail.js
```diff
@@ -0,0 +1,43 @@
...
+    cy.visit('/ui/repo/published/namespace_detail_test');
+  });
+
+  it('should display the collections belonging to the namespace', () => {
+    cy.get('[data-cy="CollectionListItem"]').should('have.length', 2);
+    cy.get('[data-cy="CollectionListItem"]').contains('collection1');
+    cy.get('[data-cy="CollectionListItem"]').contains('collection2');
+  });
+
+  it('should the deprecation button apply the label (after page reload)', () => {
```
ShaiahWren, line 25:
```suggestion
  it('should show deprecation label after button click and page reload', () => {
```

test/cypress/integration/namespace_detail.js
```diff
@@ -0,0 +1,43 @@
...
+    cy.get('[data-cy="CollectionListItem"]:first').contains('DEPRECATED');
+  });
+
+  it('should show the correct URL when clicking on the CLI configuration tab', () => {
+    cy.get('.pf-c-tabs__link').eq(1).click();
+    cy.get('[aria-label="Copyable input"]')
+      .invoke('val')
+      .should('contain', '/inbound-namespace_detail_test/');
+  });
+});
```
ShaiahWren, line 43: This test suite looks good :) It looks like we're just missing a test that checks the form and filename upon uploading a new version. Do you have the capacity to implement this? Lmk if you need any input. :)

ShaiahWren (commented):
@himdel what do you think of renaming namespaces.js => `namespaces_list.js` ?

test/cypress/integration/namespace_detail.js
brumik: I was trying, but I am not sure how to store in the tests the collections that I would like to upload. We would need at least 3:
1. one good collection (demo_v1)
2. one bad collection name (not sure here)
3. one good new version (demo_v2)

I didn't see any file handling in our tests yet, I think we could store it somewhere in fixtures maybe? (That one I am not sure is suited for these kind of files.)

Also we probably need to add the `npm install --save-dev cypress-file-upload` dependency.

So yeah, I could do it but need some input and did not wanted to lean on other people writing these tests.

ZitaNemeckova: @brumik I did it in https://github.com/ansible/ansible-hub-ui/pull/1084 .

brumik: Thank you, I missed this. (Forgot that cy uses different package.json)

I will in that case maybe move it and extend it with a incorrect filename and error chechking.

brumik:
@ShaiahWren Did you tried to run the tests with empty db locally? (`make docker/resetdb && make docker/loaddata). For me they are passing just fine locally. 

Otherwise I updated with all the requested changes for the upload button.

MilanPospisil:
> @MilanPospisil I encountered a really weird error. If I am running the tests multiple times, the deprecation is not resetting, even though between two runs the collections is definitely deleted. Have you ever seen this kind of behavior? I think it won't make big problems neither on local nor CI testing but it is bugging me.

Sorry, Im replying late (I was probably ill at the time and missed the comment). Deprecation lasts even if the collection is deleted, when you insert the same collection again, it stays deprecated. I dont know if that is the right behavior and yes, it can mess the tests :-( I try to undeprecte any collection I deprecate in the test, but if that fails, other tests may be broken if they insert the collection with the same name. I suggest we try to deprecate only collections with quite uniqe names not used in others tests.

himdel:
Any problem with deprecation staying after collection delete should be resolved as of https://github.com/pulp/pulp_ansible/pull/888, pulp_ansible 0.13.0+, https://github.com/ansible/galaxy_ng/pull/1204 :)

himdel (approved):

PR #1762, opened 2022-03-14
Role List Table component

jerabekjiri:
No-Issue

Adding prototype component for basic Role List Table component that is identical on new/changed `RBAC` pages.

I decided to use a functional component and use hooks (`useState`). I've seen it's used for `collection signing` (@himdel is that okay? I can revert it to class component).

Use the component in code:
```
<RoleListTable
    roles={roles}
    dropdownActions={[
        <DropdownItem
            key='edit-role'
            onClick={(role) => {
                // edit role action
            }}
        >
            {t`Edit Role`}
        </DropdownItem>,
        <DropdownItem
            key='duplicate-role'
            onClick={(role) => {
                // duplicate role action
            }}
        >
            {t`Duplicate role`}
        </DropdownItem>,
        <DropdownItem
            key='delete-role'
            onClick={(role) => {
            // delete role action
            }}
        >
          {t`Delete role`}
        </DropdownItem>,
    ]}
    params={params}
    updateParams={(p) => {
        this.updateParams(p, () => this.queryGroup());
    }}
    isUserManagementDisabled={isUserMgmtDisabled}
/>
 ```
### Expanded table 
Screenshot from 2022-03-14 15-16-02
### Dropdown actions
Screenshot from 2022-03-14 15-16-14
### Dropdown actions 2
Screenshot from 2022-03-14 15-19-08

 Would this work for your `Role Page`? @ShaiahWren

PR #1797, opened 2022-03-21
Rework groups view

jerabekjiri:
Issue: [AAH-1130](https://issues.redhat.com/browse/AAH-1130)

This PR adds group roles management and the process of adding roles to the group.

Using `RoleListTable` component from [#1762](https://github.com/ansible/ansible-hub-ui/pull/1762).

### Access tab with list of roles
Screenshot from 2022-03-21 12-26-26

### Delete role modal
Screenshot from 2022-03-20 01-35-46

### Empty list of roles
Screenshot from 2022-03-19 22-49-53

### Wizard - Select role(s) step
Screenshot from 2022-03-21 12-26-59

### Wizard - Preview roles
Screenshot from 2022-03-21 12-27-08

### DIFF
--- a/src/api/generic-pulp.ts
+++ b/src/api/generic-pulp.ts
@@ -6,15 +6,6 @@ export class API extends PulpAPI {
   get(id: string, apiPath?: string) {
     return this.http.get(this.getPath(apiPath) + id);
   }
-
-  list(params) {
-    const changedParams = { ...params };
-    if (changedParams['sort']) {
-      changedParams['ordering'] = changedParams['sort'];
-      delete changedParams['sort'];
-    }
-    return super.list(changedParams);
-  }
 }
 
 export const GenericPulpAPI = new API();
--- a/src/api/generic-pulp.ts
+++ b/src/api/generic-pulp.ts
@@ -6,6 +6,15 @@ export class API extends PulpAPI {
   get(id: string, apiPath?: string) {
     return this.http.get(this.getPath(apiPath) + id);
   }
+
+  list(params) {
+    const changedParams = { ...params };
+    if (changedParams['sort']) {
+      changedParams['ordering'] = changedParams['sort'];
+      delete changedParams['sort'];
+    }
+    return super.list(changedParams);
+  }
 }
 
 export const GenericPulpAPI = new API();
--- a/src/api/group-role.ts
+++ b/src/api/group-role.ts
@@ -1,0 +1,57 @@
+import { PulpAPI } from './pulp';
+import { RoleAPI } from './role';
+class API extends PulpAPI {
+  apiPath = 'groups/';
+
+  constructor() {
+    super();
+  }
+
+  // TODO: filter out the role that are already assigned to the group (cant be reassigned)
+  // FIXME: not_contains in filter params
+  // exluce 'core.' results from results?
+  // if done this in UI, page, page_size wont work
+
+  async getRolesWithPermissions(id, params?) {
+    const assignedRoles = await this.list(
+      params,
+      this.apiPath + `${id}/roles/`,
+    );
+
+    // allow limit 1000
+    const allRoles = await RoleAPI.list({ page_size: 1000 });
+
+    return new Promise((resolve, reject) => {
+      Promise.all([assignedRoles, allRoles])
+        .then(([assigned, all]) => {
+          // match roles with assigned roles
+          const data = assigned.data.results.map(({ role, pulp_href }) => ({
+            ...all['data'].results.find(({ name }) => name === role),
+
+            // swap pulp_href role with assigned pulp_href role
+            // to delete the assigned role
+            pulp_href,
+          }));
+
+          resolve({
+            data,
+            count: data.length,
+          });
+        })
+        .catch((e) => reject(e));
+    });
+  }
+
+  removeRole(id, pulpId) {
+    return this.http.delete(this.apiPath + `${id}/roles/${pulpId}`);
+  }
+
+  addRoleToGroup(id, role, content_object = null) {
+    return this.http.post(this.apiPath + `${id}/roles/`, {
+      role: role.name,
+      content_object,
+    });
+  }
+}
+
+export const GroupRoleAPI = new API();
--- a/src/api/group-role.ts
+++ b/src/api/group-role.ts
@@ -46,7 +46,7 @@ class API extends PulpAPI {
   }
 
   removeRole(id, pulpId) {
-    return this.http.delete(this.apiPath + `${id}/roles/${pulpId}`);
+    return this.http.delete(this.apiPath + `${id}/roles/${pulpId}/`);
   }
 
   addRoleToGroup(id, role, content_object = null) {
--- a/src/api/group-role.ts
+++ b/src/api/group-role.ts
@@ -2,55 +2,44 @@ import { PulpAPI } from './pulp';
 import { RoleAPI } from './role';
 
 class API extends PulpAPI {
-  apiPath = 'groups/';
-
   constructor() {
-    super();
+    super('groups/');
   }
 
-  async getRolesWithPermissions(id, params?) {
-    const assignedRoles = await this.list(
-      params,
-      this.apiPath + `${id}/roles/`,
-    );
+  getRolesWithPermissions(id, params?) {
+    const assignedRoles = this.list(params, `${id}/roles/`);
 
-    const allRoles = await RoleAPI.list();
+    const allRoles = RoleAPI.list();
 
-    return new Promise((resolve, reject) => {
-      Promise.all([assignedRoles, allRoles])
-        .then(([assigned, all]) => {
-          // match roles with assigned roles
-          const data = assigned.data.results
-            .map(({ role, pulp_href }) => {
-              const data = all['data'].results.find(
-                ({ name }) => name === role,
-              );
-              if (data) {
-                return {
-                  ...data,
-                  // swap pulp_href role with assigned pulp_href role
-                  // to delete the assigned role
-                  pulp_href,
-                };
-              }
-            })
-            .filter(Boolean);
-
-          resolve({
-            data,
-            count: data.length,
-          });
+    return Promise.all([assignedRoles, allRoles]).then(([assigned, all]) => {
+      // match roles with assigned roles
+      const results = assigned.data.results
+        .map(({ role, pulp_href }) => {
+          const data = all['data'].results.find(({ name }) => name === role);
+          if (data) {
+            return {
+              ...data,
+              // swap pulp_href role with assigned pulp_href role
+              // to delete the assigned role
+              pulp_href,
+            };
+          }
         })
-        .catch((e) => reject(e));
+        .filter(Boolean);
+
+      return {
+        data: results,
+        count: results.length,
+      };
     });
   }
 
   removeRole(id, pulpId) {
-    return this.http.delete(this.apiPath + `${id}/roles/${pulpId}/`);
+    return this.http.delete(`${id}/roles/${pulpId}/`);
   }
 
   addRoleToGroup(id, role, content_object = null) {
-    return this.http.post(this.apiPath + `${id}/roles/`, {
+    return this.http.post(`${id}/roles/`, {
       role: role.name,
       content_object,
     });
--- a/src/api/group-role.ts
+++ b/src/api/group-role.ts
@@ -1,5 +1,6 @@
 import { PulpAPI } from './pulp';
 import { RoleAPI } from './role';
+
 class API extends PulpAPI {
   apiPath = 'groups/';
 
@@ -7,31 +8,33 @@ class API extends PulpAPI {
     super();
   }
 
-  // TODO: filter out the role that are already assigned to the group (cant be reassigned)
-  // FIXME: not_contains in filter params
-  // exluce 'core.' results from results?
-  // if done this in UI, page, page_size wont work
-
   async getRolesWithPermissions(id, params?) {
     const assignedRoles = await this.list(
       params,
       this.apiPath + `${id}/roles/`,
     );
 
-    // allow limit 1000
-    const allRoles = await RoleAPI.list({ page_size: 1000 });
+    const allRoles = await RoleAPI.list();
 
     return new Promise((resolve, reject) => {
       Promise.all([assignedRoles, allRoles])
         .then(([assigned, all]) => {
           // match roles with assigned roles
-          const data = assigned.data.results.map(({ role, pulp_href }) => ({
-            ...all['data'].results.find(({ name }) => name === role),
-
-            // swap pulp_href role with assigned pulp_href role
-            // to delete the assigned role
-            pulp_href,
-          }));
+          const data = assigned.data.results
+            .map(({ role, pulp_href }) => {
+              const data = all['data'].results.find(
+                ({ name }) => name === role,
+              );
+              if (data) {
+                return {
+                  ...data,
+                  // swap pulp_href role with assigned pulp_href role
+                  // to delete the assigned role
+                  pulp_href,
+                };
+              }
+            })
+            .filter(Boolean);
 
           resolve({
             data,
--- a/src/api/pulp.ts
+++ b/src/api/pulp.ts
@@ -4,4 +4,13 @@ export class PulpAPI extends BaseAPI {
   constructor() {
     super(API_HOST + PULP_API_BASE_PATH);
   }
+
+  list(params?, apiPath?) {
+    const changedParams = { ...params };
+    if (changedParams['sort']) {
+      changedParams['ordering'] = changedParams['sort'];
+      delete changedParams['sort'];
+    }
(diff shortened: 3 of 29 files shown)


jerabekjiri:
I think PR is ready for review.

Adding additional screen changes:

- disabling roles that are already assigned to the group (for now)
  Screenshot from 2022-04-06 01-53-37

  Screenshot from 2022-04-06 04-06-06

- updated preview roles with bold category permissions 
  Screenshot from 2022-04-06 04-06-22

- Wizard empty state (group has no assignable roles)
Screenshot from 2022-04-06 01-09-48

jerabekjiri:
Reworked the `group_management` test, but for the other tests we would need to create a custom role with permissions and that's not possible yet. `galaxykit` doesn't support that as well. So I think skipping the tests and adding them later when the feature for creating roles is merged?

himdel:
Agreed, `it.skip` should work for now.

(Once we can create roles via the UI, we should unskip the tests even if we have to do the setup via UI code (as we'll want to have tests for that anyway) - but that's probably for a second PR, as role creation UI is not merged yet either)

src/api/generic-pulp.ts
```diff
@@ -1,9 +1,19 @@
 import { PulpAPI } from './pulp';
 
 export class API extends PulpAPI {
-  apiPath = '';
+  apiPath = 'roles/';
```
himdel, line 4: this will break task detail screen

if you need to list roles, create a new api file?

src/containers/group-management/group-detail-role-management/group-detail-role-management.scss
```diff
@@ -0,0 +1,52 @@
+.empty-state-box {
```
himdel, line 1: any new css class being introduced should start with `hub-` (https://github.com/ansible/ansible-hub-ui/pull/1657)

src/api/group-role.ts
```diff
@@ -0,0 +1,60 @@
...
+            .filter(Boolean);
+
+          resolve({
+            data,
+            count: data.length,
+          });
+        })
+        .catch((e) => reject(e));
+    });
+  }
```
himdel, line 46: this is a surprising combination of async and promise, and is waiting in places it probably shouldn't

```suggestion
  getRolesWithPermissions(id, params?) {
    const assignedRoles = this.list(
      params,
      this.apiPath + `${id}/roles/`,
    );

    const allRoles = RoleAPI.list();

    return Promise.all([assignedRoles, allRoles])
      .then(([assigned, all]) => {
        // match roles with assigned roles
        const data = assigned.data.results
          .map(({ role, pulp_href }) => {
            const data = all['data'].results.find(
              ({ name }) => name === role,
            );
            if (data) {
              return {
                ...data,
                // swap pulp_href role with assigned pulp_href role
                // to delete the assigned role
                pulp_href,
              };
            }
          })
          .filter(Boolean);

        return {
          data,
          count: data.length,
        };
      })
  }
```

src/api/group-role.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  removeRole(id, pulpId) {
+    return this.http.delete(this.apiPath + `${id}/roles/${pulpId}/`);
```
himdel, line 49: looks like 3 calls in this file need to override the path ... sounds like you need a second `RoleAPI` with the right apiPath so that you don't have to do this

test/cypress.env.json.template
```diff
@@ -1,5 +1,6 @@
 {
   "prefix": "/api/automation-hub/",
+  "pulpPrefix":"/pulp/api/v3/",
```
himdel, line 3: :+1: related to #1857

src/api/group-role.ts
```diff
@@ -0,0 +1,60 @@
+import { PulpAPI } from './pulp';
+import { RoleAPI } from './role';
```
newswangerd, line 2: Duplicate imports with the same name.

newswangerd: There really should just be one RoleAPI. Defining the same thing in two different files is confusing.

src/containers/group-management/group-detail-role-management/group-detail-role-management.tsx
```diff
@@ -0,0 +1,418 @@
...
+  const deleteModal = (
+    <DeleteModal
+      title={t`Delete role: ${selectedDeleteRole?.name}`}
+      cancelAction={() => setSelectedDeleteRole(null)}
+      deleteAction={deleteRole}
+      spinner={isRoleDeleting}
+      isDisabled={isRoleDeleting}
+    >
+      <Trans>
+        You are about to remove <strong>{selectedDeleteRole?.name}</strong>{' '}
```
newswangerd, line 150: "You are about to remove {role} from {group}".

src/containers/group-management/group-detail-role-management/group-detail-role-management.tsx
```diff
@@ -0,0 +1,418 @@
...
+        under <strong>{group?.name}</strong>.
+      </Trans>
+      <br />
+      <Trans>
+        This will also remove all associated permissions under this role.
```
newswangerd, line 155: "This will revoke all permissions associated with this role from the group."

newswangerd (changes requested):

src/api/group-role.ts
jerabekjiri: > There really should just be one RoleAPI. Defining the same thing in two different files is confusing.

API path for just roles is `/pulp/api/v3/roles`, but the path for managing assigned roles to the group is `pulp/api/v3/groups/roles`. My solution works the same based on how our API interfaces are built and composed on UI, but I can merge the group-role.ts and role.ts files, but that will mean setting a different `apiPath` for every method.

src/api/pulp.ts
```diff
@@ -4,4 +4,13 @@ export class PulpAPI extends BaseAPI {
   constructor() {
     super('/pulp/api/v3/');
   }
+
+  list(params?, apiPath?) {
```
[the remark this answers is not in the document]
jerabekjiri: yes, that's what I was planning on initially. Updated and removed from `RoleAPI`, `GenericPulpAPI` and `TaskManagementAPI`.

src/api/group-role.ts
jerabekjiri: So maybe this would be possible solution by extending the pulp path like this? I know this introduces a new way of setting path because we already have `apiPath` for this?
```
export class PulpAPI extends BaseAPI {
  constructor(extendPath?: string) {
    if (extendPath) {
      super('/pulp/api/v3/' + extendPath);
    } else {
      super('/pulp/api/v3/');
    }
  }
```
Or do we really want another file just for setting path? Or we could set the `apiPath` in this file (`group-role.ts`) and in container files set the additional path and compose role params there? but that's no good as well.

newswangerd (approved):

src/containers/group-management/group-detail-role-management/preview-roles.tsx
```diff
@@ -0,0 +1,64 @@
...
+  );
+};
+
+const PreviewRoles = ({ group, selectedRoles }: Props) =>
+  Object.keys(selectedRoles).length <= 0 ? (
+    <div>No roles selected</div>
+  ) : (
+    <div className='hub-custom-wizard-layout'>
+      <p>
+        The following roles will be applied to group:{' '}
```
himdel, line 33: Trans

src/containers/group-management/group-detail-role-management/preview-roles.tsx
```diff
@@ -0,0 +1,64 @@
...
+  return (
+    <>
+      <strong>{catTitle}:</strong>&nbsp;{permission}
+    </>
+  );
+};
+
+const PreviewRoles = ({ group, selectedRoles }: Props) =>
+  Object.keys(selectedRoles).length <= 0 ? (
+    <div>No roles selected</div>
```
himdel, line 29: i18n

jerabekjiri: I missed this, I will remove the part with `selectedRoles` validation. Shouldn't be here anyway, keep the component simple. The validation for empty state can be done elsewhere.

And in the wizard, we can't visit this screen if `selectedRoles` are empty.

src/containers/group-management/group-detail-role-management/preview-roles.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+import './group-detail-role-management.scss';
+
+interface Props {
+  group: GroupObjectPermissionType;
+  selectedRoles: RoleType[];
+}
+
+const splitByDot = (perm: string) => {
+  const [category, permission] = perm.split('.');
```
himdel, line 19:
```suggestion
  const [category, permission] = perm.split('.', 2);
```

(if the string is `foo.a.b.c`, you want permission to be `a.b.c`, not just `a`, right?)

src/components/patternfly-wrappers/wizard-modal.tsx
```diff
@@ -0,0 +1,43 @@
...
+    isOpen
+    variant={variant ?? ModalVariant.large}
+    showClose={false}
+    aria-label={title}
+    hasNoBodyWrapper
+  >
+    <PFWizard
+      hasNoBodyPadding
+      navAriaLabel={`${title} steps`}
+      mainAriaLabel={`${title} content`}
```
himdel, line 34:
```suggestion
      navAriaLabel={t`${title} steps`}
      mainAriaLabel={t`${title} content`}
```

src/containers/group-management/group-detail-role-management/group-detail-role-management.tsx
```diff
@@ -0,0 +1,419 @@
...
+                  <ToolbarItem>
+                    <CompoundFilter
+                      inputText={inputText}
+                      onChange={(p) => setInputText(p)}
+                      updateParams={updateParams}
+                      params={params}
+                      filterConfig={[
+                        {
+                          id: 'role__icontains',
+                          title: t`name`,
```
himdel, line 292:
```suggestion
                          title: t`Name`,
```

src/containers/group-management/group-detail-role-management/select-roles.tsx
```diff
@@ -0,0 +1,240 @@
...
+                ? 'justifyContentFlexStart'
+                : 'justifyContentSpaceBetween',
+            }}
+            direction={{ default: 'column' }}
+          >
+            {Object.keys(selectedRoles).length !== 0 && (
+              <FlexItem>
+                <Flex>
+                  <FlexItem>
+                    <strong>Selected Roles</strong>
```
himdel, line 127:
```suggestion
                    <strong><Trans>Selected Roles</Trans></strong>
```

himdel:
I didn't really find anything non-i18n related,
stable-4.5 is split already,
so feel free to merge once the above suggestions are resolved :)

src/api/group-role.ts
```diff
@@ -0,0 +1,49 @@
...
+
+class API extends PulpAPI {
+  constructor() {
+    super('groups/');
```
himdel, line 6: ... just realized .. how is this differetnt from `apiPath = 'groups/`;`. Isn't it doing exactly the same thing now, with extra code?

src/containers/group-management/group-detail-role-management/group-detail-role-management.tsx
```diff
@@ -0,0 +1,419 @@
...
+                    })
+                    .finally(() => resolve(role));
+                }),
+            );
+
+            Promise.allSettled(selectedRolesPromises).then(() => {
+              queryRolesWithPermissions();
+              setShowAddRolesModal(false);
+              setSelectedRoles([]);
+            });
```
himdel, line 264: this should be identical to...

```suggestion
            const selectedRolesPromises = selectedRoles.map(
              (role) =>
                GroupRoleAPI.addRoleToGroup(group.id, role)
                  .then(() => {
                    addAlert(
                      t`Role ${role.name} has been successfully added to ${group.name}.`,
                      'success',
                    );
                  })
                  .catch((e) => {
                    const { status, statusText, data } = e.response;

                    const errMessage =
                      data?.non_field_errors?.length > 0
                        ? data.non_field_errors[0]
                        : errorMessage(status, statusText);

                    addAlert(
                      t`Role ${role.name} could not be assigned to group ${group.name}.`,
                      'danger',
                      errMessage,
                    );
                  })
                  .finally(() => role);
            );

            Promise.all(selectedRolesPromises).then(() => {
              queryRolesWithPermissions();
              setShowAddRolesModal(false);
              setSelectedRoles([]);
            });
```

(but either works)

jerabekjiri: thanks, makes sense, with `finally`, no need for `allSettled`

src/api/group-role.ts
jerabekjiri: Well, this goes back to our previous conversation https://github.com/ansible/ansible-hub-ui/pull/1797#discussion_r845133535 , where `apiPath` had to be overwritten and then passed in every method call. With `super` it gets set automatically.

himdel: This works, we can unify that separately :+1:

himdel:
The select roles screen does double scrollbars now...



Looks like this fixes it... (+ a few more cleanups)

```diff
diff --git a/src/components/patternfly-wrappers/pagination.tsx b/src/components/patternfly-wrappers/pagination.tsx
index c1814a65..a6df7bf5 100644
--- a/src/components/patternfly-wrappers/pagination.tsx
+++ b/src/components/patternfly-wrappers/pagination.tsx
@@ -31,8 +31,6 @@ interface IProps {
 
   /** Options for the number of items that can be displayed per page */
   perPageOptions?: number[];
-
-  style?: object;
 }
 
 export class Pagination extends React.Component<IProps> {
@@ -44,7 +42,6 @@ export class Pagination extends React.Component<IProps> {
       isTop,
       perPageOptions,
       isCompact,
-      style,
     } = this.props;
 
     const extraProps = {};
@@ -75,7 +72,6 @@ export class Pagination extends React.Component<IProps> {
           items: null,
         }}
         toggleTemplate={(props) => <ToggleTemplate ofWord={t`of`} {...props} />}
-        style={style}
       />
     );
   }
diff --git a/src/containers/group-management/group-detail-role-management/group-detail-role-management.scss b/src/containers/group-management/group-detail-role-management/group-detail-role-management.scss
index d5602f2a..e19ddd69 100644
--- a/src/containers/group-management/group-detail-role-management/group-detail-role-management.scss
+++ b/src/containers/group-management/group-detail-role-management/group-detail-role-management.scss
@@ -9,10 +9,13 @@
   padding: 1rem 1.5rem 0;
 
   .hub-select-roles-content {
-    height: 530px;
     overflow: auto;
   }
 
+  .pf-c-pagination.pf-m-bottom {
+    padding-top: 0;
+  }
+
   .hub-permission {
     margin: 3px;
   }
diff --git a/src/containers/group-management/group-detail-role-management/select-roles.tsx b/src/containers/group-management/group-detail-role-management/select-roles.tsx
index 91226d4d..b69b9dd5 100644
--- a/src/containers/group-management/group-detail-role-management/select-roles.tsx
+++ b/src/containers/group-management/group-detail-role-management/select-roles.tsx
@@ -231,7 +231,6 @@ const SelectRoles: React.FC<SelectRolesProps> = ({
               params={localParams}
               updateParams={(p) => setLocalParams(p)}
               count={rolesItemCount}
-              style={{ padding: '0px' }}
             />
           </FlexItem>
         )}
@@ -239,4 +238,5 @@ const SelectRoles: React.FC<SelectRolesProps> = ({
     </div>
   );
 };
+
 export default SelectRoles;
diff --git a/src/containers/group-management/group-list.tsx b/src/containers/group-management/group-list.tsx
index f4651c96..38b51e31 100644
--- a/src/containers/group-management/group-list.tsx
+++ b/src/containers/group-management/group-list.tsx
@@ -407,13 +407,9 @@ class GroupList extends React.Component<RouteComponentProps, IState> {
           onClick={() => {
             this.setState({
               selectedGroup: { ...group },
-              redirect: formatPath(
-                Paths.groupDetail,
-                {
-                  group: group.id,
-                },
-                { isEditing: true },
-              ),
+              redirect: formatPath(Paths.groupDetail, {
+                group: group.id,
+              }),
             });
           }}
         >
```

himdel:
The list view in Access tab - expanded roles always show no permissions, despite the data...

Unknown permissions should still be visible when showing what the group can do,
while the "No permission" lines should be hidden in non-edit view.



looks like this now only works for our permissions, not core, etc.

himdel:
Restarting tests after https://github.com/ansible/ansible-hub-ui/pull/1974

jerabekjiri:
@himdel What about this? Hide `No permission` and added `Custom permissions` (or maybe call it `Other permissions`?) to display custom permissions? 
Screenshot from 2022-04-26 16-18-25

himdel (approved):
Looks ready as well :+1:

(just a rebase.. :))

PR #1798, opened 2022-03-22
Insights token page - user.username -> cloud-services in the curl command

himdel:
Fixes AAH-1376

looks like using the actual username for `client_id` in the "renew token" curl command now triggers a 400, while using `cloud-services` works

TODO: figure out why it should be so, document the actual error in the 400 response


himdel:
Merging before cloud stage freeze, TODO above still remains.

PR #1800, opened 2022-03-22
Context alerts example

brumik:
Minimal working example of context alerts in 2 components for execution environment delete persist.

connects to #1775 

@ShaiahWren When we worked together I thought the base is the base of the list view too. This was not the case, after I looked it with fresh eyes, I realized the `list view` is the one that should display the error and that one was still using the local state to display alerts. I modified that list too to use just the context for alerts.

This posed a problem that the mixin for closing the alerts is strangely tied to the `state` (I am not sure if that is a good practice or not). I added a default function which operates on the context. I think that could be baked into that component if we would like to use it exclusively in context (and then we could move the component up to the root). But I leave small details like this up to you.

P.S.: Opening a new PR was easier than sending a message and a diff file :D

### DIFF
--- a/CHANGES/1383.task
+++ b/CHANGES/1383.task
@@ -1,0 +1,1 @@
+Make visible the delete alerts on ex env detail page using global context alerts.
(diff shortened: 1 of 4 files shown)


ShaiahWren (commented):
Good thinking, yes, the list view should also use the context alerts on the redirect. using your new function for closing the alerts in context sounds like a good idea. is it the alert-list.tsx file that you were thinking of moving? walk me through your thoughts and we'll go from there. :)

PR #1802, opened 2022-03-23
cloud-stage-cron: fetch to ensure master-stable and master have all the commits

himdel:
Follow-up to #1770 

initial checkout just goes to depth: 1,
we need more commits to have both master and master-stable (and all the commits in between), so fetching all of them now.

Should fix https://github.com/ansible/ansible-hub-ui/runs/5645368756?check_suite_focus=true#step:4:7 .


himdel:
(merging to test today's deployment)

PR #1805, opened 2022-03-24
Fixed the redirect in insights mode when deleting namespaces

brumik:
Issue: AAH-1461

**Before on insights**
Screenshot from 2022-03-18 09-26-39
Screenshot from 2022-03-18 09-27-09

**After on insights**
Screenshot from 2022-03-24 11-05-51
Screenshot from 2022-03-24 11-06-06
Screenshot from 2022-03-24 11-06-13

**After on standalone**
Screenshot from 2022-03-24 11-18-15
Screenshot from 2022-03-24 11-18-27
Screenshot from 2022-03-24 11-18-34

### DIFF
--- a/CHANGES/1461.bug
+++ b/CHANGES/1461.bug
@@ -1,0 +1,1 @@
+Fixed the broken redirect on insights mode when deleting a namespace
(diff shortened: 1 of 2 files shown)


himdel:
Good catch .. but we should not be adding ad-hoc logic to `formatPaths` just to fix 1 constant.

Replacing with https://github.com/ansible/ansible-hub-ui/pull/1807 :)

PR #1806, opened 2022-03-24
cloud-stage-cron - add debugging info

himdel:
Follow up to #1802 and #1770 

still doesn't work..  https://github.com/ansible/ansible-hub-ui/runs/5676767046?check_suite_focus=true#step:4:66

adding debugging info


himdel:
Merging before cloud stage freeze

PR #1807, opened 2022-03-24
Fixed insights mode redirect when deleting a namespace

himdel:
Issue: AAH-1461

we never use `Paths.namespaces` because that only works in standalone,
`namespaceBreadcrumb.url` is always valid in both, using that one.

This replaces #1805 - same bug, different fix.


himdel:
Merging to catch the cloud stage freeze, tested on standalone.

PR #1808, opened 2022-03-24
Cloud smoke tests revision 2

jctanner:
Thank you for a new Pull Request!

Please, make reviewers work easier by checking, if you are not introducing any technical debt:
https://github.com/ansible/ansible-hub-ui/blob/master/developer_guidelines.md

Describe the rationale and use case for this pull request.  Provide any background, examples, and images that provide further information to accurately describe what it is that you are adding to the repo.  Add subsections as necessary to organize and feel free to link and reference other PRs as necessary, but also include them in the links section below as a quick reference.

Guidelines:
* Keep Pull Request titles short and to the point, ideally under 72 characters
* Provide as much context/info in the description as necessary to get the reviewer up to the same domain knowledge level as yourself
* Keep code changes as short as possible and implementing a single feature/fix/refactoring, when possible

Links [Optional]
----------------

* http://documentation.for/library/that/I/am/adding
* [PR relevant issue or pull_request](#123)

If there are any manual steps that you would like the reviewer(s) to take to verify your changes, please describe in detail the steps to reproduce the features added by the pull request, or the bug before and after the change.


test/cypress/support/index.js
```diff
@@ -19,3 +19,14 @@ import 'cypress-file-upload';
 
 // Alternatively you can use CommonJS syntax:
 // require('./commands')
+
+// ************************************************************
+// The chrome ui in cloud is unstable and causes random
+// exceptions, so we need to ignore those wherever possible
+// *************************************************************/
+// https://github.com/quasarframework/quasar/issues/2233#issuecomment-414070235
+Cypress.on('uncaught:exception', (err, runnable) => {
```
himdel, line 28: any way to enable this only for cloud? (If it's still needed? Not sure how `quasar` relates to cloud?)

This would also mean a test won't fail unless it explicitly checks for a success notification, even when the API responds with an error.

test/package.json
```diff
@@ -19,7 +19,9 @@
   },
   "dependencies": {
     "cypress-file-upload": "^5.0.8",
+    "cypress-tags": "^0.3.0",
     "shell-escape-tag": "^2.0.2",
+    "typescript": "^4.6.2",
```
himdel, line 24: I understand cypress-tags, but I'm not seeing any typescript here? (And just adding typescript doesn't enable typescript in cypress.)

test/cypress/integration/cloud_smoke.js
```diff
@@ -0,0 +1,101 @@
...
+     * To reduce clutter in each test function, we'll
+     * navigate to the automation hub app before each 
+     * test and allow the tests to assume they are at
+     * the root level of the app
+     * **************************************************/
+
+    // wait for navbar to appear
+    cy.get('#nav-toggle', { timeout: 10000 }).should('be.visible');
+    // wait for the ansible dashboard button to appear and then click on it
+    cy.get('[data-quickstart-id="ansible_ansible-dashboard"]', { timeout: 1000 }).click();
```
himdel, line 23: I wonder if we need all these explicit timeouts... AFAIK the default is `{ timeout:5000 }`, why would we want to shorten it?

test/cypress/support/index.js
jctanner: The stage environment has `ResizeObserver loop limit exceeded` errors which force test failures unless ignored. I had similar issues when trying to write tests for analytics. The chrome stack is a noisy system.

test/package.json
jctanner: https://www.npmjs.com/package/cypress-tags

> This plugin uses TypeScript to parse the tests, so you will need typescript installed in your project.

test/cypress/integration/cloud_smoke.js
jctanner: The 1000 was a typo, it should have been 10K. UI response in the cloud over vpn is terribly slow and cypress tests fail intermittently unless the timeout is increased.

test/cypress/support/index.js
himdel: Right, that's why I'm asking to enable this only for cloud, not standalone.

PR #1809, opened 2022-03-24
cloud-stage-cron: clone manually

himdel:
the previous fails to push master to master-stable because it believes they're not related

     ! [rejected]        master -> master-stable (non-fast-forward)

that's not true, works locally, so it's probably an artifact of how the checkout action checks out other branches.

This *might* work, if not, `push -f` probably will, but I'd still like to know the reason.

### DIFF
--- a/.github/workflows/cloud-stage-cron.yml
+++ b/.github/workflows/cloud-stage-cron.yml
@@ -10,20 +10,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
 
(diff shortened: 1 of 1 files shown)


himdel:
Getting this in before the cloud stage freeze

himdel:
This worked: https://github.com/ansible/ansible-hub-ui/actions/runs/2034837781 :)

But didn't trigger the follow-up action.
Will folllow up on this :)

PR #1769, opened 2022-03-15
Update Localization strings for 4.4

rooftopcellist:
Updates the stable-4.4 Localization strings

Jira Issue: https://issues.redhat.com/browse/AAP-2545

PR #1811, opened 2022-03-27
Update locale/ to match 4.4

himdel:
git checkout upstream/stable-4.4 locale/*po
    npm run gettext:extract ; npm run gettext:compile

master translations were last updated in #1187, except for periodic re-extracts
4.4 locales were last updated in #1769, newer :)

so, taking the update to master, and re-extracting (& merging) from there


himdel:
retest after #1827

ZitaNemeckova (approved):
Needs to be done again because of the conflicts. Then it's LGTM and may be merged :)

PR #1827, opened 2022-03-28
cypress.yml - fix podman save failure

himdel:
https://github.com/ansible/ansible-hub-ui/runs/5712872845?check_suite_focus=true#step:6:501

`podman save` now fails with

    Error: docker-archive doesn't support modifying existing images

when an image already exists .. adding an explicit removal
(not quite sure why it exists when `cache-hit != 'true'` but..)


himdel:
(Merging to get tests green.)

PR #1833, opened 2022-03-30
testing for remote and local tabs in repo management

ShaiahWren:
No-Issue, but I think https://issues.redhat.com/browse/AAH-626 is relevant?

Fills in a couple more tests from https://github.com/himdel/ansible-hub-ui/issues/1 in the Repo Management department. 

So far as the '4 remote repos by type', 3 so far have been implemented. `rh-certified` and `community` types are complete. The test for a remote registry has me stuck because the username and password encryption makes those fields return `undefined` when attempting to attempt the input values. The fourth, 'none' remote repo has not been yet implemented because there is no UI flow for adding one. @himdel , any suggestions? :)


newswangerd:
I don't have any experience with cypres so I'll defer to @himdel here

test/cypress/integration/repo_management.js
```diff
@@ -38,6 +40,80 @@ describe('Repo Management tests', () => {
...
+    cy.visit(localRepoUrl);
+    cy.get('[data-cy="get-token"]').contains('Get token').click();
+    cy.get('button').contains('Load token').click();
+    cy.get('[aria-label="Copyable input"]')
+      .invoke('val')
+      .then((token) => {
+        cy.get('[aria-label="Copy to clipboard"]').realClick();
+        cy.window().then((win) => {
+          win.navigator.clipboard.readText().then((copiedToken) => {
+            assert.equal(copiedToken, token);
```
himdel, line 53:
```suggestion
            expect(copiedToken).to.equal(token);
```

so we don't have to introduce another libary

test/cypress/integration/repo_management.js
```diff
@@ -38,6 +40,80 @@ describe('Repo Management tests', () => {
...
+      'table > tbody > tr:first-child > td:nth-child(6) > .pf-c-clipboard-copy > .pf-c-clipboard-copy__expandable-content > pre',
+    )
+      .invoke('text')
+      .then((CLI) => {
+        cy.get(
+          'table > tbody > tr:first-child > td:nth-child(6) > .pf-c-clipboard-copy > .pf-c-clipboard-copy__group > button[aria-label="Copy to clipboard"]',
+        ).realClick();
+        cy.window().then((win) => {
+          win.navigator.clipboard.readText().then((copiedCLI) => {
+            assert.equal(copiedCLI, CLI);
```
himdel, line 77:
```suggestion
            expect(copiedCLI).to.equal(CLI);
```

test/cypress/integration/repo_management.js
```diff
@@ -1,3 +1,5 @@
+import assert from 'assert';
+
```
himdel, line 2:
```suggestion
```

test/cypress/integration/repo_management.js
```diff
@@ -38,6 +40,80 @@ describe('Repo Management tests', () => {
...
+      .click();
+  });
+
+  it('starts remote repo sync', () => {
+    cy.visit(remoteRepoUrl);
+
+    //checks sync status === 'Running' after sync post request
+    cy.intercept(
+      'POST',
+      'http://localhost:8002/api/automation-hub/content/rh-certified/v3/sync/',
```
himdel, line 101:
```suggestion
      Cypress.env('prefix') + 'content/rh-certified/v3/sync/',
```

tsconfig.json
```diff
@@ -15,7 +15,8 @@
     "skipLibCheck": true,
     "sourceMap": true,
     "strictNullChecks": false,
-    "target": "es5"
+    "target": "es5",
+    "types": ["cypress", "cypress-real-events"]
```
himdel, line 19: we may need to create a separate tsconfig for tests, otherwise these will be valid in real code as well

unless we don't need it? the tests don't use typescript now so they're probably not affected by tsconfig

test/cypress/integration/repo_management.js
```diff
@@ -33,11 +34,209 @@ describe('Repo Management tests', () => {
...
+  });
+
+  it('expands and copies CLI config from local repo list', () => {
+    cy.visit(localRepoUrl);
+    cy.get(
+      'table > tbody > tr:first-child > td:nth-child(6) > .pf-c-clipboard-copy > .pf-c-clipboard-copy__group > button[aria-label="Show content"]',
+    ).click();
+    cy.get(
+      'table > tbody > tr:first-child > td:nth-child(6) > .pf-c-clipboard-copy > .pf-c-clipboard-copy__expandable-content > pre',
+    ).contains('http://localhost:8002/api/automation-hub/content/community/');
```
himdel, line 70:
```suggestion
    ).contains(Cypress.env('prefix') + 'content/community/');
```

package.json
```diff
@@ -32,7 +33,8 @@
         "react-router-hash-link": "^2.4.3",
         "redux": "^4.2.0",
         "redux-logger": "^3.0.6",
-        "redux-promise-middleware": "^6.1.2"
+        "redux-promise-middleware": "^6.1.2",
+        "renovate": "^32.38.0"
```
himdel, line 37: what's `renovate`? You don't seem to be using it?

test/package.json
```diff
@@ -13,7 +13,7 @@
   "author": "",
   "license": "ISC",
   "devDependencies": {
-    "cypress": "^9.6.0",
+    "cypress": "^9.5.4",
```
himdel, line 16: do we need to get back to previous version of cypress or is that a rebase artifact?

package.json
```diff
@@ -19,6 +19,7 @@
         "axios": "~0.26.1",
         "classnames": "^2.3.1",
         "csstype": "^3.0.11",
+        "cypress-real-events": "^1.7.0",
```
himdel, line 22: this probably belongs in `test/package.json`, not here

test/cypress/integration/repo_management.js
```diff
@@ -33,11 +34,194 @@ describe('Repo Management tests', () => {
     cy.get('[data-cy="SortTable-headers"]').contains('Last updated');
     cy.get('[data-cy="SortTable-headers"]').contains('Last sync');
     cy.get('[data-cy="SortTable-headers"]').contains('Sync status');
-
-    cy.contains('Configure');
```
himdel, line 37: intentional?

test/cypress/support/index.js
```diff
@@ -16,6 +16,7 @@
 // Import commands.js using ES2015 syntax:
 import './commands';
 import 'cypress-file-upload';
+import 'cypress-real-events/support';
```
himdel, line 19: (if real-events is not used anymore, we don't need it here either)

tsconfig.json
```diff
@@ -15,7 +15,8 @@
     "skipLibCheck": true,
     "sourceMap": true,
     "strictNullChecks": false,
-    "target": "es5"
+    "target": "es5",
+    "types": ["cypress-real-events"]
```
himdel, line 19: (and here)

package.json
himdel: (also update or reset `package-lock.json` please, so that there are no changes (unless I missed something and we need some of those new deps?))

test/cypress/integration/repo_management.js
ShaiahWren: Intentional, yes, the `Configure` button no longer appears. I think it's replaced by the kebab `Edit` dropdown

ShaiahWren: Intentional, yes, `Configure` button no longer appear. I think it was replaced by the kebab `Edit` dropdown

test/package.json
ShaiahWren: It's a rebase artifact:)

package.json
ShaiahWren: Should be cleaned up now.

test/cypress/integration/repo_management.js
himdel: ..wait, but how? This is not changign code and there's no new `before`, so how does this PR make Configure disappear?

The task* test failures are happening because both of those use the Configure button to create a task, until galaxykit can do it.
(You may need to change them to use Sync?)

himdel: Oh.. Configure here will still work, the first time.
Then the tests after it make it disappear, affecting the following task* tests, r < t.

So.. looks like we need an `after` to unconfigure the repo?

test/cypress/integration/task_management_detail.js
```diff
@@ -4,18 +4,17 @@ describe('Task detail', () => {
...
       'POST',
-      Cypress.env('prefix') + '/content/rh-certified/v3/sync/',
+      Cypress.env('prefix') + 'content/community/v3/sync/',
     ).as('sync');
 
     cy.intercept('GET', Cypress.env('prefix') + '_ui/v1/remotes/?*').as(
       'remotes',
     );
-
+    cy.wait(10000);
```
himdel, line 17: this shouldn't need a new wait

test/cypress/integration/task_list.js
```diff
@@ -4,12 +4,11 @@ describe('Task table contains correct headers and filter', () => {
     cy.visit('/ui/repositories?tab=remote');
 
     cy.contains('Repo Management');
-    cy.contains('Configure');
     cy.contains('Sync');
 
     cy.intercept(
       'POST',
-      Cypress.env('prefix') + '/content/rh-certified/v3/sync/',
+      Cypress.env('prefix') + '/content/community/v3/sync/',
```
himdel, line 11: extra leading `/` here? (or extra removal of the slash in the other file below, but looks like we never add slash right after prefix)

test/cypress/integration/repo_management.js
```diff
@@ -54,6 +222,7 @@ describe('Repo Management tests', () => {
     cy.get('input[id="proxy_url"]').type('https://example.org');
     cy.get('input[id="proxy_username"]').type('test');
     cy.get('input[id="proxy_password"]').type('test');
+    cy.contains('Save').click();
```
himdel, line 225: changing a skipped test?

test/cypress/integration/repo_management.js
```diff
@@ -33,11 +34,178 @@ describe('Repo Management tests', () => {
...
+    cy.get('[aria-label="Copyable input"]')
+      .invoke('val')
+      .should('have.length', 40);
+    cy.get('button[aria-label="Copy to clipboard"]').should('be.enabled');
+  });
+
+  it('expands and copies CLI config from local repo list', () => {
+    cy.visit(localRepoUrl);
+    cy.get(
+      'table > tbody > tr:first-child > td:nth-child(6) > .pf-c-clipboard-copy > .pf-c-clipboard-copy__group > button[aria-label="Show content"]',
```
himdel, line 59: these 4 selectors are too long and probably needlessly specific?

would `.pf-c-clipboard-copy button[aria-label="Show content"]'` do the same thing?

also, lines 61-63 and 64-66 are exactly the same .. no need to check twice?

test/cypress/integration/repo_management.js
```diff
@@ -33,11 +34,178 @@ describe('Repo Management tests', () => {
...
+    cy.get(
+      'table > tbody > tr:first-child > td:nth-child(6) > .pf-c-clipboard-copy > .pf-c-clipboard-copy__expandable-content > pre',
+    ).contains(Cypress.env('prefix') + 'content/community/');
+
+    cy.get(
+      'table > tbody > tr:first-child > td:nth-child(6) > .pf-c-clipboard-copy > .pf-c-clipboard-copy__group > button[aria-label="Copy to clipboard"]',
+    ).should('be.enabled');
+  });
+
+  it('edits remote registry', () => {
```
himdel, line 73: should this live in a different file?

this is testing remote registries, same form, but all the setup is done in RR tests, no need to duplicate it here

ShaiahWren: Yes, I'll move this to another file for https://issues.redhat.com/browse/AAH-618

test/cypress/integration/repo_management.js
ShaiahWren: 100%, I tightened them up.:)

test/cypress/integration/repo_management.js
ShaiahWren: Do you mean, remove this `it` altogether?

ShaiahWren:
Tests in `task_list.js` and `task_management.js` , files unrelated, are failing

test/cypress/integration/repo_management.js
himdel: yes, remove it here and move it to `remote_registry.js` .. then you can have less setup inside the test because it will already have been done in the `before` bits

himdel:
You're changing both failing tests, and they're not failing on master.

It's probably related after all?

test/cypress/integration/repo_management.js
ShaiahWren: @himdel , edit remote repo logic has been moved to `remote_registry.js`.

ShaiahWren:
@himdel I think the problem with the task management files is that after `repo_management.js` has run, the 'Configure' button no longer appears and the 'Sync' button has already been clicked - so clicking it again in task_list won't make a network request.

himdel:
Would it help if we `skip` the sync test for now? It's not really syncing from a real server so maybe we can bring it back once we can test sync too.

himdel (approved):

ShaiahWren:
pr closed.

PR #1855, opened 2022-04-04
dependabot: switch to monthly for 4.2 and 4.3

himdel:
with 4.5 soon to be cut, we'll have dependabot updating 5 branches

reducing the frequency on the oldest 2


himdel:
TODO add 4.5

PR #1856, opened 2022-04-04
src/api: remove extra constructors

himdel:
js classes already have a default constructor calling super, no need for the construct unless we want to do something else :)


ZitaNemeckova (approved):
After rebase it' good to be merged :+1:

PR #1858, opened 2022-04-04
insights - disable global filter

himdel:
A global filter was added for the whole ansible org in insights mode, but it's only for AAP - introduced as part of https://issues.redhat.com/browse/RHCLOUD-18330 & https://github.com/RedHatInsights/insights-chrome/pull/1827

Disabling, same as https://github.com/RedHatInsights/tower-analytics-frontend/pull/777 :)

--- 

TODO screenshot, ensure this doesn't just work the first time, or move to render
TODO ensure the css thing doesn't affect other non-AAH pages

---

We may not need this after all, sounds like the change will be updated to restrict the filter to aap before it goes in prod.
Keeping for now not to forget.
Also, https://issues.redhat.com/browse/AAP-3077

### DIFF
--- a/src/loaders/app.scss
+++ b/src/loaders/app.scss
@@ -19,6 +19,11 @@
     var(--pf-global--palette--black-800);
 }
 
+// hide crc global filter - enabled for the whole ansible org but irrelevant outside AAP
+#global-filter {
+  display: none;
+}
+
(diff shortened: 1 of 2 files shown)


himdel:
No longer present in stage or prod.

PR #1860, opened 2022-04-06
Role list view

ShaiahWren:
Issue: https://issues.redhat.com/browse/AAH-1131

A big pr for role list view and create role. :)
*Role list view is in compact mode
*Create-roles component should be in MVP mode, with the exception of errorHandling when a new role new is not unique.

Screen Shot 2022-04-05 at 9 49 49 PM
Screen Shot 2022-04-05 at 9 50 03 PM


ZitaNemeckova:
Screenshot 2022-04-07 at 10 20 16
Edit button is for some reason blue. This should be fixed in a followup PR. Definetely not something to block merging :)

ZitaNemeckova:
Editing has no Save or Cancel button for locked/unloocked roles.

Screenshot 2022-04-07 at 10 27 30

src/containers/role-management/role-edit.tsx
```diff
@@ -0,0 +1,572 @@
...
+    this.state = {
+      role: null,
+
+      params: {
+        id: id,
+        page: 0,
+        page_size: params['page_size'] || 10,
+        sort: params['sort'] || 'username',
+        tab: params['tab'] || 'permissions',
+        isEditing: params['isEditing'] === 'true',
```
ZitaNemeckova, line 108: It seems that only one that you get and need is `id`. There's no pagination or sorting. And `isEditing` will always be `false` because it's not part of `params`.

src/containers/role-management/role-edit.tsx
```diff
@@ -0,0 +1,572 @@
...
+        .then((result) => {
+          this.setState({
+            originalPermissions: result.data.data.map((p) => ({
+              id: p.id,
+              name: p.permission,
+            })),
+            permissions: result.data.data.map((x) => x.permission),
+          });
+        })
+        .catch((e) => {
```
ZitaNemeckova, line 160: This gives me `404` but doesn't show an alert.

src/containers/role-management/role-list.tsx
```diff
@@ -0,0 +1,480 @@
...
+      unauthorized: false,
+      selectedRole: null,
+      expandedRoleNames: [],
+      roleToEdit: null,
+      showDeleteModal: false,
+    };
+  }
+
+  componentDidMount() {
+    if (!this.context.user || !this.context.user.model_permissions.view_group) {
```
ZitaNemeckova, line 99: What about `this.context.user.is_anonymous`?

src/containers/role-management/role-create.tsx
```diff
@@ -0,0 +1,268 @@
...
+    if (isUserMgmtDisabled) {
+      Constants.USER_GROUP_MGMT_PERMISSIONS.forEach((perm) => {
+        if (perm in filteredPermissions) {
+          delete filteredPermissions[perm];
+        }
+      });
+    }
+
+    const notAuthorised =
+      !this.context.user || !this.context.user.model_permissions.add_user;
```
ZitaNemeckova, line 79: What about `this.context.user.is_anonymous` ?

src/containers/role-management/role-create.tsx
```diff
@@ -0,0 +1,268 @@
...
+    const breadcrumbs = [
+      { url: Paths.roleList, name: t`Roles` },
+      { name: t`Create new role` },
+    ];
+    const title = t`Create new role`;
+
+    return notAuthorised ? (
```
ZitaNemeckova, line 86: It's better to do
```
<React.Fragment>
        <BaseHeader
          breadcrumbs={<Breadcrumbs links={breadcrumbs}></Breadcrumbs>}
          title={title}
        ></BaseHeader>
       {
       notAuthorised ? ( <EmptyStateUnauthorized />) : (<Main>....</Main>)
       }
        <EmptyStateUnauthorized />
      </React.Fragment>
```
if in the future there's a change in `<BaseHeader>` there's only one place to change it :)

src/containers/role-management/role-edit.tsx
```diff
@@ -0,0 +1,572 @@
...
+
+    if (!role && alerts && alerts.length) {
+      return (
+        <AlertList
+          alerts={alerts}
+          closeAlert={(i) => this.closeAlert(i)}
+        ></AlertList>
+      );
+    }
+    if (unauthorised) {
```
ZitaNemeckova, line 217: This should be moved to after `<BaseHeader>`. See `user-list` for reference.

src/containers/role-management/role-edit.tsx
```diff
@@ -0,0 +1,572 @@
...
+              >
+                <TextInput
+                  id='role_name'
+                  value={this.state.name}
+                  onChange={(value) => {
+                    this.setState({ name: value });
+                  }}
+                  type='text'
+                  validated={this.toError(!this.state.roleError)}
+                  placeholder='Add a role description here'
```
ZitaNemeckova, line 269: This should be translated.

src/containers/role-management/role-edit.tsx
```diff
@@ -0,0 +1,572 @@
...
+        </Main>
+      </React.Fragment>
+    );
+  }
+
+  private renderControls() {
+    const { user } = this.context;
+    const { editPermissions } = this.state;
+
+    if (!user || !user.model_permissions.delete_group) {
```
ZitaNemeckova, line 355: What about `user.is_anonymous`?

src/containers/role-management/role-edit.tsx
```diff
@@ -0,0 +1,572 @@
...
+                </Flex>
+              ))}
+            </div>
+          </section>
+        </Main>
+      </React.Fragment>
+    );
+  }
+
+  private renderControls() {
```
ZitaNemeckova, line 351: This method is never used.

src/containers/role-management/role-edit.tsx
```diff
@@ -0,0 +1,572 @@
...
+        },
+      ],
+    });
+  }
+
+  private toError(validated: boolean) {
+    return validated ? 'default' : 'error';
+  }
+
+  private get updateParams() {
```
ZitaNemeckova, line 562: This is never used.

src/containers/role-management/role-edit.tsx
```diff
@@ -0,0 +1,572 @@
...
+    this.setState({ savingPermissions: true }); // disable Save/Cancel while waiting
+    Promise.all(promises).then(() =>
+      this.setState({
+        editPermissions: false,
+        savingPermissions: false,
+      }),
+    );
+  }
+
+  private renderPermissions() {
```
ZitaNemeckova, line 429: This is never used.

src/containers/role-management/role-edit.tsx
```diff
@@ -0,0 +1,572 @@
...
+                      );
+                    }
+                    this.setState({ permissions: Array.from(newPerms) });
+                  }}
+                />
+              </FlexItem>
+            </Flex>
+          ))}
+        </div>
+        {editPermissions && (
```
ZitaNemeckova, line 521: This needs to be added to the `render` so the form has butttons.

src/containers/role-management/role-edit.tsx
```diff
@@ -0,0 +1,572 @@
...
+          <Form>
+            <ActionGroup>
+              <Button
+                variant='primary'
+                isDisabled={savingPermissions}
+                onClick={() => console.log('Saving!')}
```
ZitaNemeckova, line 527: It should call a saving function.

src/containers/role-management/role-edit.tsx
```diff
@@ -0,0 +1,572 @@
...
+              >
+                {t`Save`}
+              </Button>
+              <Button
+                variant='secondary'
+                isDisabled={savingPermissions}
```
ZitaNemeckova, line 533: Cancel button should be always enabled.

src/containers/role-management/role-edit.tsx
```diff
@@ -0,0 +1,572 @@
...
+
+  private actionCancelPermissions() {
+    const { originalPermissions } = this.state;
+    this.setState({
+      editPermissions: false,
+      permissions: originalPermissions.map((p) => p.name),
+    });
+  }
+
+  private actionSavePermissions() {
```
ZitaNemeckova, line 380: This is never called.

src/containers/role-management/role-edit.tsx
```diff
@@ -0,0 +1,572 @@
...
+          onClick={() => this.setState({ showDeleteModal: true })}
+          variant='secondary'
+        >
+          {t`Delete`}
+        </Button>
+      </ToolbarItem>
+    );
+  }
+
+  private actionCancelPermissions() {
```
ZitaNemeckova, line 372: Clicking a Cancel button should redirect back to Role list page. It should work the same way as in Users.

src/containers/role-management/role-create.tsx
```diff
@@ -0,0 +1,268 @@
...
+      </React.Fragment>
+    );
+  }
+  private saveRole = () => {
+    const { name, permissions, description } = this.state;
+    RoleAPI.create({ name, description, permissions })
+      .then(() => this.setState({ redirect: Paths.roleList }))
+      .catch((err) => {
+        this.setState({ roleError: mapErrorMessages(err) });
+        console.log('roleError: ', this.state.roleError);
```
ZitaNemeckova, line 254: Please remove a console.log.

src/containers/role-management/role-create.tsx
```diff
@@ -0,0 +1,268 @@
...
+        </Main>
+      </React.Fragment>
+    );
+  }
+  private saveRole = () => {
+    const { name, permissions, description } = this.state;
+    RoleAPI.create({ name, description, permissions })
+      .then(() => this.setState({ redirect: Paths.roleList }))
+      .catch((err) => {
+        this.setState({ roleError: mapErrorMessages(err) });
```
ZitaNemeckova, line 253: This is not working correctly. I tried to create a role with existing name and I seen in console, that it failed but not in UI. It seems that `mapErrorMessages` is not parsing it correctly.

src/api/role.ts
```diff
@@ -0,0 +1,24 @@
...
+  updatePermissions(id, data: Object) {
+    return this.http.patch(this.apiPath + id, data);
+  }
+
+  createRole(data: Object) {
+    return this.http.post(this.apiPath, data);
+  }
+}
+
+export const RoleAPI = new API();
```
ZitaNemeckova, line 24: Don't forget to add `export` it to `src/api/index.ts` . It's kinda working but gives some ugly errors.

src/containers/role-management/role-list.tsx
```diff
@@ -0,0 +1,480 @@
...
+                    }}
+                    params={params}
+                    ignoredParams={['page_size', 'page', 'sort', 'ordering']}
+                    niceNames={{
+                      name__icontains: t`Role name`,
+                      locked: t`Locked?`,
+                    }}
+                  />
+                </div>
+                {loading ? (
```
ZitaNemeckova, line 251: This seems to be here for the second time.

src/containers/role-management/role-list.tsx
```diff
@@ -0,0 +1,480 @@
...
+                              id: 'kebab',
+                            },
+                          ],
+                        }}
+                      >
+                        {roles.map((role) => (
+                          <ExpandableRow
+                            key={role.name}
+                            expandableRowContent={
+                              <>
```
ZitaNemeckova, line 298: Maybe `<>` is not needed. I'm not sure.

ShaiahWren: Removing it caused a bug, so I'll keep it for now? :)

src/containers/role-management/role-list.tsx
```diff
@@ -0,0 +1,480 @@
...
+      'page',
+      'page_size',
+    ]);
+
+    if (!params['page_size']) {
+      params['page_size'] = 10;
+    }
+
+    if (!params['sort']) {
+      params['sort'] = '-pulp_created';
```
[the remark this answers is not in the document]
ShaiahWren: I added the pulp_created column and then kept the sorting as `-pulp-created`, just so you can see when a new role is added it shows up nicely at the top of the page. Otherwise a role by the name 'zita' gets sent to the bottom and it seems like it wasn't created :)

src/containers/role-management/role-create.tsx
ShaiahWren: Fixed:) Thanks for your help:)

src/containers/role-management/role-edit.tsx
ShaiahWren: I think this is implemented correctly now - might want to check and make sure. :)

src/containers/role-management/role-create.tsx
ShaiahWren: Here I did ` const notAuthorised = !this.context.user || this.context.user.is_anonymous;` Does that work?

src/containers/role-management/role-list.tsx
ShaiahWren: Does `const notAuthorised = !this.context.user || this.context.user.is_anonymous;` work?

src/containers/role-management/role-edit.tsx
```diff
@@ -0,0 +1,572 @@
...
+    } else {
+      RoleAPI.get(this.state.params.id)
+        .then((result) => {
+          this.setState({
+            role: result.data,
+            description: result.data.description,
+            name: result.data.name,
+          });
+        })
+        .catch((e) => {
```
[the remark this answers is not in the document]
ShaiahWren: Should be working now. :)

src/containers/role-management/role-edit.tsx
ShaiahWren: Is this still relevant? Let me know. :)

src/containers/role-management/role-create.tsx
```diff
@@ -0,0 +1,294 @@
...
+              <div>
+                <div style={{ paddingBottom: '8px', paddingTop: '16px' }}>
+                  <Title headingLevel='h2'>Details</Title>
+                </div>
+                <FormGroup
+                  isRequired={true}
+                  key='name'
+                  fieldId='name'
+                  label={t`Name`}
+                  helperTextInvalid={t`Role name must be unique.`}
```
newswangerd, line 109: Don't hardcode arbitrary error messages.

Since it looks like the API doesn't respond with which field is invalid, you'll have to just display the error message at the top of the form, which is a pretty terrible user experience :/

src/containers/role-management/role-create.tsx
```diff
@@ -0,0 +1,294 @@
...
+                  validated={this.state.nameError ? 'error' : null}
+                >
+                  <TextInput
+                    id='role_name'
+                    value={this.state.name}
+                    onChange={(value) => {
+                      this.setState({ name: value, nameError: false });
+                    }}
+                    type='text'
+                    validated={this.state.nameError ? 'error' : null}
```
newswangerd, line 119: We need to do some extra validation here. If the role name doesn't start with `galaxy.` we need to throw an error telling the user that all custom roles must start with the prefix `galaxy.`.

src/containers/role-management/role-create.tsx
```diff
@@ -0,0 +1,294 @@
...
+    return (
+      <React.Fragment>
+        <BaseHeader
+          breadcrumbs={<Breadcrumbs links={breadcrumbs}></Breadcrumbs>}
+          title={title}
+        ></BaseHeader>
+        {notAuthorised ? (
+          <EmptyStateUnauthorized />
+        ) : (
+          <Main>
```
newswangerd, line 98: The spacing on this form feels wrong. You may not be using the patternfly components correctly. Don't worry about fixing this now. If you need to merge this without fixing it, just make sure we create an issue to get this done before the final release.

src/containers/role-management/role-edit.tsx
```diff
@@ -0,0 +1,418 @@
...
+            <Trans>{role.description}</Trans>
+          </div>
+        </BaseHeader>
+        {unauthorised ? (
+          <EmptyStateUnauthorized />
+        ) : (
+          <Main>
+            <section className='body'>
+              <div>
+                <div style={{ paddingBottom: '8px', paddingTop: '16px' }}>
```
newswangerd, line 222: Please show the role name name in a disabled text field

Here's an example from the Namespace edit form.
Screen Shot 2022-04-10 at 8 18 56 PM
.

src/containers/role-management/role-edit.tsx
```diff
@@ -0,0 +1,418 @@
...
+          }
+        >
+          <div style={{ paddingBottom: '10px' }}>
+            <Trans>{role.description}</Trans>
+          </div>
+        </BaseHeader>
+        {unauthorised ? (
+          <EmptyStateUnauthorized />
+        ) : (
+          <Main>
```
newswangerd, line 219: There's a lot of duplicate code here from the edit view. This needs to be abstracted out into a reusable component that can be used on both pages.

src/containers/role-management/role-list.tsx
```diff
@@ -0,0 +1,493 @@
...
+      </Tooltip>,
+      // ),
+    ];
+    return dropdownItems;
+  };
+
+  private queryRoles = () => {
+    const { params } = this.state;
+    this.setState({ loading: true }, () => {
+      RoleAPI.list(params)
```
newswangerd, line 447: always load this page with the `?name__startswith=galaxy.` query parameter. We only want to show the galaxy roles here to avoid confusion with all of the pulp roles that get created.

newswangerd (changes requested):

src/containers/role-management/role-list.tsx
```diff
@@ -0,0 +1,493 @@
...
+                            id: 'name',
+                          },
+                          {
+                            title: t`Description`,
+                            type: 'none',
+                            id: 'description',
+                          },
+                          {
+                            title: t`Created`,
+                            type: 'none',
```
ZitaNemeckova, line 283: This should be `type: 'number',` so it's a sortable column.

src/containers/role-management/role-create.tsx
ZitaNemeckova: API returns something like this `{name: 'This field must be unique'}` so it can be cleaned to be nicer. But maybe it should be a followup PR?

src/containers/role-management/role-create.tsx
```diff
@@ -0,0 +1,294 @@
...
+                  />
+                </FormGroup>
+
+                <FormGroup
+                  isRequired={true}
+                  key='description'
+                  fieldId='description'
+                  label={t`Description`}
+                  helperTextInvalid={
+                    !this.state.roleError ? null : this.state.roleError.name
```
ZitaNemeckova, line 130: This is always returning `null`.

src/containers/role-management/role-create.tsx
```diff
@@ -0,0 +1,294 @@
...
+      </React.Fragment>
+    );
+  }
+  private saveRole = () => {
+    const { name, permissions, description } = this.state;
+    RoleAPI.create({ name, description, permissions })
+      .then(() => this.setState({ redirect: Paths.roleList }))
+      .catch((err) => {
+        err.response.status === 400
+          ? this.setState({ nameError: true })
```
ZitaNemeckova, line 259: This will need to check which of the fields is problematic.

src/containers/role-management/role-create.tsx
```diff
@@ -0,0 +1,294 @@
...
+          : this.setState({ roleError: mapErrorMessages(err) });
+      });
+  };
+
+  private checkLength = (input) => {
+    if (input.toString().length > 128) {
```
ZitaNemeckova, line 265: This should be `input.length === 0 || input.length > 128`

src/containers/role-management/role-create.tsx
```diff
@@ -0,0 +1,294 @@
...
+    } else {
+      return false;
+    }
+  };
+
+  private helperText = (input) => {
+    let text = null;
+    if (input === '') {
+      text = t`This field may not be blank.`;
+    } else if (input.toString().length > 150) {
```
ZitaNemeckova, line 276: This should be 128, right?

src/api/role.ts
```diff
@@ -0,0 +1,24 @@
...
+      delete changedParams['sort'];
+    }
+    return super.list(changedParams);
+  }
+
+  updatePermissions(id, data: unknown) {
+    return this.http.patch(this.apiPath + id, data);
+  }
+
+  createRole(data: unknown) {
```
ZitaNemeckova, line 19: This is never used and not needed :)

ZitaNemeckova:
@himdel Tested in UI. It's working in general but needs few fixes. I think it could go in and be fixed later. But not my call to make.

Needs fixing in UI:
- Error handling in the form - making difference in 400 errors - name x description
- Validation in the form - allowed lenght, required fields, **name should start with `galaxy`** (this maybe the most important)
- Missing sort for `Created` column 
- UX fixes

Needs fixing code-wise:
- unify Edit and Create in one component
- remove unused code

src/containers/role-management/role-create.tsx
```diff
@@ -0,0 +1,294 @@
...
+    const groups = Constants.PERMISSIONS;
+    const { permissions: selectedPermissions, name } = this.state;
+
+    const { featureFlags } = this.context;
+    let isUserMgmtDisabled = false;
+    const filteredPermissions = { ...Constants.HUMAN_PERMISSIONS };
+    if (featureFlags) {
+      isUserMgmtDisabled = featureFlags.external_authentication;
+    }
+    if (isUserMgmtDisabled) {
```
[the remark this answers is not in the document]
ShaiahWren: OK, does this mean remove this else block entirely?

src/api/role.ts
```diff
@@ -0,0 +1,20 @@
+import { PulpAPI } from './pulp';
+
+export class API extends PulpAPI {
+  apiPath = 'roles/';
+
+  list(params) {
```
himdel, line 6: When(/if) rebasing on top of #1797, the `list` method can go away, as it's implemented in the parent class.
The apiPath is the same for both PRs so you'll just need to make sure the result has methods from both, except list :).

src/containers/role-management/role.scss
```diff
@@ -0,0 +1,3 @@
+a {
```
himdel, line 1: this cannot be a global style .. which particular link should this apply to?

src/containers/role-management/role-edit.tsx
```diff
@@ -0,0 +1,431 @@
...
+  name: string;
+  description: string;
+  roleError: ErrorMessagesType;
+  nameError: boolean;
+  descriptionError: boolean;
+  descriptionMessage: string;
+}
+
+class EditRole extends React.Component<RouteComponentProps, IState> {
+  _isMounted = false;
```
himdel, line 73: why this `_isMounted`? Doesn't seem to be used for anything..

src/containers/role-management/role.scss
ShaiahWren: I did away with this:)

src/containers/role-management/role-edit.tsx
ShaiahWren: GONE

src/containers/role-management/role-create.tsx
```diff
@@ -0,0 +1,186 @@
...
+  redirect?: string;
+  permissions: string[];
+  name: string;
+  description: string;
+  nameHelperText: string;
+  nameValidated: string;
+  alerts: AlertType[];
+}
+
+class RoleCreate extends React.Component<RouteComponentProps, IState> {
```
newswangerd, line 32: There's still a ton of duplicate functionality in role-create and role-edit. Any functions that you've created here that are shared across both pages (such as `validateInput`) need to be de-duplicated. You could look at moving them into the `RoleForm` component or put them into another module that can be imported by both files.

src/containers/role-management/role-edit.tsx
```diff
@@ -0,0 +1,317 @@
+import { t } from '@lingui/macro';
+// import { i18n } from '@lingui/core';
```
newswangerd, line 2: don't leave this commented out. Delete it if you don't need it.

src/containers/role-management/role-list.tsx
newswangerd: I'm wondering if we should make the `name__startswith` filter part of the UI, so that users can see the pulp roles if they want to.

In other words, by default the `?name__startswith=galaxy.` filter is applied when the page loads, but users have the ability to remove it to see all the roles if they want. @himdel what do you think?

src/containers/role-management/role-edit.tsx
```diff
@@ -0,0 +1,317 @@
...
+      text = null;
+    }
+    return text;
+  };
+
+  private validateInput = (input, field) => {
+    const error = { ...this.state.errorMessages };
+    if (input === '') {
+      error[field] = t`This field may not be blank.`;
+    } else if (input.toString().length > 128) {
```
newswangerd, line 234: I don't think there are any limitations on how long or short the description can be, so you don't need to check this here. Just verify that description isn't blank.

newswangerd (changes requested):
Functionality wise, this looks pretty good. There are still some code quality issues to clean up though.

himdel:
> In other words, by default the ?name__startswith=galaxy. filter is applied when the page loads, but users have the ability to remove it to see all the roles if they want. @himdel what do you think?

That already happens in Approvals, so it sounds reasonable :).

newswangerd (approved):

PR #1864, opened 2022-04-07
RemoteForm - only show "Download only signed collections" for remotes, not registries

himdel:
Follow-up to https://github.com/ansible/ansible-hub-ui/pull/1625 (AAH-1333)

the new switch was also visible for remoteType=registry, hiding in remote registry forms :)

| |changed|unchanged|
|-|-|-|
|Before:|||
|After:|||


ZitaNemeckova (approved):
Tested in UI. LGTM 👍

PR #1893, opened 2022-04-14
frontend-components-config: downgrade from 4.6.6 to 4.6.4

himdel:
(Already merged in master-stable, works on [stage](https://console.stage.redhat.com/ansible/automation-hub), this is to fix it in master and thus [stage beta](https://console.stage.redhat.com/beta/ansible/automation-hub).)

Revert #1814

frontend-components-config 4.6.6 seem to cause a "Something went wrong" screen on cloud.
Downgrading to 4.6.4 which works.

---

(don't merge yet, we're still trying to figure out what actually changed...
MM points to `window.automationHub.init()` throwing an exception,
could be related to an extra `The resource from “https://console.stage.redhat.com/beta/apps/automation-hub/css/automationHub.f5162efe821be5a64169.css” was blocked due to MIME type (“text/css”) mismatch (X-Content-Type-Options: nosniff).` error)


himdel:
works on stage now, so we'll go with #1975

PR #1917, opened 2022-04-19
Drop unused .firebaserc

himdel:
Random cleanup, this file comes from the inital commit, I don't believe we use firebase anywhere :)


MilanPospisil (approved):
LGTM 👍 My grep also did not found any firebaserc string in the app, so I hope this is correct :-)

PR #858, opened 2021-08-31
[stable-4.3] Update to webpack 5, frontend-components-config 4, patternfly current (#742)

himdel:
Backports #742 to stable-4.3.

---

Also contains bits from #579 , #850 to resolve conflicts.

PR #1918, opened 2022-04-19
[stable-4.2] dependency cleanup, github files, README

himdel:
This drops unused `.firebaserc` (#1917),

drops insights mode and Deploying sections from README and remove `profiles/` (insights mode updated in #1526, deploying is up-to-date only on master),

remove github files that only work on master: `PULL_REQUEST_TEMPLATE.md`

cleans up unused npm dependencies - backports most of #325, and non-ts parts of #579, similar to #858 for 4.3
removes husky (only needed on master)


himdel:
Merging, dependabot will have less deps to work with.

PR #1922, opened 2022-04-19
Fix broken UI on EEs created by ansible builder

newswangerd:
[Issue: AAH-1527](https://issues.redhat.com/browse/AAH-1527)

### DIFF
--- a/CHANGES/1527.bugfix
+++ b/CHANGES/1527.bugfix
@@ -1,0 +1,1 @@
+Fix an error where images created by ansible builder couldn't be inspected in the UI.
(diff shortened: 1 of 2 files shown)


himdel (approved):
LGTM, we're hiding a couple of layers, but ones we have no info about, makes sense :)

(test failures unrelated)

Tested with..

```sh
mkdir tmp
cd tmp
pip install ansible-builder
touch execution-environment.yml
ansible-builder create
cd context
buildah build
podman image tag $HASH localhost:5001/magic:latest
podman login localhost:5001 --tls-verify=false
podman push localhost:5001/magic:latest --tls-verify=false
```

PR #1974, opened 2022-04-25
Update to use v3 apis instead of legacy ; collection test fix

himdel:
This updates us to use the new v3/plugin/ansible urls for collection deletion, deprecation and download urls,
updates the test to wait for the right request/response.
(redirects from https://github.com/pulp/pulp_ansible/blob/main/pulp_ansible/app/urls.py#L43)

This should fix the `collection.js` test failure:
https://github.com/ansible/ansible-hub-ui/runs/6152044308?check_suite_focus=true#step:17:249

---

Also updating galaxykit to (PR) to fix the `collection-list.js` test failure:
https://github.com/ansible/ansible-hub-ui/runs/6152044308?check_suite_focus=true#step:17:312

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -26,7 +26,7 @@ jobs:
 
     - name: "Install galaxykit dependency"
       run: |
-        pip install git+https://github.com/ansible/galaxykit.git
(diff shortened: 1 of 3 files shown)


himdel:
Merging now, to get tests in master, stable-4.5 and feature/rbac-roles green again.

PR #1975, opened 2022-04-25
Bump @redhat-cloud-services/frontend-components-config to 4.6.9+

himdel:
Manually bumping `@redhat-cloud-services/frontend-components-config` to 4.6.13, `@redhat-cloud-services/frontend-components-config-utilities` to 1.5.20.

> #### Disabling ESI tags in insights mode
>
> We will be dropping the support and soon your development environments will stop working if you don't upgrade your shared webpack config to [@redhat-cloud-services/frontend-components-config/v/4.6.9](https://www.npmjs.com/package/@redhat-cloud-services/frontend-components-config/v/4.6.9) 
>
> If you are not using the shared config, and only using the proxy from utilities you will have to add a new webpack dev server middleware setup. You can check an example here: https://github.com/project-koku/koku-ui/blob/main/webpack.config.js#L253
>
> I would like to ask you to finish with this change ASAP.

(Also related to https://github.com/ansible/ansible-hub-ui/pull/1893)

---

Also dropping mentions of ESI from README and index.html, which is now only used by standalone.


himdel:
TODO
bump f-c-c-utilities too,
drop ESI mentions from readme

himdel:
Tested insights dev, standalone dev and compared standalone build files .. I think this is ready :)

rochacbruno (approved):

PR #1863, opened 2022-04-07
New UI for object permission assignment

himdel:
Fixes AAH-1129
Depends on:
* API:
  * https://github.com/ansible/galaxy_ng/pull/1057
  * https://github.com/ansible/galaxy_ng/pull/1058
  * (revert https://github.com/ansible/galaxy_ng/pull/1205)
  * (unrevert https://github.com/ansible/galaxy_ng/pull/1212)
  * => [feature/rbac-roles]( https://github.com/ansible/galaxy_ng/tree/feature/rbac-roles) branch
* UI:
  * #1797
  * #1860
  * #1980
  * #1981
  * => [feature/rbac-roles]( https://github.com/ansible/ansible-hub-ui/tree/feature/rbac-roles) branch

Related to https://github.com/ansible/galaxykit/pull/39

Changes:
* Namespace detail - add a new "Namespace owners" tab, allows assigning a group with some roles to the namespace
* Namespace create modal - removed groups from modal, changed the on success redirect to go directly to Namespace owners
* Namespace edit - removed groups from form, left an info box in its place
* EE edit - removed groups from edit modal, left info box
* EE detail - add a new "Owners" tab, same as in namespaces, different set of permissions

| where | before | after |
|-|-|-|
| namespace create |  |  after, goes to  |
| namespace edit |  |  |
| remote EE edit |  |  |
| local EE edit |  |  |

| added | namespace | remote EE | local EE |
|-|-|-|-|
| empty state |  |  |  |

PR #1978, opened 2022-04-26
[rbac] Fix CI, unify breadcrumb types

himdel:
(moving out of #1863)

Unify the type of breadcrumbs to a new `BreadcrumbType[]`, to fix a `no-implicit-any` warning (from #1860 since the linter wasn't on here).

And disable bits of `namespace.js` tests breaking on roles.


himdel:
Merging to get #1980, #1981 testable

PR #1981, opened 2022-04-26
[rbac] GroupRoleAPI, RoleAPI, PulpAPI - move logic to components

himdel:
(moving out of https://github.com/ansible/ansible-hub-ui/pull/1863, follow-up to #1797)

`GroupRoleAPI` - move `getRolesWithPermissions` outside api, keep only `listRoles`
`GroupRoleAPI`, `PulpAPI` - use `apiPath` instead of custom constructor.
`RoleAPI` - don't hardcode filter, move to components using it
`TaskManagementAPI`: use PulpApi.list since the ordering logic was moved there

This should not actually change anything - except in role list, clearing the `galaxy.` filter  will actually work.

### DIFF
--- a/src/api/group-role.ts
+++ b/src/api/group-role.ts
@@ -14,6 +14,8 @@ class API extends PulpAPI {
   addRoleToGroup(groupId, role) {
     return this.http.post(`${this.apiPath}${groupId}/roles/`, {
       role: role.name,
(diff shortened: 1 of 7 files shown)


himdel:
tested in users, groups, roles and tasks.. still working :)

PR #1982, opened 2022-04-26
group-detail: fix success user remove message

himdel:

Just a group edit success alert fix: User .. has been successfully removed *from* group ...


ShaiahWren (approved):
Good catch.

PR #1983, opened 2022-04-26
ListItemActions: don't render 0 when an array is empty

himdel:

When `buttons` or `kebabItems` are empty arrays, the length is 0 and a zero can end up displayed instead of the buttons.


ShaiahWren (approved):
LGTM

PR #1984, opened 2022-04-26
RoleListTable ExpandableRow - require rowIndex, expand breaks without it

himdel:

Marking `rowIndex` non-optional in `ExpandableRow` - without it, the exapndable rows don't render right and end up without the expand chevron.

(@ShaiahWren noticed this during #1860 , making sure we don't hit this again :).)


ShaiahWren (approved):
Works for me.

PR #1989, opened 2022-04-28
webpack config ci check: compare using the right isBuild value

himdel:
The webpack config is based on NODE_ENV and ignoring TARGET_ENVIRONMENT now,
I'll fix that separately, but now it also means the webpack config change check is not actually checking the build config, just the dev one.

So, ensuring any `config/*.prod.webpack.config.js` will run with `NODE_ENV=production`.

Needed to properly test https://github.com/ansible/ansible-hub-ui/pull/1526, https://github.com/ansible/ansible-hub-ui/pull/1975 and https://github.com/ansible/ansible-hub-ui/pull/1893.

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -112,6 +112,16 @@ jobs:
           npm install
 
           for file in *.js; do
+            # FIXME: that's what TARGET_ENVIRONMENT is for
+            NODE_ENV=
+            if grep -q prod <<< "$file"; then
+              NODE_ENV=production
+            fi
+            export NODE_ENV
+
+            # eliminate differences caused by branch name based logic in f-c-config
+            export BRANCH="${{ github.base_ref }}"
+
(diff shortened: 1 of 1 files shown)


himdel:
Tested with..

```diff
diff --git a/config/webpack.base.config.js b/config/webpack.base.config.js
index 26b06a79..8fc7055a 100644
--- a/config/webpack.base.config.js
+++ b/config/webpack.base.config.js
@@ -171,7 +171,7 @@ module.exports = (inputConfigs) => {
           exposes: {
             './RootApp': resolve(
               __dirname,
-              isBuild ? '../src/app-entry.js' : '../src/dev-entry.js',
+              isBuild ? '../src/nap-entry.js' : '../src/dev-entry.js',
             ),
           },
         },
```

=> https://github.com/ansible/ansible-hub-ui/runs/6214201727?check_suite_focus=true

breaks even on problems depending on isBuild, which comes from NODE_ENV.

himdel:
linter only PR, merging to get the other 3 ready

PR #1991, opened 2022-04-29
AboutModal: show galaxy_ng_commit when available

himdel:
since AAH-1360, https://github.com/ansible/galaxy_ng/pull/1191, the api returns the galaxy_ng_commit field,
based on GIT_COMMIT at runtime

since https://github.com/ansible/ansible-hub-ui/pull/1559, the UI is displaying the ui commit in AboutModal.

So, making sure we can display both, when available.

Won't actually do anything now because the var is only added in crc build, aka the place without about modal..
but if it did:

### DIFF
--- a/src/components/about-modal/about-modal.tsx
+++ b/src/components/about-modal/about-modal.tsx
@@ -24,14 +24,22 @@ interface IProps {
 }
 
 interface IState {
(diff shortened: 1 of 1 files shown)


ZitaNemeckova (approved):
LGTM 👍

PR #1992, opened 2022-04-29
Do not use global active CSS selector

MilanPospisil:
https://issues.redhat.com/browse/AAH-1546


himdel:
:+1: wraps the use of these clases in a hub- class,
the rest can be addressed as part of #1657

himdel (approved):

PR #2026, opened 2022-05-02
[AAH-1060] Fix context of signing filter component in the search bar

ironfroggy:
Issue: AAH-1060

Fix the filter component in the search bar by wrapping it in AppContext.Consumer to inject feature flags

The search filter for signing status was reading the feature flag to enable the feature from the component's `this.context`, which is not the correct context because other contexts are used between this component and the top-level AppContext. Fixed by wrapping with AppContext.Consumer

* [[PR relevant issue or pull_request](https://issues.redhat.com/browse/AAH-1060)](AAH-1060)

### DIFF
--- a/src/components/collection-list/collection-filter.tsx
+++ b/src/components/collection-list/collection-filter.tsx
@@ -10,6 +10,7 @@ import {
 
 import { AppliedFilters, CompoundFilter } from 'src/components';
 import { Constants } from 'src/constants';
+import { AppContext } from 'src/loaders/app-context';
 
 interface IProps {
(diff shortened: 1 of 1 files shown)


brumik:
It is enough for the class components if you put this after the class: 
```
CollectionFilter.contextType = AppContext;
```

otherwise I think this is a good catch.

brumik (approved):
LGTM

himdel:
PR checks...

> ERROR: Commit 407eed7f is not signed

himdel (approved):

PR #2030, opened 2022-05-03
Localization stable-4.5 string update

rooftopcellist:
### Summary

Localization Update from 3-28 push.  Soon to follow, the translated strings from the 4-19 push.


locale/es.po
```diff
@@ -3291,54 +2797,46 @@ msgstr "liberado {0}"
...
+#: src/components/collection-list/collection-list-item.tsx:57
+#: src/components/headers/collection-header.tsx:320
 #: src/components/headers/partner-header.tsx:46
 msgid "{company} logo"
 msgstr "{company} Logotipo"
 
 #: src/utilities/get_human_size.ts:14
 msgid "{n, plural, one {# byte} other {# bytes}}"
-msgstr "{n, plural, one {# byte} other {# bytes}}"
+msgstr "{n, plural, one {# byte} otros {# bytes}}"
```
himdel, line 2834:
```suggestion
msgstr "{n, plural, one {# byte} other {# bytes}}"
```

`plural`, `one`, `other` are part of the gettext syntax, not something to translate

rooftopcellist: yep, this commonly trips up translators.  I fixed it in all the languages that needed it.

himdel (approved):
Missing `locale/ko.js` because we also needed https://github.com/ansible/ansible-hub-ui/pull/2038/files#diff-dc0af3f29e288de853e3010325e282a14cffca3e38def0e839384df72718fb04R7 for that.

But that can happen in the second pr as well :)

rooftopcellist:
Closing in favor of https://github.com/ansible/ansible-hub-ui/pull/2039

PR #2034, opened 2022-05-03
Add crc banner about sync toggle support

himdel:
Issue: AAH-1552


himdel:
updated to "The Automation Hub sync toggle is now only supported in AAP 2.0. Previous versions of AAP will continue automatically syncing all collections."

awcrosby (approved):

PR #2040, opened 2022-05-04
Remove warning when including and excluding tags.

newswangerd:
Including and excluding a set of tags is a valid usecase because pulp_container can do pattern matching. For example I can specify:

```
"include": ["version-1.2.*"],
"exclude": ["version-1.2.0", "version-1.2.1"]
```

### DIFF
--- a/src/components/execution-environment/repository-form.tsx
+++ b/src/components/execution-environment/repository-form.tsx
@@ -351,14 +351,6 @@ export class RepositoryForm extends React.Component<IProps, IState> {
                   ))}
                 </LabelGroup>
               </FormGroup>
-
-              {includeTags.length && excludeTags.length ? (
-                <Alert
-                  variant='warning'
-                  isInline
-                  title={t`It does not make sense to include and exclude tags at the same time.`}
-                />
-              ) : null}
             </>
           )}
 


himdel (approved):
LGTM :+1: 

Though we should probably add a note that wildcards are possible, annd that exclude wins over include

himdel:
What if we just did sometthing like this....

```diff
diff --git a/src/components/execution-environment/repository-form.tsx b/src/components/execution-environment/repository-form.tsx
index 6d82928d..d5a4c8d1 100644
--- a/src/components/execution-environment/repository-form.tsx
+++ b/src/components/execution-environment/repository-form.tsx
@@ -354,9 +354,9 @@ export class RepositoryForm extends React.Component<IProps, IState> {
 
               {includeTags.length && excludeTags.length ? (
                 <Alert
-                  variant='warning'
+                  variant='info'
                   isInline
-                  title={t`It does not make sense to include and exclude tags at the same time.`}
+                  title={t`When including and excluding tags at the same time, first all tags matching included tags are added, then all tags matching excluded tags are removed. Wildcards are supported.`}
                 />
               ) : null}
             </>
```


(or would it make sense to go with HelperText?)

(the PR is approved either way :))

PR #2059, opened 2022-05-09
Delete non empty namespace test

ShaiahWren:
No-Issue, fills in another test from. https://github.com/himdel/ansible-hub-ui/issues/1 in the `Collections > Namespaces` department: "can only delete [namespace] when empty".

### DIFF
--- a/test/cypress/integration/namespace_delete.js
+++ b/test/cypress/integration/namespace_delete.js
@@ -56,7 +56,13 @@ describe('Delete a namespace', () => {
 
     // sign collection
 
+    cy.intercept(
+      'GET',
+      Cypress.env('prefix') +
+        '_ui/v1/repo/published/?deprecated=false&offset=0&limit=10',
+    ).as('collections');
     cy.menuGo('Collections > Collections');
+    cy.wait('@collections');
     cy.contains('network').click();
     cy.get('[data-cy="kebab-toggle"]').click();
     cy.contains('Sign entire collection').click();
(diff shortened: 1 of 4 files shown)


himdel:
you'd need `cy.settings({ GALAXY_REQUIRE_CONTENT_APPROVAL: true })` to enable approval if you need it.

(Though, sounds like here you'd just need to wait until the import is successful since you don't actually need manual approval to test deleting namespaces.)

test/cypress/integration/namespace_delete.js
```diff
@@ -21,4 +22,53 @@ describe('Delete a namespace', () => {
...
+    ).as('reload');
+    cy.galaxykit('-i namespace create', 'ansible');
+    cy.menuGo('Collections > Namespaces');
+    cy.wait('@reload');
+
+    cy.get('a[href*="ui/repo/published/ansible"]').click();
+
+    //upload a collection
+
+    cy.contains('Upload collection').click();
```
brumik, line 40: Why not upload the collection trough galaxykit? I thought we have separate tests for testing upload collection.

brumik (commented):
Martin: `cy.settings({ GALAXY_REQUIRE_CONTENT_APPROVAL: true })` would not work, since we are overriding in the env. The PR #2069 tries to fix this.

Shaiah: Wouldn't be easier to use galaxykit at the `before` and test on existing data? So example
```js
before(() => {
  cy.deleteNamespacesAndCollections();
  galaxikit('namespace create test_delete_success');
  galaxikit('namespace create test_delete_non_empty');
  galaxikit('collection upload test_delete_non_empy coll1');
});

after(() => { cy.deleteNamespacesAndCollections(); });
```

Also `deleteNamespacesAndCollections()` is a bit slow, so using the above pattern would maybe make it more performant too.

himdel (approved):

ShaiahWren:
pr closed.

PR #1923, opened 2022-04-20
[WIP] Adding test for rejection, approval and test for view import logs.

MilanPospisil:
Issue: [AAH-628](https://issues.redhat.com/browse/AAH-628)

https://issues.redhat.com/browse/AAH-628

Finish the rest of the functionality in issue.

Note - this PR is new version of old closed PR - https://github.com/ansible/ansible-hub-ui/pull/1527.
The old PR has been long waiting for the galaxykit so it started to become obsolete and it was easier to create new PR than to repair the old one (and for example broken galaxykit collection deletion makes galaxykit deleting data useless anyway). 

cy.settings({ GALAXY_REQUIRE_CONTENT_APPROVAL: true }) is not present in this test because it seems no longer creates unapproved collections - this may be because Signing changes some things in the backend.

After issues will be repaired, whole approval process can be completed.

PR #2060, opened 2022-05-10
test the Collections > Approval page

ShaiahWren:
No-Issue

Fills in a few more test from https://github.com/himdel/ansible-hub-ui/issues/1, Collections > Approval page.

- [ ] has a default Needs Review filter
- [ ] links to the import log page
- [ ] updates certification status - reject and approve labels

### DIFF
--- a/test/cypress/integration/collection_approval.js
+++ b/test/cypress/integration/collection_approval.js
@@ -9,11 +9,12 @@ describe('tests the approval list screen ', () => {
     cy.deleteNamespacesAndCollections();
     cy.galaxykit('-i namespace create', 'ansible');
     cy.galaxykit('-i collection upload ansible network');
-    cy.menuGo('Collections > Approval');
+    cy.visit('/ui/approval-dashboard');
   });
-  // after(() => {
-  //   cy.settings();
-  // });
+
+  after(() => {
+    cy.settings();
+  });
 
   it('has a default Needs Review filter', () => {
     cy.get('.pf-c-chip-group__list').contains('Needs Review');
@@ -38,7 +39,6 @@ describe('tests the approval list screen ', () => {
     cy.get('button[aria-label="Actions"]:first').click();
     cy.contains('Approve').click();
     cy.get('tr').eq(1).contains('Approved');
-    cy.settings();
   });
 
   it('view the imports logs', () => {
--- a/test/cypress/integration/collection_approval.js
+++ b/test/cypress/integration/collection_approval.js
@@ -1,0 +1,59 @@
+describe('tests the approval list screen ', () => {
+  beforeEach(() => {
+    cy.login();
+    cy.deleteNamespacesAndCollections();
+    cy.galaxykit('-i namespace create', 'ansible');
+    cy.galaxykit('-i collection upload ansible network');
+    cy.menuGo('Collections > Approval');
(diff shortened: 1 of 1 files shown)


ShaiahWren:
@himdel, this passes locally. It looks like the 'Sign and approve' button isn't visible for gh and I don't know why. Any insights?

himdel:
sure, you're running with signing enabled, the CI isn't :)

So.. what do you actually want to test? If this is about collection approval, you don't need signing, so this should assume signing is not enabled.

(If you do need signing, we'll have to use `cy.settings` to ensure signing is enabled.)

brumik:
@ShaiahWren I think you can test for only "Approve" (does not need any flags), since "Sign and Approve" requires at least 3 extra flags (I am debugging it) and we need tests for only "Approve" too. I am testing the "Sing and Approve" in the signing cypress tests here: #2069.

MilanPospisil:
Ah, this looks like exactly what Im already doing in this PR: https://github.com/ansible/ansible-hub-ui/pull/1923. It is nearly done, and I can see, yours too :-)

test/cypress/integration/collection_approval.js
```diff
@@ -0,0 +1,67 @@
+describe('tests the approval list screen ', () => {
+  before(() => {
+    cy.settings({
+      GALAXY_AUTO_SIGN_COLLECTIONS: true,
+    });
+  });
```
himdel, line 6:
```suggestion

  before(() => {
    cy.settings({
      GALAXY_REQUIRE_CONTENT_APPROVAL: true,
    });
  });
  
```

test/cypress/integration/collection_approval.js
```diff
@@ -0,0 +1,67 @@
...
+  after(() => {
+    cy.settings();
+  });
+
+  it('has a default Needs Review filter', () => {
+    cy.get('.pf-c-chip-group__list').contains('Needs Review');
+  });
+
+  it.skip('rejects certification status and approves it again', () => {
```
himdel, line 22:
```suggestion
  it('rejects certification status and approves it again', () => {
```

test/cypress/integration/collection_approval.js
```diff
@@ -0,0 +1,67 @@
...
+    //reject
+
+    cy.get('button[aria-label="Actions"]:first').click();
+    cy.contains('Reject').click();
+    cy.get('tr').eq(1).contains('Rejected');
+
+    //approve
+    cy.get('button[aria-label="Actions"]:first').click();
+    cy.contains('Approve').click();
+    cy.get('tr').eq(1).contains('Signed and approved');
```
himdel, line 40:
```suggestion
    cy.get('tr').eq(1).contains('Approved');
```

himdel (commented):
I think you want something like this?

Signing is not available in CI, so we should not be enabling it for these tests.
The tests themselves can work, you don't need to `skip` one.

test/cypress/integration/collection_approval.js
himdel: (unless you don't need approval enabled either? .. but if this is testing the approval page, it should probably test the normal process where the collection has to be approved first)

himdel (approved):

PR #2061, opened 2022-05-11
Delete Collections in Different Repositories

MilanPospisil:
Issue: AAH-1591

https://issues.redhat.com/browse/AAH-1591´

Ability to delete collection in different repositories. This PR adds ability to cypress to use new galaxykit command, that can delete also for example rejected collections.

This PR depends on: https://github.com/ansible/galaxykit/pull/25


test/cypress/support/commands.js
```diff
@@ -717,10 +717,29 @@ Cypress.Commands.add('deleteCollections', {}, (namespace) => {
   });
 });
 
+Cypress.Commands.add('deleteAllCollections', {}, () => {
+  let url =
+    Cypress.env('prefix') +
+    '_ui/v1/collection-versions/?sort=-pulp_created&offset=0&limit=100';
+  cy.request(url).then((res) => {
```
himdel, line 724: (this is still a future TODO (galaxykit should handle it), but works for now :))

test/cypress/support/commands.js
```diff
@@ -717,10 +717,29 @@ Cypress.Commands.add('deleteCollections', {}, (namespace) => {
...
+        'None',
+        record.repository_list[0],
+      );
+    });
+  });
+});
+
 Cypress.Commands.add('deleteNamespacesAndCollections', {}, () => {
+  cy.login();
+  cy.deleteAllCollections();
```
brumik, line 740: Unless we make the galaxykit to wait for the dispatched delete tasks to finish, we should add here a 10 sec wait, that should be enough to make sure that there are no collections in the namespace. Otherwise we cannot delete the namespace if there is a left over collection in it.

test/cypress/support/commands.js
brumik, line 724: Maybe we could use after merge a galaxykit command for this. I am trying to get this into galaxykit here: https://github.com/ansible/galaxykit/pull/45/files#diff-0b10698fda1acdd6a78ba7d73ce77f7a1f7fe08cc149a83e291bb8f2a0f44ae3R21

test/cypress/support/commands.js
```diff
@@ -717,10 +717,29 @@ Cypress.Commands.add('deleteCollections', {}, (namespace) => {
...
+        record.name,
+        'None',
+        record.repository_list[0],
+      );
+    });
+  });
+});
+
 Cypress.Commands.add('deleteNamespacesAndCollections', {}, () => {
+  cy.login();
```
brumik, line 739: Galaxykit does not need login, so if we can replace the only one fetch on line 724 this could be obsolete.

brumik (commented):
I really like this idea, this would make testing more stable. Also if we replace the `deleteCollections` method, we could remove that command in this PR.

This one: 
https://github.com/ansible/ansible-hub-ui/pull/2061/files#diff-b91907333f838f733b881c7aeb346223e7b657df79f22e72df11e5a12b1bffc3R710-R718

test/cypress/support/commands.js
MilanPospisil: Galaxykit does not. But Im using here request for the API to fetch all collection including rejected and others.

brumik: https://github.com/ansible/galaxykit/pull/45 this may solve this

test/cypress/support/commands.js
```diff
@@ -717,10 +717,29 @@ Cypress.Commands.add('deleteCollections', {}, (namespace) => {
...
+    Cypress.env('prefix') +
+    '_ui/v1/collection-versions/?sort=-pulp_created&offset=0&limit=100';
+  cy.request(url).then((res) => {
+    let data = res.body.data;
+    data.forEach((record) => {
+      cy.galaxykit(
+        'collection delete',
+        record.namespace,
+        record.name,
+        'None',
```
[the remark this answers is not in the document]
MilanPospisil: This will be next PR, but for now, I would like to have the same functionality ASAP, because one PR is awaiting for it.

test/cypress/support/commands.js
MilanPospisil: You are right, but 10 seconds is a lot I think, 5 seconds should be enough.

test/cypress/support/commands.js
```diff
@@ -717,10 +717,27 @@ Cypress.Commands.add('deleteCollections', {}, (namespace) => {
...
 
+Cypress.Commands.add('deleteAllCollections', {}, () => {
+  cy.galaxykit('collection list').then((res) => {
+    let data = JSON.parse(res[0]).data;
+    data.forEach((record) => {
+      cy.galaxykit(
+        'collection delete',
+        record.namespace,
+        record.name,
+        'None',
```
brumik, line 718: This command gives you a list of all `versions`. I think the foreach is good but we could filter it before for the same collection + different versions if we pass `None` for version. Or just pass the version there too, not sure how it impacts performance.

MilanPospisil: Hm, this is true. It may fail for different versions. So lets pass version here and if it will fail somewhere in future, we can fix it.

test/cypress/support/commands.js
```diff
@@ -707,20 +707,27 @@ Cypress.Commands.add('deleteContainersManual', {}, () => {
...
 
-Cypress.Commands.add('deleteCollections', {}, (namespace) => {
-  range(5).forEach(() => {
-    cy.galaxykit('namespace list-collections ' + namespace).then((json) => {
-      JSON.parse(json).data.forEach((collection) => {
-        cy.galaxykit('collection delete', namespace, collection.name);
-      });
+Cypress.Commands.add('deleteAllCollections', {}, () => {
+  cy.galaxykit('collection list').then((res) => {
+    let data = JSON.parse(res[0]).data;
```
brumik, line 712: Teoretically const is enough :) (this is a small detail I know, you can leave it if you want does not matter)

test/cypress/support/commands.js
```diff
@@ -707,20 +707,27 @@ Cypress.Commands.add('deleteContainersManual', {}, () => {
...
+        record.name,
+        record.version,
+        record.repository_list[0],
+      );
     });
   });
 });
 
 Cypress.Commands.add('deleteNamespacesAndCollections', {}, () => {
+  cy.login();
```
brumik, line 726: The login I believe is not needed anymore since galaxykit takes care of everything.

brumik (commented):
This looks like nearly finished. Also there is an unused import form lodash, see the failing PR-check.

test/cypress/support/commands.js
MilanPospisil: I dont understand what you mean?

brumik: Well, I think it is somewhere in the best practices that if you don't modify a variable you use `const` as constant instead of `let`. `const` will prevent changing the variable reference (but will still, unfortunately, allow you to mutate the object)

MilanPospisil: Ah, const, I red it wrong. I understand now.

brumik (approved):
LGTM if the CI passes as well :)

MilanPospisil:
> 



> LGTM if the CI passes as well :)

Martin said that we can ignore it for now, because it is failing everywhere.

PR #2062, opened 2022-05-11
No alert after clicking (un)deprecate

MilanPospisil:
Issue: AAH-1578

https://issues.redhat.com/browse/AAH-1578

Adding deprecation/undeprecation alert after deprecate/undeprecate button clicked.

PR #2063, opened 2022-05-11
add waitForTask to set deprecation before loadCollections

ShaiahWren:
Issue: AAH-1596

adding waitForTask to run after setDeprecation and before loadCollection and success handlers. 

Also see https://github.com/ansible/ansible-hub-ui/pull/2062 @MilanPospisil

### DIFF
--- a/src/containers/namespace-detail/namespace-detail.tsx
+++ b/src/containers/namespace-detail/namespace-detail.tsx
@@ -377,7 +377,13 @@ export class NamespaceDetail extends React.Component<IProps, IState> {
           !collection.deprecated,
           this.context.selectedRepo,
         )
-          .then(() => this.loadCollections())
+          .then((result) => {
+            const taskId = result.data.task.split('tasks/')[1].replace('/', '');
+            console.log('result id', result.data);
+            waitForTask(taskId).then(() => {
+              this.loadCollections();
+            });
+          })
           .catch(() => {
             this.setState({
               warning: t`API Error: Failed to set deprecation.`,
--- a/CHANGES/1596.bug
+++ b/CHANGES/1596.bug
@@ -1,0 +1,1 @@
+Wait for setDeprecation task before running loadCollections and success handler.


src/containers/namespace-detail/namespace-detail.tsx
```diff
@@ -377,7 +377,12 @@ export class NamespaceDetail extends React.Component<IProps, IState> {
           !collection.deprecated,
           this.context.selectedRepo,
         )
-          .then(() => this.loadCollections())
+          .then((result) => {
+            const taskId = result.data.task.split('tasks/')[1].replace('/', '');
```
MilanPospisil, line 381: This is good for now, but in my next PR (https://github.com/ansible/ansible-hub-ui/pull/1766) I added getIdFromTask which return taskId from the string. We should not copy paste this functionality over and over through the code.

src/containers/namespace-detail/namespace-detail.tsx
```diff
@@ -377,7 +377,12 @@ export class NamespaceDetail extends React.Component<IProps, IState> {
...
+            waitForTask(taskId).then(() => {
```
MilanPospisil, line 382: Maybe we can implement taskId parsing directly in waitForTask if string is not taskId, what do you think @himdel? Is the task string always returning string that begins with / ? Or we can implement lets say waitForTaskUrl that does parsing?

src/containers/namespace-detail/namespace-detail.tsx
```diff
@@ -377,7 +377,12 @@ export class NamespaceDetail extends React.Component<IProps, IState> {
...
+              this.loadCollections();
+            });
+          })
           .catch(() => {
             this.setState({
               warning: t`API Error: Failed to set deprecation.`,
```
MilanPospisil, line 388: This error handling does not seems right to me @himdel . This returns only if task failed to enter the queue, right? So this will fire only if server is unreachable or something like that. The error handling for the task is done in waitForTask, so error handling could go also there. But this is too much repeatable code around one simple task, I think we should refactor that in future, because we should do the exact same think in many many places. What say you @himdel?

src/containers/namespace-detail/namespace-detail.tsx
himdel, line 382: missing return (if we want the catch to apply to the result)

src/containers/namespace-detail/namespace-detail.tsx
himdel: I think this already exists in master as well... 

```
import { parsePulpIDFromURL } from 'src/utilities/parse-pulp-id';
const taskId = parsePulpIDFromURL(result.data.task);
```

himdel (approved):
Works for me :+1: 

(at some point we should also go through these and make sure they have the blue waiting for task alert, but that's a separate issue :))

ShaiahWren:
Pr. closed

PR #2066, opened 2022-05-13
Add missing alert name in EE list 

jerabekjiri:
Minor fixes which I noticed while writing tests. The EE name is missing in the alert title on adding execution environment.   

Before:
Screenshot from 2022-05-13 18-09-15

After:
Screenshot from 2022-05-13 17-52-12

Plus fixing console error on undefined value in textarea
Screenshot from 2022-05-13 18-08-46

### DIFF
--- a/src/containers/execution-environment-list/execution_environment_list.tsx
+++ b/src/containers/execution-environment-list/execution_environment_list.tsx
@@ -435,7 +435,7 @@ class ExecutionEnvironmentList extends React.Component<
         remotePulpId={pulp_id}
         distributionPulpId={distributionPulpId}
         formError={this.state.formError}
-        onSave={(promise) => {
+        onSave={(promise, form) => {
           promise
             .then(() => {
               this.setState(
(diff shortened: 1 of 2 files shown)


src/containers/execution-environment-list/execution_environment_list.tsx
```diff
@@ -446,12 +446,12 @@ class ExecutionEnvironmentList extends React.Component<
                     variant: 'success',
                     title: isNew ? (
                       <Trans>
-                        Execution environment &quot;{name}&quot; has been added
```
himdel, line 449: why is `itemToEdit.name` not working here?

`form` comes from repository form state, but ... `name` sets that state

ZitaNemeckova (approved):

src/containers/execution-environment-list/execution_environment_list.tsx
jerabekjiri: `itemToEdit` is populated and updated only on the editing execution environment, it's not updated on adding new EE, therefore we need to return the component's state to get the name :)

PR #2068, opened 2022-05-15
[AAH-1313] Added detailed info to the sign all modal

brumik:
Issue: AAH-1313

Additionally fixed an issue when the sign all on namespace displayed
the number of collections to be signed instead of number of
versions inside all those collections.

P.S.: The api has support for total_versions, signed_versions and unsigned_versions now, this PR utilizes on that feature.

**Before**:
Screenshot from 2022-05-15 10-25-47
-- Note the bug here, just in foob there are two version and in fooc there is one more, but namespace signing says 2 (number of collections instead of versions) --
Screenshot from 2022-05-15 10-26-00

**After**:
Screenshot from 2022-05-15 10-20-22
Screenshot from 2022-05-15 10-20-52
Screenshot from 2022-05-15 10-21-01

### DIFF
--- a/CHANGES/1313.feature
+++ b/CHANGES/1313.feature
@@ -1,0 +1,1 @@
+Added detailed information to the sign all modal
--- a/src/api/response-types/collection.ts
+++ b/src/api/response-types/collection.ts
@@ -66,6 +66,9 @@ export class CollectionListType {
     avatar_url: string;
     company: string;
   };
+  total_versions: number;
+  signed_versions: number;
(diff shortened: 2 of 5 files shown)


brumik:
@himdel Is the webpack-config check failing OK or what it want's me to (not) do?

himdel:
> @himdel Is the webpack-config check failing OK or what it want's me to (not) do?

You're fine, I just need to better filter out the extra messags in the check :)

PR #2088, opened 2022-05-17
add frontend crd

drodowic:
Description:
----------
* adds frontend.yaml to define the deployment of the frontend container used in the ephemeral environment

Issue: AAH-1598

### DIFF
--- a/CHANGES/1598.misc
+++ b/CHANGES/1598.misc
@@ -1,0 +1,1 @@
+Enable ephemeral pr_check.sh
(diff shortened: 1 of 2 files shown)


awcrosby (approved):
LGTM, seems this is needed for our ephemeral env in other areas to continue deploying as expected. The PR check can be made non-required for this repo until we are ready to use it.

himdel:
Any way to add this *without* adding an always-breaking PR check?



(should we keep the cloud stuff in the cloud branches instead of having perfectly reasonable PRs fail on this?)

PR #2092, opened 2022-05-18
[rbac] Role lists - default sort by name, show description, pretty "Galaxy only" in AppliedFilters

himdel:
Depends on https://github.com/pulp/pulpcore/issues/2703 for the sort to work => https://github.com/pulp/pulpcore/pull/2766 => https://github.com/ansible/galaxy_ng/pull/1279 => just the [branch](https://github.com/ansible/galaxy_ng/commits/feature/rbac-roles) now.

Prettify the "galaxy only" filter indicator in role list:

And make sure role lists are sorting by name everywhere by default, and showing the description where possible.

Current state:

* User Access > Roles
  * has description, locked/unlocked, expandable
  * sort by name works, not default => **made `name` default**
* User Access > Groups > detail
  * tab Access
    * just name, expandable => **added `description`, use permission data already available (since pulp 3.20)**
    * sort by name broken, not default => **it is default, indicator has problems because the tabs share state (for later), api should work after https://github.com/pulp/pulpcore/issues/2703 + removed the `sort` -> `ordering` logic from `GroupRoleAPI` .. then reenabled after https://github.com/pulp/pulpcore/pull/2849**
  * tab Access, modal Add roles
    * has description, (not expandable)
    * sort by name works, default but doesn't indicate it => **fixed**
* Collections > Namespaces > detail
  * tab Namespace owners, modal Select a group (step 2)
    * has description, (not expandable)
    * sort by name works, default but doesn't indicate it => **fixed**
  * tab Namespace owners > group detail
    * just name, expandable (permissions & description lazy loaded when expanding)
    * no sort => **sorting in ui, just by name**
  * tab Namespace owners > group detail, modal Add roles
    * has description, (not expandable)
    * sort by name works, default but doesn't indicate it => **fixed**
* Execution Environments > Execution Environments > detail
  * tab Owners, modal Select a group (step 2)
    * has description, (not expandable)
    * sort by name works, default but doesn't indicate it => **fixed**
  * tab Owners > group detail
    * just name, expandable (permissions & description lazy loaded when expanding)
    * no sort => **sorting in ui, just by name**
  * tab Owners > group detail, modal Add roles
    * has description, (not expandable)
    * sort by name works, default but doesn't indicate it => **fixed**

### DIFF
--- a/src/api/container-distribution.ts
+++ b/src/api/container-distribution.ts
@@ -2,6 +2,8 @@ import { PulpAPI } from './pulp';
 
 export class API extends PulpAPI {
   apiPath = 'distributions/container/container/';
+
+  // patch(id, data)
 }
 
 export const ContainerDistributionAPI = new API();
--- a/src/api/group-role.ts
+++ b/src/api/group-role.ts
@@ -2,6 +2,7 @@ import { PulpAPI } from './pulp';
 
 class API extends PulpAPI {
   apiPath = 'groups/';
+  useOrdering = false; // listRoles produces surprising order when using ?ordering=role, but works with ?sort=role
 
   listRoles(groupId, params?) {
     return super.list(params, `${this.apiPath}${groupId}/roles/`);
--- a/src/api/group-role.ts
+++ b/src/api/group-role.ts
@@ -2,7 +2,7 @@ import { PulpAPI } from './pulp';
 
 class API extends PulpAPI {
   apiPath = 'groups/';
-  useOrdering = false; // listRoles produces surprising order when using ?ordering=role, but works with ?sort=role
+  useOrdering = true;
 
   listRoles(groupId, params?) {
     return super.list(params, `${this.apiPath}${groupId}/roles/`);
--- a/src/api/pulp.ts
+++ b/src/api/pulp.ts
@@ -1,13 +1,15 @@
 import { BaseAPI } from './base';
 
 export class PulpAPI extends BaseAPI {
+  useOrdering = false; // translate ?sort into ?ordering in list()
+
   constructor() {
     super(API_HOST + PULP_API_BASE_PATH);
   }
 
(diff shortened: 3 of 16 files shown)


MilanPospisil:
User Access -> Roles. In the table, there is column Locked, while in the filter drop down, it is named as Status, which looks confusing to me. Is it really what it was desired?

MilanPospisil:
When Galaxy only filter is closed, it cant be set again (only by reloading) - is this supposed behavior?

MilanPospisil:
tab Namespace owners, modal Select a group (step 2) -  has description, (not expandable)
tab Owners, modal Select a group (step 2) - has description, (not expandable).

This does not seems to be true, I dont see any description in group selections. Anyway, groups have no description at all.

himdel:
> User Access -> Roles. In the table, there is column Locked, while in the filter drop down, it is named as Status, which looks confusing to me. Is it really what it was desired?

Good catch, we were supposed to rename this to "built-in"/"editable" to make the meaning clear... updating.

himdel:
> When Galaxy only filter is closed, it cant be set again (only by reloading) - is this supposed behavior?

Yeaah, kinda .. it is consistent with Approval Dashboard's Status: Needs Review filter.
So, it's the behaviour we have, but eventually we should come up with a better UI for default filters.

The idea is that the user should almost never care about non-galaxy roles, but we still want to make it possible to see the rest, on this screen only.

himdel:
> tab Namespace owners, modal Select a group (step 2) - has description, (not expandable)
> tab Owners, modal Select a group (step 2) - has description, (not expandable).
>
> This does not seems to be true, I dont see any description in group selections. Anyway, groups have no description at all.

Hence *step 2* :) - step 1 is groups, step 2 is roles.

MilanPospisil (approved):
LGTM :+1:

PR #2114, opened 2022-05-24
Sync delete collections - wait for collection versions to be empty

MilanPospisil:
https://issues.redhat.com/browse/AAH-1641

It is good to wait for collections to be empty while deleting them, because they are deleted in the background and galaxykit command only returns when the operation is queued. This may avoid some unexpected problems.

This PR depends on: https://github.com/ansible/galaxykit/pull/48

### DIFF
--- a/test/cypress/support/commands.js
+++ b/test/cypress/support/commands.js
@@ -723,7 +723,7 @@ Cypress.Commands.add('deleteAllCollections', {}, () => {
         cy.log('Collections are empty!');
       }
     });
-  }
+  };
 
   cy.galaxykit('collection list').then((res) => {
     const data = JSON.parse(res[0]).data;
--- a/test/cypress/support/commands.js
+++ b/test/cypress/support/commands.js
@@ -707,10 +707,8 @@ Cypress.Commands.add('deleteContainersManual', {}, () => {
 });
 
 Cypress.Commands.add('deleteAllCollections', {}, () => {
-  function waitForEmptyCollection(maxLoops)
-  {
-    if (maxLoops == 0)
-    {
+  function waitForEmptyCollection(maxLoops) {
+    if (maxLoops == 0) {
       cy.log('Max loops reached while waiting for the empty collections.');
       return;
     }
@@ -719,15 +717,14 @@ Cypress.Commands.add('deleteAllCollections', {}, () => {
 
     cy.galaxykit('collection list').then((res) => {
       const data = JSON.parse(res[0]).data;
-      if (data.length != 0)
-      {
+      if (data.length != 0) {
         waitForEmptyCollection(maxLoops - 1);
-      }else{
+      } else {
         cy.log('Collections are empty!');
       }
     });
   }
-  
+
   cy.galaxykit('collection list').then((res) => {
     const data = JSON.parse(res[0]).data;
     cy.log(data.length + ' collections found for deletion.');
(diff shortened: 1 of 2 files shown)


test/cypress/support/commands.js
```diff
@@ -707,8 +707,27 @@ Cypress.Commands.add('deleteContainersManual', {}, () => {
...
+      cy.log('Max loops reached while waiting for the empty collections.');
+      return;
+    }
+
+    cy.wait(3000);
+
+    cy.galaxykit('collection list').then((res) => {
+      const data = JSON.parse(res[0]).data;
+      if (data.length != 0) {
+        waitForEmptyCollection(maxLoops - 1);
```
brumik, line 721: I wouldn't go with recursion here. It is not needed and recursions can be potentially dangerous. You can do a 
```ts
while (true): { if (not empty) wait(5000) else return; }
```

MilanPospisil: You cant do that in async environment... because if not empty will be resolved outside of this loop.

test/cypress/support/commands.js
```diff
@@ -707,8 +707,27 @@ Cypress.Commands.add('deleteContainersManual', {}, () => {
 });
 
 Cypress.Commands.add('deleteAllCollections', {}, () => {
+  function waitForEmptyCollection(maxLoops) {
```
brumik, line 710: In this case I would maybe prefer arrow function. I leave up to you which one you prefer but here is a nice blogpost (5-10min reading) what are the differences (they are expected not to have their own context so they are better suited probably for this kind of small functions).

https://medium.com/swlh/javascript-arrow-functions-vs-regular-functions-5ec4a9076796

test/cypress/support/commands.js
MilanPospisil: The async/await seems worse to me, because it is basicaly doing the same thing while hiding the asynchronicity at the same time. And your solution while(true) can be also stuck in infinite loop just like recursion. But we have counter for that, so I think we will be fine :-)

brumik: My idea with async await would be to not to have recursion. But truth to be told, i am not 100% familiar with the cypress async queue which is different in some ways from plain Js. It was just an idea, i think the ultimate solution will be Galaxy kit wait for tasks finishing type of solution.

brumik: For now this will be good (as long as it works :)

MilanPospisil: But good point, I want to learn about cypress queues more. I think they are hierarchical - so when you are inserting new task using cy.whatever in then condition after some another task, it is inserted in the queue after that task and behind another tasks that were inserted after orginal task in sequential control flow. For example:

cy.Task1
cy.Task2 then Task2a then Task2b
cy.Task3

the queue will contain Task1, Task2, Task3.
Task1 will run and 2,3 will wait until Task1 is finished. But when Task1 is finished, it insert Task2a between Task1 and Task2, so it then waits for Task2a and so on.

test/cypress/support/commands.js
```diff
@@ -707,8 +707,27 @@ Cypress.Commands.add('deleteContainersManual', {}, () => {
 });
 
 Cypress.Commands.add('deleteAllCollections', {}, () => {
+  const waitForEmptyCollection = (maxLoops) => {
+    if (maxLoops == 0) {
+      cy.log('Max loops reached while waiting for the empty collections.');
```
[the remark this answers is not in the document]
himdel: I'd keep this one because that's an error case, subsequent failures will make more sense with the line in.

But, we probably don't need to be so verbose about "'Collections are empty!' :)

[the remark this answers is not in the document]
MilanPospisil: Thats thing to disscuss, I personaly like comments, It easier to discover errors when you have those.

PR #2115, opened 2022-05-25
Namespaces - Ensure feature parity between list/detail screens

MilanPospisil:
https://issues.redhat.com/browse/AAH-1087


himdel:
Closing, we can do these either as part of the Page refactor, or the framework switch, but it doesn't make sense to do this the old way anymore.

PR #2117, opened 2022-05-25
[rbac] Group roles test

jerabekjiri:
Issue: [AAH-1616](https://issues.redhat.com/browse/AAH-1616)

This test includes:
- testing of assigning roles to groups
   - should add a new role to group
   - should test filtering of assigned roles
   - should test filtering of roles in "Add roles" wizard
   - should correctly select and preview roles
   - should be able to remove role from group
   - should not display deleted role in group detail
   - should show group empty state
~~- moving permissions to the `permissions.js` file for reusability~~
- adding `cy.commands`:  `createRole`, `addRoleToGroup`, `deleteRole` => using galaxykit now
- removed `removeAllPermissions` and `addAllPermissions`
- fixed `group_management.js` test
- removed `group_permissions.js` (`rbac.js` test does the job now)
- removed `permissions.js` file, which is not useful with new galaxykit commands.


test/cypress/support/commands.js
```diff
@@ -201,193 +201,76 @@ Cypress.Commands.add('createGroup', {}, (name) => {
...
-];
-
-Cypress.Commands.add('removeAllPermissions', {}, (groupName) => {
-  cy.removePermissions(groupName, allPerms);
-});
-
-Cypress.Commands.add('addAllPermissions', {}, (groupName) => {
-  cy.addPermissions(groupName, allPerms);
-});
+// Cypress.Commands.add('addPermissions', {}, (groupName, permissions) => {
```
himdel, line 210: can you remove the commented out code please?

Otherwise, I think this looks good :)

test/cypress/integration/group_roles.js
```diff
@@ -1,54 +1,172 @@
...
-    cy.get('input[id="role_description"]').type('test description');
-    cy.contains('Collection Namespaces')
-      .parent()
-      .find('input', 'Select Permissions')
-      .click();
-    cy.get('ul[role="listbox"] > li:first').click();
-    cy.get('button').contains('Save').click();
+  before(() => {
+    cy.login();
+    cy.createGroup(groupName);
```
brumik, line 46: Why are we using the UI for this? Seemingly galaxykit should be able to do this.

On that same note it seems like galaxykit could use a `role create` command too, Maybe we should create an issue for it in the galaxykit repo?

brumik (commented):
I just have one small comment, otherwise looking good :+1:

test/cypress/integration/group_roles.js
jerabekjiri: I agree that this should be doing galaxykit, but I feel like it takes quite a lot of time to process and merge to galaxykit. So I wanted to do it manual way in here, but it definitely would be better to use galaxykit in future

>Maybe we should create an issue for it in the galaxykit repo?

yes :+1:

jerabekjiri: I see now, `createGroup` isn't using galaxykit, but is manual. So I might switch to galaxykit for that in here.

And created issues for managing roles in galaxykit cli:
- [AAH-1691](https://issues.redhat.com/browse/AAH-1691)
- [AAH-1692](https://issues.redhat.com/browse/AAH-1692 
)

jerabekjiri:
Fixed double `data-cy` conflict and merging.

PR #2135, opened 2022-05-30
standalone config - change pulp api path to '/api/automation-hub/pulp/api/v3/' (dev) or '/api/galaxy/pulp/api/v3/' (prod)

himdel:
https://github.com/ansible/galaxy_ng/pull/1247 was merged, so tasks won't work in dev ui without this

Cc @awcrosby .. do you know what happens for standalone prod? Should we change the pulp apis to `/api/galaxy/pulp/api/v3/` or do they stay in the original place?

### DIFF
--- a/config/standalone.dev.webpack.config.js
+++ b/config/standalone.dev.webpack.config.js
@@ -12,7 +12,7 @@ module.exports = webpackBase({
 
   // Path to the API on the API host. EX: /api/automation-hub
   API_BASE_PATH: apiBasePath,
-  PULP_API_BASE_PATH: '/pulp/api/v3/',
+  PULP_API_BASE_PATH: apiBasePath + 'pulp/api/v3/',
 
   // Path on the host where the UI is found. EX: /apps/automation-hub
   UI_BASE_PATH: '/ui/',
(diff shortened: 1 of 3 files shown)


himdel:
Before second commit (and also everywhere on master now):

|  ✖  6 of 35 failed (17%) | 150      134        7        1        8 |
|-|-|
| ✖  execution_environments_edit.js | 2        -        2        -        - |
| ✖  execution_environments_use_in_controller.js | 5        -        1        -        4 |
| ✖  imports_filter.js | 10        5        1        -        4 |
| ✖  task_list.js | 1        -        1        -        - |
| ✖  task_management_detail.js | 1        -        1        -        - |
| ✖  task_status.js | 1        -        1        -        - |

After second commit:

| ✖  1 of 35 failed (3%) | 150      148        1        1        -  |
|-|-|
| ✖  task_status.js | 1        -        1        -        - |

.. yeah, I think we can definitely say that prod needs `/api/galaxy/pulp/api/v3/`, and that https://github.com/ansible/galaxy_ng/pull/1247 (AAH-1544) is causing those new cypress failures :)

This may also cause issues with downstream builds though, but that's for a separate MR.

himdel:
Merging to fix tests on master

awcrosby:
@himdel the pulp api moved for all deployment modes, so for standlone  the UI will need to access it in it's new location. We also plan to add a redirect for standalone.

PR #2136, opened 2022-05-31
l10n: initialize moment locale with the rest of l10n init, don't do it on every render

himdel:
No visible change, just moving moment.js locale init together with the rest of locale init :).


jerabekjiri (approved):
LGTM :+1:  

should be backported to 4.4.0?

himdel:
Makes sense :+1:

PR #2164, opened 2022-06-04
test fix - collection_upload "collection should be uploaded"

himdel:
fails on master now,

    Collection Upload Tests
    ✓ should not upload new collection version in collection list when user does not have permissions (24119ms)
    ✓ should not upload new collection version in collection detail when user does not have permissions
    ✓ should see upload new collection version in collection list when user does have permissions
    ✓ should see upload new collection version in collection detail when user does have permissions
    ✓ user should not be able to upload new collection without permissions
    1) collection should be uploaded
    ✓ should not upload new collection version when user does not have permissions
    ✓ should deprecate let user deprecate and undeprecate collections (5678ms)

    7 passing (1m)
    1 failing

    1) Collection Upload Tests
         collection should be uploaded:
       AssertionError: Timed out retrying after 4000ms: Expected to find content: 'Completed' within the element: <span.pf-c-label__content> but never did.
        at Context.eval (http://localhost:8002/__cypress/tests?p=cypress/integration/collection_upload.js:183:58)

.. after first commit it became

    AssertionError: Timed out retrying after 4000ms: Expected not to find content: 'Failed' within the element: <span.pf-c-label__content> but continuously found it.

we're using `ansible-network` as a fixture for upload, but `ansible-network` is using `requires_ansible: '*'`,
which no longer passes validation since galaxy-importer 0.4.5 (https://github.com/ansible/galaxy_ng/pull/1280, AAH-981)

so it fails the test by failing to import
=> replacing with `ansible-posix`


ZitaNemeckova (approved):

himdel:
Backporting to 4.5 via #2498

PR #2185, opened 2022-06-07
Collections - Ensure feature parity between list detail screens - Deletion (AAH 1025)

MilanPospisil:
https://issues.redhat.com/browse/AAH-1025

Continue work of: https://github.com/ansible/ansible-hub-ui/pull/1766

It adds deletion operation to the collection list and namespace detail collection list.

### DIFF
--- a/src/components/collection-list/collection-list.tsx
+++ b/src/components/collection-list/collection-list.tsx
@@ -1,17 +1,20 @@
-import { t } from '@lingui/macro';
+import { t, Trans } from '@lingui/macro';
 import * as React from 'react';
 import './list.scss';
 
 import { Button, DropdownItem, DataList } from '@patternfly/react-core';
-
 import { CollectionListType } from 'src/api';
+
 import {
   CollectionListItem,
   Pagination,
   StatefulDropdown,
   EmptyStateFilter,
+  DeleteCollectionModal,
 } from 'src/components';
 import { ParamHelper } from 'src/utilities/param-helper';
+
+import { deleteCollectionUtils } from 'src/utilities';
 
 interface IProps {
   collections: CollectionListType[];
@@ -28,10 +31,28 @@ interface IProps {
   showControls?: boolean;
   handleControlClick?: (id, event) => void;
   repo?: string;
+  addAlert: (alert) => void;
+  reload: () => void;
+  context: any;
+}
+
+interface IState {
+  deleteCollection: CollectionListType;
+  confirmDelete: boolean;
+  isDeletionPending: boolean;
 }
 
 // only used in namespace detail, collections uses individual items
-export class CollectionList extends React.Component<IProps> {
+export class CollectionList extends React.Component<IProps, IState> {
+  constructor(props) {
+    super(props);
+    this.state = {
+      deleteCollection: null,
+      confirmDelete: false,
+      isDeletionPending: false,
+    };
+  }
+
   render() {
     const {
       collections,
@@ -43,8 +64,30 @@ export class CollectionList extends React.Component<IProps> {
       repo,
     } = this.props;
 
+    const { deleteCollection, isDeletionPending, confirmDelete } = this.state;
+
     return (
       <React.Fragment>
+        <DeleteCollectionModal
+          deleteCollection={deleteCollection}
+          isDeletionPending={isDeletionPending}
+          confirmDelete={confirmDelete}
+          collectionVersion={null}
+          cancelAction={() => this.setState({ deleteCollection: null })}
+          deleteAction={() =>
+            this.setState({ isDeletionPending: true }, () => {
+              deleteCollectionUtils.deleteCollection(
+                this,
+                false,
+                this.props.context.selectedRepo,
+                (alert) => {
+                  this.props.addAlert(alert);
+                },
+              );
+            })
+          }
+          onChange={(val) => this.setState({ confirmDelete: val })}
+        ></DeleteCollectionModal>
         <DataList aria-label={t`List of Collections`}>
           {collections.length > 0 ? (
             collections.map((c) => (
@@ -89,6 +132,13 @@ export class CollectionList extends React.Component<IProps> {
         </Button>
         <StatefulDropdown
           items={[
+            <React.Fragment key='fragment'>
+              {deleteCollectionUtils.deleteMenuOption(
+                true,
+                this.props.context,
+                () => this.tryOpenDeleteModalWithConfirm(collection),
+              )}
+            </React.Fragment>,
             <DropdownItem
               onClick={() =>
                 this.props.handleControlClick(collection.id, 'deprecate')
@@ -102,5 +152,38 @@ export class CollectionList extends React.Component<IProps> {
         />
       </div>
     );
+  }
+
+  private tryOpenDeleteModalWithConfirm(collection) {
+    deleteCollectionUtils.getUsedbyDependencies(
+      collection,
+      (noDependencies) =>
+        this.openDeleteModalWithConfirm(noDependencies, collection),
+      (alerts) => this.props.addAlert(alerts),
+    );
+  }
+
+  private openDeleteModalWithConfirm(noDependencies, collection) {
+    if (noDependencies) {
+      this.setState({
+        deleteCollection: collection,
+        confirmDelete: false,
+      });
+    } else {
+      this.props.addAlert({
+        title: (
+          <Trans>
+            Cannot delete until collections <br />
+            that depend on this collection <br />
+            have been deleted.
+          </Trans>
+        ),
+        variant: 'warning',
+      });
+    }
+  }
+
+  public load() {
+    this.props.reload();
   }
 }
--- a/src/components/collection-list/collection-list.tsx
+++ b/src/components/collection-list/collection-list.tsx
@@ -1,20 +1,16 @@
-import { t, Trans } from '@lingui/macro';
+import { t } from '@lingui/macro';
 import * as React from 'react';
 import './list.scss';
 
-import { Button, DropdownItem, DataList } from '@patternfly/react-core';
+import { DataList } from '@patternfly/react-core';
 import { CollectionListType } from 'src/api';
 
 import {
   CollectionListItem,
   Pagination,
-  StatefulDropdown,
   EmptyStateFilter,
-  DeleteCollectionModal,
 } from 'src/components';
 import { ParamHelper } from 'src/utilities/param-helper';
-
-import { deleteCollectionUtils } from 'src/utilities';
 
 interface IProps {
   collections: CollectionListType[];
@@ -31,26 +27,13 @@ interface IProps {
   showControls?: boolean;
   handleControlClick?: (id, event) => void;
   repo?: string;
-  addAlert: (alert) => void;
-  reload: () => void;
-  context: any;
-}
-
-interface IState {
-  deleteCollection: CollectionListType;
-  confirmDelete: boolean;
-  isDeletionPending: boolean;
+  renderCollectionControls: (collection) => any;
 }
 
 // only used in namespace detail, collections uses individual items
-export class CollectionList extends React.Component<IProps, IState> {
+export class CollectionList extends React.Component<IProps> {
   constructor(props) {
     super(props);
-    this.state = {
-      deleteCollection: null,
-      confirmDelete: false,
-      isDeletionPending: false,
-    };
   }
 
   render() {
@@ -64,36 +47,14 @@ export class CollectionList extends React.Component<IProps, IState> {
       repo,
     } = this.props;
 
-    const { deleteCollection, isDeletionPending, confirmDelete } = this.state;
-
     return (
       <React.Fragment>
-        <DeleteCollectionModal
-          deleteCollection={deleteCollection}
-          isDeletionPending={isDeletionPending}
-          confirmDelete={confirmDelete}
-          collectionVersion={null}
-          cancelAction={() => this.setState({ deleteCollection: null })}
-          deleteAction={() =>
-            this.setState({ isDeletionPending: true }, () => {
-              deleteCollectionUtils.deleteCollection(
-                this,
-                false,
-                this.props.context.selectedRepo,
-                (alert) => {
-                  this.props.addAlert(alert);
-                },
-              );
-            })
-          }
-          onChange={(val) => this.setState({ confirmDelete: val })}
-        ></DeleteCollectionModal>
         <DataList aria-label={t`List of Collections`}>
           {collections.length > 0 ? (
             collections.map((c) => (
               <CollectionListItem
                 controls={
-                  showControls ? this.renderCollectionControls(c) : null
+                  showControls ? this.props.renderCollectionControls(c) : null
                 }
                 key={c.id}
                 {...c}
@@ -119,71 +80,5 @@ export class CollectionList extends React.Component<IProps, IState> {
         />
       </React.Fragment>
     );
-  }
-
-  private renderCollectionControls(collection: CollectionListType) {
-    return (
-      <div style={{ display: 'flex', alignItems: 'center' }}>
-        <Button
-          onClick={() => this.props.handleControlClick(collection.id, 'upload')}
-          variant='secondary'
-        >
-          {t`Upload new version`}
-        </Button>
-        <StatefulDropdown
-          items={[
-            <React.Fragment key='fragment'>
-              {deleteCollectionUtils.deleteMenuOption(
-                true,
-                this.props.context,
-                () => this.tryOpenDeleteModalWithConfirm(collection),
-              )}
-            </React.Fragment>,
-            <DropdownItem
-              onClick={() =>
-                this.props.handleControlClick(collection.id, 'deprecate')
-              }
-              key='deprecate'
-            >
-              {collection.deprecated ? t`Undeprecate` : t`Deprecate`}
-            </DropdownItem>,
-          ]}
-          ariaLabel='collection-kebab'
-        />
-      </div>
-    );
-  }
-
-  private tryOpenDeleteModalWithConfirm(collection) {
-    deleteCollectionUtils.getUsedbyDependencies(
-      collection,
-      (noDependencies) =>
-        this.openDeleteModalWithConfirm(noDependencies, collection),
-      (alerts) => this.props.addAlert(alerts),
-    );
-  }
-
-  private openDeleteModalWithConfirm(noDependencies, collection) {
-    if (noDependencies) {
-      this.setState({
-        deleteCollection: collection,
-        confirmDelete: false,
-      });
-    } else {
-      this.props.addAlert({
-        title: (
-          <Trans>
-            Cannot delete until collections <br />
-            that depend on this collection <br />
-            have been deleted.
-          </Trans>
-        ),
-        variant: 'warning',
-      });
-    }
-  }
-
-  public load() {
-    this.props.reload();
   }
 }
--- a/src/components/collection-list/collection-list.tsx
+++ b/src/components/collection-list/collection-list.tsx
@@ -22,10 +22,7 @@ interface IProps {
   updateParams: (params) => void;
   itemCount: number;
   ignoredParams: string[];
-
-  showNamespace?: boolean;
   showControls?: boolean;
-  handleControlClick?: (id, event) => void;
   repo?: string;
   renderCollectionControls: (collection) => React.ReactNode;
 }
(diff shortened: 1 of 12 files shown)


src/components/headers/collection-header.tsx
```diff
@@ -367,62 +344,26 @@ export class CollectionHeader extends React.Component<IProps, IState> {
...
-            deleteAction={() =>
-              this.setState({ isDeletionPending: true }, () => {
-                collectionVersion
-                  ? this.deleteCollectionVersion(collectionVersion)
-                  : this.deleteCollection();
-              })
-            }
-            isDisabled={!confirmDelete || isDeletionPending}
-            title={
+        {deleteCollectionUtils.deleteModal(
```
himdel, line 347: sounds like this should be a component?

src/containers/search/search.tsx
```diff
@@ -131,6 +150,25 @@ class Search extends React.Component<RouteComponentProps, IState> {
           alerts={this.state.alerts}
           closeAlert={(i) => this.closeAlert(i)}
         />
+        {deleteCollectionUtils.deleteModal(
+          deleteCollection,
+          isDeletionPending,
+          confirmDelete,
+          collectionVersion,
```
himdel, line 157: collection list screens should not really be dealing with collection *versions* at all

src/containers/search/search.tsx
```diff
@@ -364,6 +407,37 @@ class Search extends React.Component<RouteComponentProps, IState> {
...
+
+  private openDeleteModalWithConfirm(noDependencies, collection) {
+    if (noDependencies) {
+      this.setState({
+        deleteCollection: collection,
+        confirmDelete: false,
+      });
+    } else {
+      const newAlerts = this.state.alerts;
+      const newAlert = new AlertType();
```
himdel, line 427: (whatever you did elsewhere about these already :))

src/components/collection-list/collection-list.tsx
```diff
@@ -29,10 +31,30 @@ interface IProps {
...
+  isDeletionPending: boolean;
+  collectionVersion: string | null;
 }
 
 // only used in namespace detail, collections uses individual items
-export class CollectionList extends React.Component<IProps> {
+export class CollectionList extends React.Component<IProps, IState> {
+  constructor(props) {
+    super(props);
+    this.state = {
```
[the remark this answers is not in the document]
MilanPospisil: Yes, I agree I may have to rethink this, because this codes were written before we agree to use this.state in utility functions.

src/containers/search/search.tsx
MilanPospisil: Yes, this was copy pasted from other code and collectionVersion should be null here anyway :-)

src/containers/search/search.tsx
MilanPospisil: Leftover from old codes before our review of previous PR, I will fix it there too.

src/components/headers/collection-header.tsx
MilanPospisil: Maybe, I will try to refactor this.

src/containers/namespace-detail/namespace-detail.tsx
```diff
@@ -314,18 +312,25 @@ export class NamespaceDetail extends React.Component<IProps, IState> {
...
+                      this.handleCollectionAction(id, action)
+                    }
+                    repo={this.context.selectedRepo}
+                    addAlert={(alert) =>
+                      this.setState({ alerts: [...alerts, alert] })
+                    }
+                    reload={() => this.loadAll()}
+                    context={this.context}
+                  />
+                }
```
himdel, line 333:
```suggestion
                <CollectionList
                  updateParams={updateParams}
                  params={params}
                  ignoredParams={ignoredParams}
                  collections={collections}
                  itemCount={itemCount}
                  showControls={this.state.showControls}
                  handleControlClick={(id, action) =>
                    this.handleCollectionAction(id, action)
                  }
                  repo={this.context.selectedRepo}
                  addAlert={(alert) =>
                    this.setState({ alerts: [...alerts, alert] })
                  }
                  reload={() => this.loadAll()}
                  context={this.context}
                />
```

(just remoes the extra `{}` around a component)

src/components/delete-modal/delete-collection-modal.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { DeleteModal } from 'src/components';
+
+interface IProps {
+  deleteCollection: any;
+  isDeletionPending: boolean;
+  confirmDelete: boolean;
+  collectionVersion: string | null;
+  cancelAction: () => void;
+  deleteAction: () => void;
+  onChange: (val: any) => void;
```
himdel, line 13:
```suggestion
  onChange: (val) => void;
```

src/components/collection-list/collection-list.tsx
```diff
@@ -89,6 +132,13 @@ export class CollectionList extends React.Component<IProps> {
         </Button>
         <StatefulDropdown
           items={[
+            <React.Fragment key='fragment'>
+              {deleteCollectionUtils.deleteMenuOption(
```
himdel, line 136: and here (https://github.com/ansible/ansible-hub-ui/pull/2185/files#r899066360)

src/components/delete-modal/delete-collection-modal.tsx
```diff
@@ -0,0 +1,82 @@
+import React from 'react';
+import { t, Trans } from '@lingui/macro';
+import { Text, Checkbox } from '@patternfly/react-core';
+import { DeleteModal } from 'src/components';
+
+interface IProps {
+  deleteCollection: any;
+  isDeletionPending: boolean;
+  confirmDelete: boolean;
+  collectionVersion: string | null;
```
himdel, line 10:
```suggestion
  collectionVersion?: string;
```

himdel:
(also the namespace-detail test waiting for DEPRECATED to appear seems to be failing, this sounds possibly related)

src/components/collection-list/collection-list.tsx
```diff
@@ -77,30 +81,4 @@ export class CollectionList extends React.Component<IProps> {
       </React.Fragment>
     );
   }
-
-  private renderCollectionControls(collection: CollectionListType) {
```
himdel, line 81: why? This is not depending on any state

src/components/collection-list/collection-list.tsx
```diff
@@ -28,10 +27,15 @@ interface IProps {
...
   repo?: string;
+  renderCollectionControls: (collection) => any;
 }
 
 // only used in namespace detail, collections uses individual items
 export class CollectionList extends React.Component<IProps> {
+  constructor(props) {
+    super(props);
+  }
+
```
himdel, line 38:
```suggestion
```

src/components/collection-list/collection-list.tsx
```diff
@@ -28,10 +27,15 @@ interface IProps {
   showControls?: boolean;
   handleControlClick?: (id, event) => void;
   repo?: string;
+  renderCollectionControls: (collection) => any;
```
himdel, line 30: (any)

src/components/delete-modal/delete-collection-modal.tsx
```diff
@@ -0,0 +1,83 @@
...
+            collectionVersion
+              ? t`Delete collection version?`
+              : t`Delete collection?`
+          }
+        >
+          <>
+            <Text style={{ paddingBottom: 'var(--pf-global--spacer--md)' }}>
+              {collectionVersion ? (
+                <>
+                  {deleteCollection['all_versions'].length === 1 ? (
```
himdel, line 46:
```suggestion
                  {deleteCollection.all_versions.length === 1 ? (
```

src/components/collection-list/collection-list.tsx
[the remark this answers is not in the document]
himdel: I can't tell what you did since it's not a separate commit, sounds like you moved more then I was asking for, I was only talking about the state vars, but we might as well, containers are the place for most of that logic anyway :).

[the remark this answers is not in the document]
MilanPospisil: Its better like now, much less props are now required and List only calls render menu using props, nothing else.

src/components/collection-list/collection-list.tsx
MilanPospisil: Because of several reasons. First, I wanted as much as possible in one place - the detail container. Having logic scattered between two components was bit confusing. Second, the renderCollectionControls calls several methods and is using this, those methods than use setStates, so if any of these methods will be in List, many props will be required. Now they can directly use state of the container.

src/components/delete-modal/delete-collection-modal.tsx
MilanPospisil: Because the type is defined as CollectionDetailType | CollectionListType and all_versions is only in one of those = typing throws error, so this was solution around, but not that good. Do you have better idea?

himdel: Can't you just use the same logic as in https://github.com/ansible/ansible-hub-ui/pull/2185#discussion_r899057493 ?

The `collectionVersion` condition already tells us the type, even if typescript can't quite figure it out by itself, so you can then use `(deleteCollection as CollectionDetailType).all_versions`

But yeah, up to you, both are fine :)

src/components/collection-list/collection-list.tsx
```diff
@@ -28,6 +27,7 @@ interface IProps {
   showControls?: boolean;
   handleControlClick?: (id, event) => void;
   repo?: string;
+  renderCollectionControls: (collection) => React.ReactElement;
```
himdel, line 30: ah, better, thanks :)

Though you may want to prefer ReactNode over ReactElement...

https://stackoverflow.com/a/58123882/535728
> A ReactElement is an object with a type and props.
> A ReactNode is a ReactElement, a ReactFragment, a string, a number or an array of ReactNodes, or null, or undefined, or a boolean

```suggestion
  renderCollectionControls: (collection) => React.ReactNode;
```

src/utilities/delete-collection.tsx
```diff
@@ -0,0 +1,191 @@
...
+            <DropdownItem isDisabled>
+              {t`Delete entire collection`}
+            </DropdownItem>
+          </Tooltip>
+        );
+
+    if (!ret) {
+      ret = <div></div>;
+    }
+    return ret;
```
himdel, line 65: can we simplify the logic by returning early?

```suggestion
    if (!delete_collection_permission) {
      return <div></div>;
    }
    
    return noDependencies ? (
          <DropdownItem
            key='delete-collection-enabled'
            onClick={() => onClick()}
            data-cy='delete-collection-dropdown'
          >
            {t`Delete entire collection`}
          </DropdownItem>
        ) : (
          <Tooltip
            key='delete-collection-disabled'
            position='left'
            content={
              <Trans>
                Cannot delete until collections <br />
                that depend on this collection <br />
                have been deleted.
              </Trans>
            }
          >
            <DropdownItem isDisabled>
              {t`Delete entire collection`}
            </DropdownItem>
          </Tooltip>
        );
```

himdel: also .. wasn't `<div></div>` problematic inside menu options? Should this return null instead?

(adding a padding-sized item)

src/containers/search/search.tsx
```diff
@@ -358,6 +398,40 @@ class Search extends React.Component<RouteComponentProps, IState> {
...
+  private tryOpenDeleteModalWithConfirm(collection) {
+    deleteCollectionUtils.getUsedbyDependencies(
+      collection,
+      (noDependencies) =>
+        this.openDeleteModalWithConfirm(noDependencies, collection),
+      (alerts) => this.setState({ alerts: [...this.state.alerts, alerts] }),
+    );
+  }
+
+  private openDeleteModalWithConfirm(noDependencies, collection) {
```
himdel, line 410: this is identical to the `openDeleteModalWithConfirm` in `namespace-detail` .. should we move them to utils as well?

src/components/collection-list/collection-list.tsx
```diff
@@ -28,6 +27,7 @@ interface IProps {
   showControls?: boolean;
   handleControlClick?: (id, event) => void;
```
himdel, line 28: dead prop now?

src/utilities/delete-collection.tsx
```diff
@@ -0,0 +1,191 @@
...
+          component.setState({
+            collectionVersion: null,
+            deleteCollection: null,
+            isDeletionPending: false,
+          });
+
+          if (redirect) {
+            component.setState({
+              redirect: formatPath(Paths.namespaceByRepo, {
+                repo: component.context.selectedRepo,
```
himdel, line 163:
```suggestion
                repo: selectedRepo,
```

src/containers/namespace-detail/namespace-detail.tsx
```diff
@@ -224,6 +231,24 @@ export class NamespaceDetail extends React.Component<IProps, IState> {
...
+        <DeleteCollectionModal
+          deleteCollection={deleteCollection}
+          isDeletionPending={isDeletionPending}
+          confirmDelete={confirmDelete}
+          collectionVersion={null}
+          cancelAction={() => this.setState({ deleteCollection: null })}
+          deleteAction={() =>
+            this.setState({ isDeletionPending: true }, () => {
+              deleteCollectionUtils.deleteCollection(
+                this,
```
himdel, line 243: can we be a bit more specific instead of passing `this`?

```suggestion
                {
                  state: this.state,
                  setState: this.setState,
                  load: this.load,
                },
```

src/containers/search/search.tsx
```diff
@@ -131,6 +148,24 @@ class Search extends React.Component<RouteComponentProps, IState> {
...
         />
+        <DeleteCollectionModal
+          deleteCollection={deleteCollection}
+          isDeletionPending={isDeletionPending}
+          confirmDelete={confirmDelete}
+          collectionVersion={null}
+          cancelAction={() => this.setState({ deleteCollection: null })}
+          deleteAction={() =>
+            deleteCollectionUtils.deleteCollection(
+              this,
```
himdel, line 159: and here (https://github.com/ansible/ansible-hub-ui/pull/2185/files#r906736990)

src/components/headers/collection-header.tsx
```diff
@@ -360,63 +340,27 @@ export class CollectionHeader extends React.Component<IProps, IState> {
...
-                onChange={(val) => this.setState({ confirmDelete: val })}
-                label={t`I understand that this action cannot be undone.`}
-                id='delete_confirm'
-              />
-            </>
-          </DeleteModal>
-        )}
+                ? this.deleteCollectionVersion(collectionVersion)
+                : deleteCollectionUtils.deleteCollection(
+                    this,
```
himdel, line 354: and here (https://github.com/ansible/ansible-hub-ui/pull/2185/files#r906736990)

src/utilities/delete-collection.tsx
```diff
@@ -0,0 +1,191 @@
...
+              label={t`I understand that this action cannot be undone.`}
+              id='delete_confirm'
+            />
+          </>
+        </DeleteModal>
+      )
+    );
+  }
+
+  public deleteCollection(component, redirect, selectedRepo, addAlert) {
```
himdel, line 132: might want to make this 

```suggestion
  public deleteCollection({ state, setState, load }, redirect, selectedRepo, addAlert) {
```

src/containers/namespace-detail/namespace-detail.tsx
```diff
@@ -734,6 +759,74 @@ export class NamespaceDetail extends React.Component<IProps, IState> {
...
+            ),
+            <DropdownItem
+              onClick={() =>
+                this.handleCollectionAction(collection.id, 'deprecate')
+              }
+              key='deprecate'
+            >
+              {collection.deprecated ? t`Undeprecate` : t`Deprecate`}
+            </DropdownItem>,
+          ]}
```
himdel, line 824:
```suggestion
          ].filter(Boolean)}
```

src/containers/search/search.tsx
```diff
@@ -358,6 +398,40 @@ class Search extends React.Component<RouteComponentProps, IState> {
     );
```
himdel, line 398: ^ `menuItems.filter(Boolean)`

(https://github.com/ansible/ansible-hub-ui/pull/2185#discussion_r899066360)

src/utilities/delete-collection.tsx
MilanPospisil: The null threw the errors and it works with div.

src/containers/search/search.tsx
MilanPospisil: This was done before we agreed to access the state of the component in utilities function. Before that, I used props and there was no reason to move it to utilities and cover that small code with lots of props. Now it makes sense :-)

MilanPospisil: I did it, but this (along with another menu operations) needs another PR for refactor, because with access to state of the component in utils, other things can be refactored as well.

src/components/collection-list/collection-list.tsx
MilanPospisil: I wonder why the compiler does not marked them as unused?

src/containers/namespace-detail/namespace-detail.tsx
MilanPospisil: But this will cause to redo everything else Im doing now in this and Namespace PR (https://github.com/ansible/ansible-hub-ui/pull/2115). The code can remain the same, but I need to replace all this references by the object and this will be no much different than props. I suggest we disscuss this change, maybe in thursday in office, if we will meet there.

himdel: how would it cause you to redo everything? You're still passing the same methods, just being explicit about which they are.

Passing whole class component instances to other functions is not maintainable, we wouldn't know which functions are being used, we couldn't rename/transform stuff when needed, and there's no `this` in functional components anyway.

Passing `state`, `setState()` and `reload()` seems like a reasonable interface that could be usable from any stateful component (though eventually we'll need to use types to validate the passed in state for compatibility).

src/components/collection-list/collection-list.tsx
himdel: it can do that for functions, exports, but not methods, `this` is too magic for that :)

src/utilities/delete-collection.tsx
himdel: Yeah, because you were missing those `filter(Boolean)` bits before.

Again, we can't use empty divs here, it has to be falsy

src/containers/namespace-detail/namespace-detail.tsx
MilanPospisil: State and setState seems uselless to me, because every class component has state and setState and it really is not checking anything - we will access further variables of state and setState without type checking. In functional component, it makes sense, because it will pass its own state and setState for every variable, so type checking will do its work. The "load: this.load" can work, but it will cause problems because this will not work anymore in the function, I would have to use lambda and the whole code will become very talkative.  I now really understand why they introduced functional components :-) So what now? Lets pass every state variable and functions as lambda call and use only setState as "magic" dirty access to class state in helper functions? Or pass everything and not use direct state access in helpers at all? I agree that we need mantainability, even if that meens talkative code.

himdel: Well, this way, all the functions can accept the same interface (state, setState, reload), but do their own type-checking by imposing type constrains on the state param.

That or we can try passing individual getter/setter pairs, but that does seem too talkative indeed, that's what interfaces are for.

(And yeah, you may need `load: () => this.load()` (and similar for setState) for this to work.)

src/utilities/delete-collection.tsx
MilanPospisil: I only slightly changed this to pass deleteCollection(state, setState, load, redirect...) without braces {}. If you insist on braces, tell me please why do you want that.

himdel: This is fine :) the braces were there mostly for if we decided to only be explicit here, so we could still pass this or a object of that shape. Not needed when passing like that, so that's a detail now :)

Eventually we'll want an object instead of N params, to be able to add new (and named) fields, but that's a job for future us :)

himdel:
Tested, looks like everything still works :+1: 

I also noticed an unused `deleteModal` method, looks like this was moved to `DeleteConfirmModal`, removing the original method,
`getUsedbyDependencies` had a `setDependencies` callback that was opening a modal .. changed to return a Promise instead,
replaced `state, setState` with a bit more explicit `addAlert, setState` for the open modal function,
and updated to use named params since I kept getting lost in which param is which :)

Let me know if this works for you :)

himdel (approved):

PR #2209, opened 2022-06-13
Only show `add ee` button with user permissions

ShaiahWren:
Issue: AAH-1696

Relevant screens: EE list view with data, EE list view empty
Show the `Add EE` button _only_ when `this.context.user?.model_permissions?.add_containernamespace` is set.

Notes: In the empty list view, I removed the spacing div between where the add button used to be and the `pushImagesButton`. As such, the existence of the div is connected to the variable `addRemoteButton`. I'm open to other ideas for adjusting for the funky spacing.

On the empty EE list view I am also noticing that, with or without the add button, it has language `You currently have no container repositories. Add a container repository via the CLI to get started.` -- implying that the user **is** able to add an EE regardless. Shall we change the language here for non-superusers?


himdel (approved):

src/containers/execution-environment-list/execution_environment_list.tsx
```diff
@@ -142,18 +142,22 @@ class ExecutionEnvironmentList extends React.Component<
...
+            this.setState({
+              showRemoteModal: true,
+              itemToEdit: {} as ExecutionEnvironmentType,
+            })
+          }
+          variant='primary'
+        >
+          <Trans>Add execution environment</Trans>
+        </Button>
+        <div>&nbsp;</div>
```
himdel, line 159: this breaks the toolbar vertical alignment when there are data (addRemoteButton is also used in a ToolbarItem)

fixing in https://github.com/ansible/ansible-hub-ui/pull/2238

PR #2211, opened 2022-06-14
Update collection signing feature flags

himdel:
Fixes: AAH-1710
Depends on: (AAH-1690 PR)

### DIFF
--- a/CHANGES/1710.misc
+++ b/CHANGES/1710.misc
@@ -1,0 +1,1 @@
+Update collection signing feature flags
--- a/src/api/response-types/collection.ts
+++ b/src/api/response-types/collection.ts
@@ -1,3 +1,5 @@
+type SignState = 'signed' | 'unsigned' | 'partial';
+
 export class CollectionUploadType {
   id: string;
   file: File;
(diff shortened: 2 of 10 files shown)


brumik (commented):
Cannot we ommit the `signature_enabled` from most of the checks? They depend on it in any case (even the defaults take it into consideration). I see the added value (ui less probable to crash) but if the user sets for example the `display_signatures == True` and nothing happens is just as much a "failure" as the ui crashing saying something about API failure and realizing that he forgot to set the `signatures_enabled` setting

himdel:
Agreed, looks like with `PULP_GALAXY_COLLECTION_SIGNING_SERVICE` unset, `PULP_GALAXY_AUTO_SIGN_COLLECTIONS=true` doesn't make it into feature flags, so sounds it's a safe thing to do :+1: 

Updated to remove `signature_enabled` except for one case where no other feature flag is used.

Also updated to give `require_upload_signatures` priority over `collection_auto_sign`,
and found one more place trying to show signature witthout checking `display_signatures` :).

brumik (approved):
Nice improvement! LGTM

PR #2212, opened 2022-06-15
Add loading page to all pages

ShaiahWren:
Issue: AAH-518

This pr makes sure each UI page has proper page loaders - either `LoadingPageSpinner`  - most often used for components like tables, or `LoadingPageWithHeader` , with is good for loading whole pages. Most pages use either one or the other.

I have compiled a list of current and new additions in [here](https://docs.google.com/document/d/1cD2njsQJJ6JyZDzQLARcaPzGSHRUoJUMtrx-yvoH0a8/edit).

### DIFF
--- a/src/containers/token/token-standalone.tsx
+++ b/src/containers/token/token-standalone.tsx
@@ -15,6 +15,7 @@ import {
   AlertType,
   closeAlertMixin,
   LoadingPageWithHeader,
+  LoadingPageSpinner,
 } from 'src/components';
 import { ActiveUserAPI } from 'src/api';
 import { AppContext } from 'src/loaders/app-context';
@@ -35,7 +36,7 @@ class TokenPage extends React.Component<RouteComponentProps, IState> {
       token: undefined,
       alerts: [],
       loading: true,
-      loadingToken: true,
+      loadingToken: false,
     };
   }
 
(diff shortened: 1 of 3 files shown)


src/containers/token/token-standalone.tsx
```diff
@@ -31,14 +34,25 @@ class TokenPage extends React.Component<RouteComponentProps, IState> {
...
+  }
+
   render() {
-    const { token, alerts } = this.state;
+    const { token, alerts, loading } = this.state;
     const unauthorised = !this.context.user || this.context.user.is_anonymous;
     const expiration = this.context.settings.GALAXY_TOKEN_EXPIRATION;
     const expirationDate = new Date(Date.now() + 1000 * 60 * expiration);
 
+    if (loading) {
```
himdel, line 52: should this be `loading || loadingToken`?

or should `loadingToken` be used elsewhere? it's currently write-only

src/containers/token/token-standalone.tsx
```diff
@@ -31,14 +35,25 @@ class TokenPage extends React.Component<RouteComponentProps, IState> {
     this.state = {
       token: undefined,
       alerts: [],
+      loading: true,
+      loadingToken: false,
     };
   }
+
+  componentDidMount() {
+    this.setState({ loading: false });
```
himdel, line 44: oh, one more thing I didn't notice .. if componentDidMount sets loading: false, loading is not really used either .. there would be no render before it mounted

so I guess we can remove the whole `loading` part and only keep `loadingToken`?

himdel (approved):

PR #2230, opened 2022-06-22
[rbac] Test Namespace/EE owners

himdel:
Depends on https://github.com/ansible/ansible-hub-ui/pull/2092, will fail until then

Tests Namespace owners tab and EE Owners tab:
 * empty state,
 * add group and role,
 * open group, add role,
 * expand role,
 * remove role,
 * breadcrumbs back,
 * remove group

Also fixes a bug in EE tab detection - when a container is named "anything_owners_anything", the Owners tab becomes active by default, breaking the Detail tab.


ZitaNemeckova:
@himdel cy failure seems to be related

himdel:
Yup, this depends on #2092, will fail until then

PR #2228, opened 2022-06-20
Edit group permissions - correctly hide user/group-related permissions in keycloak mode

himdel:
Fixes AAH-1688

When in keycloak mode, editing group permissions should still work, but blows up...

```
core.development.js:435 Uncaught TypeError: Cannot read properties of undefined (reading 'values')
    at I18n._ (core.development.js:435:1)
    at permission-chip-selector.tsx:86:44
    at Array.map (<anonymous>)
    at PermissionChipSelector.render (permission-chip-selector.tsx:84:45)
    at finishClassComponent (react-dom.development.js:17160:1)
    at updateClassComponent (react-dom.development.js:17110:1)
    at beginWork (react-dom.development.js:18620:1)
    at HTMLUnknownElement.callCallback (react-dom.development.js:188:1)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:237:1)
    at invokeGuardedCallback (react-dom.development.js:292:1)
_ @ core.development.js:435

The above error occurred in the <PermissionChipSelector> component:
    in PermissionChipSelector (created by GroupDetail)
    in div (created by FlexItem)
    in FlexItem (created by GroupDetail)
    in div (created by Flex)
    in Flex (created by GroupDetail)
    in div (created by GroupDetail)
    in section (created by GroupDetail)
    in section (created by Main)
    in Main (created by GroupDetail)
    in GroupDetail (created by Context.Consumer)
    in withRouter(GroupDetail) (created by AuthHandler)
    in AuthHandler (created by Context.Consumer)
    in Route (created by Routes)
    in Switch (created by Routes)
    in Routes (created by App)
    in main (created by Page)
    in div (created by Page)
    in Page (created by App)
    in App (created by Context.Consumer)
    in withRouter(App)
    in I18nProvider
    in Router (created by BrowserRouter)
    in BrowserRouter
```

that's because we're correctly creating `filteredPermissions` but then using them only for mapping the values, instead of taking the original list from there.

The end result is the UI is trying to display these:

but failing to look up human strings for `'galaxy.delete_user', 'galaxy.add_user', 'galaxy.change_user', 'galaxy.delete_group', 'galaxy.add_group'` causing them to become `undefined`, and blowing up trying to translate that.

Rewrote the `availablePermissions` and `selectedPermissions` logic, now:

PR #2165, opened 2022-06-04
test fix - imports_filter "before each" hook for "exact filter for name is working."

himdel:
Imports filter test
     ✓ should display success info after importing collection (21515ms)
     ✓ should fail on importing existing collection (5266ms)
     ✓ should redirect to new uploaded collection
     ✓ should be able to switch between namespaces
     ✓ partial filter for name is working.
     1) "before each" hook for "exact filter for name is working."

    5 passing (44s)
    1 failing

    1) Imports filter test
       "before each" hook for "exact filter for name is working.":
       Error: The following error originated from your application code, not from Cypress. It was caused by an unhandled promise rejection.
       > Request failed with status code 403

looks like an earlier test is not waiting for all requests to finish first ... because it's using intercept with duplicate names and thus waiting for an earlier request :)

This is a sporadic failure, but happening often enough .. will try 5 consecutive runs:
https://github.com/ansible/ansible-hub-ui/runs/7055995886?check_suite_focus=true :heavy_check_mark: 
https://github.com/ansible/ansible-hub-ui/runs/7056145293?check_suite_focus=true :heavy_check_mark: 
https://github.com/ansible/ansible-hub-ui/runs/7056369214?check_suite_focus=true :heavy_check_mark: 
https://github.com/ansible/ansible-hub-ui/runs/7056584518?check_suite_focus=true :heavy_check_mark: 
https://github.com/ansible/ansible-hub-ui/runs/7056716828?check_suite_focus=true :heavy_check_mark: (this one failed in remote registry, but not in imports)
https://github.com/ansible/ansible-hub-ui/runs/7056907661?check_suite_focus=true :heavy_check_mark: (this one failed in view_only, but not in imports)
https://github.com/ansible/ansible-hub-ui/runs/7064717705?check_suite_focus=true :heavy_check_mark:

PR #2231, opened 2022-06-22
Upload signature on collection detail page

brumik:
When the flag `can_upload_signatures` is set it takes precedence over the backend signing. Therefore we are hiding the `Sign all` buttons on the collection and namespace detail pages, and changing the `Sign version x.x` to bring up the certificate upload modal. 

In addition it fixes the `waitForTask` function to return error if the task is not `running` and not `completed` so we can report failures.

**After with the `can_upload_signatures = true`**:
Screenshot from 2022-06-20 13-06-02
Screenshot from 2022-06-20 13-06-42
Screenshot from 2022-06-20 13-07-09
Screenshot from 2022-06-20 13-07-15
Screenshot from 2022-06-20 13-07-31
Screenshot from 2022-06-20 13-08-41
)

### DIFF
--- a/src/components/headers/collection-header.tsx
+++ b/src/components/headers/collection-header.tsx
@@ -37,6 +37,7 @@ import {
   DeleteModal,
   SignSingleCertificateModal,
   SignAllCertificatesModal,
+  UploadSingCertificateModal,
   ImportModal,
 } from 'src/components';
 
@@ -46,6 +47,9 @@ import {
   SignCollectionAPI,
   CollectionListType,
   MyNamespaceAPI,
+  CollectionVersion,
+  Repositories,
+  CertificateUploadAPI,
 } from 'src/api';
 import { Paths, formatPath } from 'src/paths';
 import {
@@ -93,6 +97,8 @@ interface IState {
   isDeletionPending: boolean;
   updateCollection: CollectionListType;
   showImportModal: boolean;
+  uploadCertificateModalOpen: boolean;
+  versionToUploadCertificate: CollectionVersion;
 }
 
 export class CollectionHeader extends React.Component<IProps, IState> {
@@ -120,6 +126,8 @@ export class CollectionHeader extends React.Component<IProps, IState> {
       isDeletionPending: false,
       updateCollection: null,
       showImportModal: false,
+      uploadCertificateModalOpen: false,
+      versionToUploadCertificate: undefined,
     };
   }
 
@@ -177,7 +185,8 @@ export class CollectionHeader extends React.Component<IProps, IState> {
 
     const latestVersion = collection.latest_version.created_at;
 
-    const { display_signatures } = this.context?.featureFlags || {};
+    const { display_signatures, can_upload_signatures } =
+      this.context?.featureFlags || {};
 
     const signedString = (v) => {
       if (display_signatures && 'sign_state' in v) {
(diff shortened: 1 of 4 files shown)


himdel (approved):
LGTM :)

src/components/headers/collection-header.tsx
```diff
@@ -639,6 +662,69 @@ export class CollectionHeader extends React.Component<IProps, IState> {
     return versions.slice(0, count);
   }
 
+  private async submitCertificate(file: File) {
+    const version = this.state.versionToUploadCertificate;
+    const response = await Repositories.getRepository({
+      name: this.context.selectedRepo,
+    });
+    const signed_collection = `${PULP_API_BASE_PATH}content/ansible/collection_versions/${version.id}/`;
+    console.log(
```
himdel, line 671: is the logging still useful?

himdel:
(restarting once more, the test failures *seem* unrelated, but...)

src/components/headers/collection-header.tsx
brumik: Good catch, oversight after long debugging :)

brumik:
@himdel I re run it, but the same imports_filter test fails every time. I checked out locally and the tests is passing there. It seems pretty unrelated though.

brumik:
@himdel same tests fails here too: https://github.com/ansible/ansible-hub-ui/pull/2228

himdel:
Agreed, imports_filter is broken on master, feel free to ignore it for now :). (I started looking into it in https://github.com/ansible/ansible-hub-ui/pull/2165)

But I'm also seeing ✖`collection_detail.js` ✖`collection.js` ✖`collections_list.js` here

brumik:
That's new, I didnt see them before, let me look into it

brumik:
@himdel I cannot reproduce these locally, and sometimes other tests are failing

Screenshot from 2022-06-30 12-23-09
Screenshot from 2022-06-30 11-08-52

himdel:
Restarting tests, hopefully that should tell us if it's something in this PR or coming from elsewhere :)

src/utilities/wait-for-task.ts
```diff
@@ -3,6 +3,9 @@ import { TaskAPI } from 'src/api';
 
 export function waitForTask(task, bailAfter = 10) {
   return TaskAPI.get(task).then((result) => {
+    if (result.data.state !== 'completed' && result.data.state !== 'running') {
+      return Promise.reject(result.data.error.description);
```
himdel, line 7: this is causing the test failures ... error may be null

himdel: I think this may be causing the rest as well, some callers are not expecting waitForTask to fail.

I think this *may* be because the task doesn't have to be started yet .. we may need to check for `queued` as well

brumik: Oh, yeah, and locally it is not a problem, this explains the flankyness. Thank you looking into this.

himdel: Any time :) hope that gets all the errors :)

src/utilities/wait-for-task.ts
```diff
@@ -3,6 +3,12 @@ import { TaskAPI } from 'src/api';
 
 export function waitForTask(task, bailAfter = 10) {
   return TaskAPI.get(task).then((result) => {
+    const failing = ['skipped', 'failed', 'canceled'];
+
+    if (failing.includes(result.data.state)) {
+      return Promise.reject(result.data.error.description);
```
himdel, line 9: It still doesn't need to have non-null errror...

```suggestion
      return Promise.reject(result.data.error?.description);
```

himdel (approved):
:tada: :+1:  :)

brumik:
Passing, merging

PR #2234, opened 2022-06-23
[stable-4.2] Change node to 16+ (#1829)

himdel:
backport node 16 update to stable branches (#1829)


himdel:
Oh, right .. for 4.2 it's either node 16 or node-sass.
IIRC non-node- sass has some interdependencies with sass-loader, but maybe solvable.

PR #2238, opened 2022-06-23
EE list - fix toolbar alignment

himdel:
comes from #2209, it was the right fix for empty state but now the "Add execution environment" button is misaligned in the toolbar (when there are data)

moving the `<div>&nbsp;</div>` bit to the empty state only

before:

after:


himdel:
Thanks :), looks like the `imports_filter` failure happens on master as well, addressing in #2165

PR #2244, opened 2022-06-25
cypress plugins: use cypress-log-to-output instead of custom script, remove task log

himdel:
replacing local console-logger with https://github.com/flotwig/cypress-log-to-output

cy.task is unused, removing the log


himdel:
oh this outputs a LOT

we may want to just remove the plugin if cypress 10 has a problem with our custom one

PR #2267, opened 2022-06-29
Fix additional perms

jerabekjiri:
No Issue

- hide `sync from registry` button in EE detail for users without permission (list view fixed in [#2189](https://github.com/ansible/ansible-hub-ui/pull/2189))
before:
Screenshot from 2022-06-28 17-41-29
after:
Screenshot from 2022-06-28 17-42-06

- hide the delete button in the EE image tab for users without permission
Screenshot from 2022-06-28 18-32-39

### Owner's tab 
If the user doesn't have group permissions (`view_group`), he can't load roles in the list or modal (at the moment, `select roles` keeps loading in the modal), therefore should we limit visiting this tab? @himdel 

missing permissions (infinite loading) errors:
Screenshot from 2022-06-29 14-00-04

Screenshot from 2022-06-29 14-00-31

and only the superuser and users with group permissions should be display tab: 
Screenshot from 2022-06-28 19-48-44

- hide kebab button `Remove group` if user doesn't have `change_containernamespace` permission
Screenshot from 2022-06-29 15-18-33

- hide kebab button `Remove role` and `Add roles` button if user isn't superuser
Screenshot from 2022-06-29 15-02-11

### DIFF
--- a/src/components/rbac/owners-tab.tsx
+++ b/src/components/rbac/owners-tab.tsx
@@ -9,6 +9,8 @@ import {
   ToolbarContent,
   ToolbarItem,
 } from '@patternfly/react-core';
+import { ExclamationTriangleIcon } from '@patternfly/react-icons';
+
 import { GroupType, RoleType } from 'src/api';
 import {
   DeleteModal,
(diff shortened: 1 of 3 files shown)


himdel:
This looks about right, we're still letting users access the data, but not edit it :+1:

himdel (approved):

PR #2268, opened 2022-06-29
[rbac] cypress: install galaxykit branch with roles

himdel:
So that we can use `galaxykit role create/remove`, `galaxykit group role add/remove` etc. before this is merged in master.

AAH-1691, AAH-1692, https://github.com/ansible/galaxykit/pull/39


jerabekjiri (approved):
LGTM :+1: 
clever to switch to a fork :smile:

PR #2269, opened 2022-06-30
Update repo management collection count

ShaiahWren:
Issue: AAH-1677

This pr updates what used to be the `content count` for each repository to a `collection count` that reflects the excludes at `content/${repositoryPath}/v3/excludes/` subtracted from the published collections at `v3/plugin/ansible/content/${repositoryPath}/collections/index/`(meta.count).

### DIFF
--- a/src/api/collection.ts
+++ b/src/api/collection.ts
@@ -64,6 +64,29 @@ export class API extends HubAPI {
     }));
   }
 
+  getPublishedCount(repositoryPath: string) {
+    return new Promise((resolve, reject) => {
+      this.http
+        .get(`v3/plugin/ansible/content/${repositoryPath}/collections/index/`)
+        .then((result) => {
+          resolve(result.data.meta.count);
+        })
+        .catch((err) => reject(err));
+    });
+  }
+
+  getExcludesCount(repositoryPath: string) {
+    return new Promise((resolve, reject) => {
+      this.http
+        .get(`content/${repositoryPath}/v3/excludes/`)
+        .then((result) => {
+          resolve(result.data);
+          console.log('data: ', result.data);
+        })
+        .catch((err) => reject(err));
+    });
+  }
+
   setDeprecation(
     collection: CollectionListType,
     isDeprecated: boolean,
--- a/src/api/collection.ts
+++ b/src/api/collection.ts
@@ -64,17 +64,17 @@ export class API extends HubAPI {
     }));
   }
 
-  getPublishedCount(repositoryPath: string) {
+  getPublishedCount(distributionPath: string) {
     return this.http
-      .get(`v3/plugin/ansible/content/${repositoryPath}/collections/index/`)
+      .get(`v3/plugin/ansible/content/${distributionPath}/collections/index/`)
       .then((result) => {
         return result.data.meta.count;
       });
   }
 
-  getExcludesCount(repositoryPath: string) {
+  getExcludesCount(distributionPath: string) {
     return this.http
-      .get(`content/${repositoryPath}/v3/excludes/`)
+      .get(`content/${distributionPath}/v3/excludes/`)
       .then((result) => {
         return result.data;
       });
--- a/src/api/collection.ts
+++ b/src/api/collection.ts
@@ -81,7 +81,6 @@ export class API extends HubAPI {
         .get(`content/${repositoryPath}/v3/excludes/`)
         .then((result) => {
           resolve(result.data);
-          console.log('data: ', result.data);
         })
         .catch((err) => reject(err));
     });
--- a/src/api/collection.ts
+++ b/src/api/collection.ts
@@ -65,25 +65,21 @@ export class API extends HubAPI {
   }
 
   getPublishedCount(repositoryPath: string) {
-    return new Promise((resolve, reject) => {
-      this.http
-        .get(`v3/plugin/ansible/content/${repositoryPath}/collections/index/`)
-        .then((result) => {
-          resolve(result.data.meta.count);
-        })
-        .catch((err) => reject(err));
-    });
+    return this.http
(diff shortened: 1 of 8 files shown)


src/api/collection.ts
```diff
@@ -64,6 +64,28 @@ export class API extends HubAPI {
...
 
+  getPublishedCount(repositoryPath: string) {
+    return new Promise((resolve, reject) => {
+      this.http
+        .get(`v3/plugin/ansible/content/${repositoryPath}/collections/index/`)
+        .then((result) => {
+          resolve(result.data.meta.count);
+        })
+        .catch((err) => reject(err));
+    });
```
himdel, line 75: this whole bit is equivalent to

```suggestion
    return this.http
      .get(`v3/plugin/ansible/content/${repositoryPath}/collections/index/`)
      .then((result) => result.data.meta.count);
```

src/api/collection.ts
```diff
@@ -64,6 +64,28 @@ export class API extends HubAPI {
...
+
+  getExcludesCount(repositoryPath: string) {
+    return new Promise((resolve, reject) => {
+      this.http
+        .get(`content/${repositoryPath}/v3/excludes/`)
+        .then((result) => {
+          resolve(result.data);
+        })
+        .catch((err) => reject(err));
+    });
```
himdel, line 86: (and here)

```suggestion
    return this.http
      .get(`content/${repositoryPath}/v3/excludes/`)
      .then((result) => result.data);
```

src/api/response-types/collection.ts
```diff
@@ -111,6 +111,10 @@ export class ContentSummaryType {
   description: string;
 }
 
+export class CollectionExcludesType {
+  collections: Array<string>;
```
himdel, line 115:
```suggestion
  collections: string[];
```

src/components/repositories/local-repository-table.tsx
```diff
@@ -110,7 +135,7 @@ export class LocalRepositoryTable extends React.Component<IProps> {
       <tr key={distribution.name}>
         <td>{distribution.name}</td>
         <td>{distribution.repository.name}</td>
-        <td>{distribution.repository.content_count}</td>
+        {this.getCollectionCount(distribution.base_path)}
```
himdel, line 138: `getCollectionCount` returns a Promise, not the actual value .. this has to be 2 step process:

* call this on load, not every time we render, wait on the promise and save the result in state
* render the value from state here

(alternately, you could wrap all that in component, so you'd do `<td><CollectionCount repositoryPath={distribution.base_path} /></td>` here, and have a new component which does the API calling and saving in state and renders a spinner or something while loading)

src/components/repositories/local-repository-table.tsx
```diff
@@ -110,7 +111,7 @@ export class LocalRepositoryTable extends React.Component<IProps> {
       <tr key={distribution.name}>
         <td>{distribution.name}</td>
         <td>{distribution.repository.name}</td>
-        <td>{distribution.repository.content_count}</td>
+        <td>{<CollectionCount repositoryPath={distribution.base_path} />}</td>
```
himdel, line 114:
```suggestion
        <td><CollectionCount repositoryPath={distribution.base_path} /></td>
```

src/components/collection-count/collection-count.tsx
```diff
@@ -0,0 +1,89 @@
...
+import { CollectionAPI, CollectionExcludesType } from 'src/api';
+import { Spinner } from '@patternfly/react-core';
+import { AlertType } from 'src/components';
+import { errorMessage } from 'src/utilities';
+
+interface IProps {
+  repositoryPath: string;
+}
+interface IState {
+  collectionCount: string;
```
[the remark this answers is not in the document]
ShaiahWren: yeah, I was thinking about `0` being falsy, but otherwise, just a number :)

My concern is if `!collectionCount && !loading` then it won't render anything and mess up the table row layout... but then again, `<CollectionCount>` in the local repo table is wrapped by a `td` anyway, so maybe it's fine?

awcrosby (approved):

src/api/collection.ts
```diff
@@ -64,6 +64,22 @@ export class API extends HubAPI {
...
+    return this.http
+      .get(`v3/plugin/ansible/content/${repositoryPath}/collections/index/`)
+      .then((result) => {
+        return result.data.meta.count;
+      });
+  }
+
+  getExcludesCount(repositoryPath: string) {
+    return this.http
+      .get(`content/${repositoryPath}/v3/excludes/`)
```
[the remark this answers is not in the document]
ShaiahWren: I can make the change.

src/components/collection-count/collection-count.tsx
```diff
@@ -0,0 +1,83 @@
...
+    };
+  }
+
+  componentDidMount() {
+    this.getCollectionCount(this.props.distributionPath);
+  }
+
+  render() {
+    const { collectionCount, loading } = this.state;
+    return !loading ? <td>{collectionCount}</td> : <Spinner />;
```
himdel, line 33:
```suggestion
    return !loading ? <>{collectionCount}</> : <Spinner />;
```

(or a span)

(`CollectionCount` is already rendered within a td, so this would produce a `<td><td>...`)

src/components/collection-count/collection-count.tsx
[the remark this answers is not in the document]
himdel: Works just fine when relying on `loading` only :+1:

himdel:
Sorryyyy.. one more thing I noticed ...

The loading spinner size is too large, and causes the table rows to change size...


can we make it smaller? :)

With `<Spinner size="md" />`:


With `<Spinner size="sm" />`:


I guess we should go for `sm` as we're using it in 4 places already.

himdel (approved):

PR #2280, opened 2022-07-01
Surfacing feature flag misconfig messages

brumik:
Issue: AAH-1739

Backend PR: https://github.com/ansible/galaxy_ng/pull/1332

Preview where it is working:
Screenshot from 2022-07-01 12-47-02

### DIFF
--- a/CHANGES/1739.feature
+++ b/CHANGES/1739.feature
@@ -1,0 +1,1 @@
+Surfacing feature misconfiguration alert messages.
--- a/src/api/response-types/feature-flags.ts
+++ b/src/api/response-types/feature-flags.ts
@@ -10,4 +10,5 @@ export class FeatureFlagsType {
   display_signatures: boolean;
   require_upload_signatures: boolean;
   signatures_enabled: boolean;
+  _messages: string[];
 }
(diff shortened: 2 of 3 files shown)


brumik:
There is a problem that some pages does not use the global state for displaying alerts. This can cause the alert not to show up on pages. The pages that the alert does shows up are the following:
* Namespaces
* Execution environment

As I see it:
- We can leave this PR as it is, and later update slowly the components to use the context for alerts.
- We can do a little refactor so the pages are using the context for alerts on every page.
- We can try to display these messages differently (like a banner or something)

@himdel What do you think?

himdel:
@brumik Ah, makes sense .. I'd go with option 1 for now.

The plan is to unify the alerts to use context anyway (https://issues.redhat.com/browse/AAH-1462),
but we also need to define when alerts reset, otherwise no alerts will ever disappear again.

I'd also be ok with just making alerts everywhere update from context on load, without moving/saving back to context, but not sure if load happens consistently after the feature flags request, or how many places which already save to context we'd have to add cleaning to. If we can ensure everything only saves alerts to context right before navigating elsewhere, without also using it for alerts only relevant on that one screen; and that featureFlags is always loaded before a container's componentDidMount happens, I think we could safely add `componentDidMount() { this.setState({ alerts: this.context.alerts || [] }); }` everywhere.

brumik:
@himdel I understand the issue we are having here. So should I add something to this PR? If not this is ready or review.

himdel (approved):

PR #2281, opened 2022-07-01
Fix the error message when uploading certificate

brumik:
Issue: AAH-1623

**After**:
Screenshot from 2022-07-01 15-06-35

### DIFF
--- a/src/containers/certification-dashboard/certification-dashboard.tsx
+++ b/src/containers/certification-dashboard/certification-dashboard.tsx
@@ -34,7 +34,13 @@ import {
   CertificateUploadAPI,
   Repositories,
 } from 'src/api';
-import { errorMessage, filterIsSet, ParamHelper } from 'src/utilities';
+import {
+  errorMessage,
+  filterIsSet,
+  ParamHelper,
+  parsePulpIDFromURL,
+  waitForTask,
+} from 'src/utilities';
 import {
   LoadingPageWithHeader,
   CompoundFilter,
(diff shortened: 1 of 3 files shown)


src/containers/certification-dashboard/certification-dashboard.tsx
```diff
@@ -522,19 +528,25 @@ class CertificationDashboard extends React.Component<
...
-          version,
-        );
-        this.addAlert(
-          <Trans>
-            Certificate for collection &quot;{version.namespace} {version.name}{' '}
-            v{version.version}&quot; has been successfully uploaded.
-          </Trans>,
-          'success',
-        );
+        waitForTask(parsePulpIDFromURL(result.data.task))
```
himdel, line 531: you may want to `return` here, so the catch can be shared? They seem identical except for the error in description.

himdel (approved):
LGTM :+1: 

I'm fixing more stuff around `waitForUpdate` in #2312 for a different issue, but this is moving `submitCertificate` off the `waitForUpdate` path, so sounds like both are relevant :)

src/containers/certification-dashboard/certification-dashboard.tsx
brumik: I am not sure if we should, the returned error can be different and I am fine for having and "almost" duplicate code if it means lesser chance of breaking the error reporting. (Note that the first error uses the error as text - how the waitForTask works, while the second is more general and using the `status` and `statusText` which is a top level property in the API return.) So unless we refactor how the `waitForTask` returns errors, I think it is better to have this.

PR #2282, opened 2022-07-01
Added a download icon to the approval page

brumik:
Issue: AAH-1621

The issue was created to make the signing process easier when user needs to download version - sign - upload signature. 

**After** 
Screenshot from 2022-07-01 15-18-57

### DIFF
--- a/src/containers/certification-dashboard/certification-dashboard.tsx
+++ b/src/containers/certification-dashboard/certification-dashboard.tsx
@@ -19,12 +19,14 @@ import {
   Button,
   DropdownItem,
   Label,
+  ButtonVariant,
 } from '@patternfly/react-core';
 
 import {
   ExclamationTriangleIcon,
   ExclamationCircleIcon,
   CheckCircleIcon,
+  DownloadIcon,
 } from '@patternfly/react-icons';
 
 import {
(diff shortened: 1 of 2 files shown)


src/containers/certification-dashboard/certification-dashboard.tsx
```diff
@@ -647,6 +658,14 @@ class CertificationDashboard extends React.Component<
     );
   }
 
+  private download(namespace: string, name: string, version: string) {
+    CollectionAPI.getDownloadURL('staging', namespace, name, version).then(
+      (downloadURL: string) => {
+        window.location.assign(downloadURL);
```
himdel, line 664: will this still work if you click it multiple times?

Similar code exists in `collection-info`, but it's triggering an `<a>` element to prevent popup blocking issues, I suspect we'd have to do the same here?

himdel: Aaah, lovely :) Thanks for looking into this, if this was about IE support, yes, we can safely ignore that, we can clean up the other place too :)

himdel (approved):

rochacbruno (approved):

PR #2307, opened 2022-07-05
Change pr_check.sh to node 16

drodowic:
Bumps the node version used in the frontend container build to Node 16
Change to Node version required to match https://github.com/ansible/ansible-hub-ui/pull/1829


himdel (approved):
Thanks! :)

PR #2309, opened 2022-07-05
Localization Scripts for Hub UI

adityamulik:
Hello,

The pull request is for the purpose of automating the localization of Ansible-Hub-UI. 
We're actively developing an Ansible collection for this process called as [Ansible Memsource Collection](https://github.com/ansible/ansible-collection-memsource)

The process comprises of two tasks:
- Pull Strings from the project (Using the project framework's strings extraction command). These strings (Usually retrieved are in English) are then uploaded to a third party software called Memsource
- Push Strings to the projects, once the translations are completed and the strings are ready to be pushed back to the repository, the push/ post_translation role of this collection is executed

For this process, we require two shell scripts to be placed in the repository. 
- pre_translation.sh - which performs the extraction of strings from a project. In this process, a folder called "translations" will be provided, the script ensures to move the strings to the translations folder
   - Empty translations folder
- post_translation.sh - which performs the process of moving the strings to appropriate directory in the project, a PR is later autogenerated by the underlying Ansible role being utilized here for merging the translated strings. Again a translations folder will be provided with all the translated strings, the script would ensure to utilize the folder and move the files according to the project needs.
   - Translation folder with all languages generated files
   e.g. 
          /translations/fr/en.po 
          /translations/jp/en.po
          /translations/ko/en.po

### DIFF
--- a/tools/scripts/l18n/post_translation.sh
+++ b/tools/scripts/l18n/post_translation.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+# Rename the zh_cn directory 
+mv translations/zh_cn translations/zh
+
+# Rename files and remove from lang directory
+cd translations
+
+for d in */ ; do
+    dir=${d%*/}
+    mv $dir/en.po $dir.po
+    rm -rf $dir
+done
+
+cd ..
+
+# locale will be dropped here
+ansible_hub_ui_path="locale" 
+
+# Overwrite the new files
+rsync -av translations/ $ansible_hub_ui_path
+
+# Cleanup
+rm -rf translations/
--- a/tools/scripts/l18n/pre_translation.sh
+++ b/tools/scripts/l18n/pre_translation.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+# Extract UI Strings
+npm run gettext:extract
+
+# Move files to translations folder
+mv locale/en.po translations/


himdel:
@adityamulik do these changes belong in this repo?

Most of the script seems concerned with:
* switching to hardcoded `_clones/ansible-hub-ui`, without showing how to create it
* copying files back and forth between `locale/` and `translations/`, without showing how `translations/` gets created or what happens in between
* renaming one specific locale form a long form to a short one, but we never use the long version here

Seems like all these are very interdependent part of some larger toolchain, so maybe they should live there?
(Or, if you really want to add them here, we should at least add a comment with a link to more info, whether it's the repo using it, interface docs, or a jira ticket describing the purpose.)

Can you also write a PR description please?

tools/scripts/l18n/post_translation.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash/
+
+# Change Directory to clones
+cd _clones/ansible-hub-ui/
+
+# Rename the zh_cn folder 
+mv translations/zh_cn translations/zh
```
himdel, line 7: if `translations/zh_cn/` is a folder, then `locale/zh/` is also a folder at the end

that's the wrong format, we expect a `locale/zh.po` file

tools/scripts/l18n/pre_translation.sh
```diff
@@ -0,0 +1,12 @@
...
+# Change Directory to clones
+cd _clones/ansible-hub-ui/
+
+# Extract UI Strings
+npm run gettext:extract
+
+# Move files to translations folder
+mv locale/en.po translations/
+
+sleep 5
```
himdel, line 11: (nothing started here is asynchronous, the sleep should not be necessary)

tools/scripts/l18n/pre_translation.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash/
```
himdel, line 1:
```suggestion
#!/bin/bash
```

adityamulik:
Hi @himdel,

I've updated the description of the PR and the purpose of these two translation shell scripts.

1. For the _clones/ansible-hub-ui name, I have opened an issue in the ansible-memsource-collection project to fix the static path being used and change to dynamic. Similarly, I've removed the static path from the next commit as the path was redundant. 
Ref: https://github.com/ansible/ansible-collection-memsource/issues/25
Ref: 4724a76bc00d9e09d89d0b45097e578a61fa044e

2. I've updated the script to have the files as per the project's current structure. 
We would be receiving file from Memsource in below structure.

e.g.
/translations/fr/en.po
/translations/jp/en.po
/translations/ko/en.po

The script would modify the files and sync it to the required location
/translations/fr/en.po -> /locale/fr.po

3. We also receive the lang zh as zh_cn from Memsource, so the script ensures to first move the folder from zh_cn -> zh
Then move the file zh/en.po -> zh.po

We have two underlying roles for localization
1. https://github.com/ansible/ansible-collection-memsource/blob/main/roles/pre_translation/README.md
2. https://github.com/ansible/ansible-collection-memsource/blob/main/roles/post_translation/README.md

tools/scripts/l18n/pre_translation.sh
adityamulik: sleep command has been removed
4724a76bc00d9e09d89d0b45097e578a61fa044e

tools/scripts/l18n/pre_translation.sh
adityamulik: trailing slash has been removed.
4724a76bc00d9e09d89d0b45097e578a61fa044e

rooftopcellist:
@himdel One benefit to having these scripts live in the repo with the code is that if the structure changes over time, the script can be modified on that branch / versioned with the code it acts on.  

But I've added the ability to specify a script local to the memsource roles which gets copied into the clone of the app (in this case ansible-hub-ui), so we can close this PR and move forward with that.  
* https://github.com/ansible/ansible-collection-memsource/pull/45

adityamulik:
Closing the PR as applicable local script changes has been done as per https://github.com/ansible/ansible-collection-memsource/pull/45

PR #2312, opened 2022-07-07
Approvals: don't show success alert on `updateCertification` failure, `waitForTask` first

himdel:
Fixes AAH-1769

Introduced in #1469, `.then(() => { setState(); waitForUpdate(); })` became `.then(() => { setState(); waitForUpdate(); }, addAlert())` instead of `.then(() => { setState(); waitForUpdate(); addAlert(); })` in `updateCertification()`, causing the success alert to appear even on failure, and definitely before any waiting was done.

Also the alerts were actually not waiting for the task, moving so this happens once the task was completed or failed.

Also, the `waitForUpdate` logic had extra copy of the failure alert because it was never rejecting the promise => replacing by the shared `waitForTask` + a single then to requery versions, and a compatible rejection.

And updated `submitCertificate` in very much the same way, though the issue there is already fixed since #2281.

### DIFF
--- a/src/containers/certification-dashboard/certification-dashboard.tsx
+++ b/src/containers/certification-dashboard/certification-dashboard.tsx
@@ -1,4 +1,4 @@
-import { t, Trans } from '@lingui/macro';
+import { t } from '@lingui/macro';
 import * as React from 'react';
 import './certification-dashboard.scss';
 
@@ -75,7 +75,7 @@ interface IState {
   unauthorized: boolean;
   inputText: string;
   uploadCertificateModalOpen: boolean;
-  versionToUploadCertificate: CollectionVersion;
+  versionToUploadCertificate?: CollectionVersion;
 }
 
 class CertificationDashboard extends React.Component<
@@ -112,7 +112,7 @@ class CertificationDashboard extends React.Component<
       unauthorized: false,
       inputText: '',
       uploadCertificateModalOpen: false,
-      versionToUploadCertificate: undefined,
+      versionToUploadCertificate: null,
     };
   }
 
@@ -526,11 +526,18 @@ class CertificationDashboard extends React.Component<
     });
   }
 
+  private closeUploadCertificateModal() {
+    this.setState({
+      uploadCertificateModalOpen: false,
+      versionToUploadCertificate: null,
+    });
+  }
+
   private submitCertificate(file: File) {
     const version = this.state.versionToUploadCertificate;
     const signed_collection = `${PULP_API_BASE_PATH}content/ansible/collection_versions/${version.id}/`;
 
-    Repositories.getRepository({
+    return Repositories.getRepository({
       name: 'staging',
     })
       .then((response) =>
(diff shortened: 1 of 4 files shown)


src/containers/certification-dashboard/certification-dashboard.tsx
```diff
@@ -527,136 +526,81 @@ class CertificationDashboard extends React.Component<
...
-            (result) =>
-              // Since pulp doesn't reply with the new object, perform a
-              // second query to get the updated data
-              {
-                this.setState({
-                  updatingVersions: [version],
-                });
-                this.waitForUpdate(result.data.remove_task_id, version);
-              },
-            this.addAlert(
```
himdel, line 610: (the actual bug, if we need to backport a smaller change)

himdel:
Restarting tests...
I expect `group_roles` and `rbac_owners` to still fail (though they both pass locally),
but not `namespace_detail`

src/containers/certification-dashboard/certification-dashboard.tsx
```diff
@@ -527,136 +526,81 @@ class CertificationDashboard extends React.Component<
...
+      )
+      .then((result) => waitForTask(parsePulpIDFromURL(result.data.task)))
+      .then(() =>
+        this.addAlert(
+          t`Certificate for collection "${version.namespace} ${version.name} v${version.version}" has been successfully uploaded.`,
+          'success',
+        ),
+      )
+      .then(() => CollectionVersionAPI.list(this.state.params))
+      .then((result) => this.setState({ versions: result.data.data }))
```
MilanPospisil, line 558: I see one minor problem, when certificate is uploaded and CollectionVersionAPI.list list fails, there will be error message t`The certificate for "${version.namespace} ${version.name} v${version.version}" could not be saved.` (along with success message). Success message will definetly be ok, since certificate will be really uploaded. But error message may be confusing, because the list will fail, not the saving of certificate.

src/containers/certification-dashboard/certification-dashboard.tsx
```diff
@@ -527,136 +526,81 @@ class CertificationDashboard extends React.Component<
...
+    )
+      .then((result) => waitForTask(result.data.remove_task_id, 500))
+      .then(() =>
+        this.addAlert(
+          t`Certification status for collection "${version.namespace} ${version.name} v${version.version}" has been successfully updated.`,
+          'success',
+        ),
+      )
+      .then(() => CollectionVersionAPI.list(this.state.params))
+      .then((result) => this.setState({ versions: result.data.data }))
```
MilanPospisil, line 591: The same problem with confusing error message here.

src/containers/certification-dashboard/certification-dashboard.tsx
himdel: :+1: Yeah, let's just replace `list` + `setState` with an actual `queryCollections` call, so that we fire the query on success, but don't let it reject the promise when that fails.

And added an error alert to `queryCollections` .. should be ready :)

MilanPospisil:
Settings for sign and approve in .compose.env:

ENABLE_SIGNING=1
PULP_GALAXY_REQUIRE_CONTENT_APPROVAL=true
PULP_GALAXY_REQUIRE_SIGNATURE_FOR_APPROVAL=false
PULP_GALAXY_AUTO_SIGN_COLLECTIONS=true

MilanPospisil (approved):
LGTM 👍

src/utilities/wait-for-task.ts
```diff
@@ -19,7 +19,7 @@ export function waitForTask(task, bailAfter = 10) {
         );
       }
 
-      return new Promise((r) => setTimeout(r, 5000)).then(() =>
+      return new Promise((r) => setTimeout(r, waitMs)).then(() =>
         waitForTask(task, bailAfter - 1),
```
himdel, line 23: introduces a bug (bailAfter - 1 used for waitMs instead), fixed in https://github.com/ansible/ansible-hub-ui/pull/2598

PR #2333, opened 2022-07-11
standardize ephemeral setup

drodowic:
Update pr_check.sh to use standard setup script for ephemeral.
Removes links to scripts deleted from https://github.com/ansible/galaxy_ng/pull/1342.


himdel (approved):
Ah, that explains it :)

LGTM :+1:

PR #2335, opened 2022-07-12
Enable FeatureFlagsAPI in insights mode

himdel:
Issue: AAH-1800

Until now, `featureFlags` is always empty in insights mode.

This brings standalone and insights a bit closer, adding `featureFlags` handling to `insights-loaders`, moving the logic to a shared `loadContext` function, making sure the list of args for `AppContext` is the same in both modes, and cleaning up some `getUser` / `getActiveUser` confusion - we wait for insights `getUser` before any API request, and the loader waits for out user API, achieving both. (Also we never use the insights user data, except for some dead code, so `ActivetUserAPI.getUser` should always return `UserType`.)
And converted insights loaders and routes to typescript, to match the rest, and to typecheck this has a chance of being right :)

Also found some extra top padding in insights mode and a navitem border outside the menu area, fixing both.

### DIFF
--- a/CHANGES/1800.misc
+++ b/CHANGES/1800.misc
@@ -1,0 +1,1 @@
+Read feature flags in insights mode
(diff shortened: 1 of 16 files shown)


awcrosby (approved):

himdel:
Restarting tests

PR #2336, opened 2022-07-14
[rbac] group-list: filter "Group name" by name__icontains, not exact name

himdel:
Issue: AAH-1806

Filtering by group name used an exact match .. changing to name__icontains for case insensitive partial match

Before:

After:

+ make sure the "Group name" is consistently used in the filter name and the column name

### DIFF
--- a/src/containers/group-management/group-list.tsx
+++ b/src/containers/group-management/group-list.tsx
@@ -128,7 +128,7 @@ class GroupList extends React.Component<RouteComponentProps, IState> {
 
     const { user } = this.context;
     const noData =
-      groups.length === 0 && !filterIsSet(params, ['name__icontains']);
+      groups.length === 0 && !filterIsSet(params, ['name__contains']);
 
     if (redirect) {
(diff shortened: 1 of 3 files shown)


himdel:
Ah, api doesn't currently support `name__icontains` in groups, using `name__contains` for now.

src/containers/group-management/group-list.tsx
```diff
@@ -220,6 +221,9 @@ class GroupList extends React.Component<RouteComponentProps, IState> {
                   }}
                   params={params}
                   ignoredParams={['page_size', 'page', 'sort']}
+                  niceNames={{
+                    name__contains: 'Group name',
```
jerabekjiri, line 225: missing translation?

jerabekjiri (approved):
Other than missing translation, LGTM :+1:

src/containers/group-management/group-list.tsx
himdel: Aah, thanks! :) Fixed

PR #2374, opened 2022-07-25
Can't see imports when there are more than 100 namespaces.

MilanPospisil:
https://issues.redhat.com/browse/AAH-1725

What is implemented:

DropDown in imports list page was replaced by search ahead component. You can type substring and available namespaces are shown in dropdown (so you can access even more than 100 of them).

Selected namespace is determined by url query parameter 'namespace' by default.

Import page queries for the imports data of namespace is based on query parameter namespace. The query parameter is updated when user changes selection in search ahead dropdown.

Namespaces are now as state in list component and are querying only for namespaces (and not any of the imports data). Thus namespaces state is removed from Import page.

Search ahead dropdown allows to clear selection and it then displays No Data label.

Tests are updated according to changes.


MilanPospisil:
Signing test seems to be broken also in master, seems like non related issue.

src/components/my-imports/import-list.tsx
```diff
@@ -219,28 +231,45 @@ export class ImportList extends React.Component<IProps, IState> {
     }
   }
 
-  private renderNamespacePicker(namespaces) {
+  private loadNamespaces(namespace_filter) {
+    MyNamespaceAPI.list({ page_size: 100, keywords: namespace_filter })
```
newswangerd, line 235: lets limit the page size to 10

newswangerd (changes requested):
Selecting a namespace with a collection on it seems to cause the filter to get cleared out.
Screen Recording 2022-08-08 at 3 37 51 PM mov

src/components/my-imports/import-list.tsx
MilanPospisil: Thanks for catching that. The selection based on filter of loaded namespaces was not good idea, now it select exactly what it is in the params.

src/components/my-imports/import-list.tsx
```diff
@@ -219,28 +231,43 @@ export class ImportList extends React.Component<IProps, IState> {
     }
   }
 
-  private renderNamespacePicker(namespaces) {
+  private loadNamespaces(namespace_filter) {
+    MyNamespaceAPI.list({ page_size: 10, keywords: namespace_filter })
+      .then((result) => {
+        this.setState({ namespaces: result.data.data });
+      })
+      .catch((result) => console.log(result));
```
himdel, line 239: console.log .. this should be an alert?

himdel:
`src/containers/my-imports/my-imports.tsx:121` is missing the new `addAlert` prop for `ImportsFilter`,
otherwsie LGTM :)

MilanPospisil:
> 

I noticed it, however Im not sure how it happened... Meanwhile, I repaired one error - when search ahead text was empty, it passed API call with filter undefined, rather empty string - this causes no namespaces returned (unless prefixed by undefined :-) ).

himdel:
When visiting `/ui/my-imports` directly, previously this would open the first namespace, now none is selected.
That's OK, the left side message looks good, but the right side keeps spinning in such case:



^ can probably be fixed by using the same ``importDetailError: t`No data`,`` logic that's used when a namespace has no imports.

(



)

himdel:
When I unselect a namespace (using the :x: button in the namespace selector), it still triggers a `/api/automation-hub/_ui/v1/collection-versions/?namespace=null&name=null&version=null&offset=0&limit=10` (note the `null`s) request.

And somehow the right side ends up displaying some previous value:

MilanPospisil:
You are right, I forgot to handle empty namespace selection. The app would still work, but it will not look proffesional. I fixed it now.

src/containers/my-imports/my-imports.tsx
```diff
@@ -112,38 +129,54 @@ class MyImports extends React.Component<RouteComponentProps, IState> {
       return null;
     }
 
+    let empty: boolean;
+    empty = this.state.params.namespace ? true : false;
```
himdel, line 133: unused

src/containers/my-imports/my-imports.tsx
```diff
@@ -112,38 +129,54 @@ class MyImports extends React.Component<RouteComponentProps, IState> {
...
+                        });
+                      }
+                    });
                   }}
                 />
               </div>
 
               <div className='hub-import-console'>
                 <ImportConsole
+                  empty={this.state.params.namespace ? false : true}
```
himdel, line 179:
```suggestion
                  empty={!this.state.params.namespace}
```

himdel:
Just the 2 changes ^ to make linter checks happy, otherwise LGTM :+1:

PR #2378, opened 2022-07-26
[rbac] Fix owner tab permissions

newswangerd:
No-Issue

Set owners tab to use the permissions returned by my_permissions on container and collection namespaces.

Requires https://github.com/ansible/galaxy_ng/pull/1389


src/components/rbac/owners-tab.tsx
```diff
@@ -182,7 +182,7 @@ export class OwnersTab extends React.Component<IProps, IState> {
             {group.name}
           </Link>
         </td>
-        <ListItemActions kebabItems={dropdownItems} />
+        {canEditOwners && <ListItemActions kebabItems={dropdownItems} />}
```
himdel, line 185: we do need ListItemActions here for the table not to miss the last column.. but if you move this to dropdownItems, dropdownItems can be an empty array

himdel:
Tests are green on the rbac branch (https://github.com/ansible/ansible-hub-ui/runs/7559166337?check_suite_focus=true),
so looks like the failing tests are caused by this change

src/containers/namespace-detail/namespace-detail.tsx
```diff
@@ -568,7 +571,9 @@ export class NamespaceDetail extends React.Component<IProps, IState> {
         },
         this.context.selectedRepo,
       ),
-      NamespaceAPI.get(this.props.match.params['namespace']),
+      NamespaceAPI.get(this.props.match.params['namespace'], {
+        include_related: 'my_permissions',
```
himdel, line 575: does this mean people can now edit namespaces they don't own?

himdel:
> Set owners tab to use the permissions returned by my_permissions on container and collection namespaces.

I don't see this touching container namespaces at all, is that an oversight?

himdel:
The `rbac_owners` failue is happening because admin can not see the "Select a group" button on the EE namespace owners tab empty state.

himdel:
The `collections` test just needs adjusting the `cy.intercept....as('reload')` to match the new params for the namespace get.

himdel:
Turns out there's no way to change the base branch of a PR after the original goes away .. rebased and opened as https://github.com/ansible/ansible-hub-ui/pull/2451

PR #2355, opened 2022-07-18
cypress: pip install ansible

himdel:
seems to be needed since https://github.com/ansible/galaxykit/pull/54

attempts to fix tests failing with

```
Run galaxykit -s 'http://localhost:8002/api/galaxy/' -u admin -p admin collection list
Traceback (most recent call last):
  File "/home/runner/.local/bin/galaxykit", line 5, in <module>
    from galaxykit.command import main
  File "/home/runner/.local/lib/python3.8/site-packages/galaxykit/__init__.py", line 2, in <module>
    from .client import GalaxyClient
  File "/home/runner/.local/lib/python3.8/site-packages/galaxykit/client.py", line 16, in <module>
    from . import namespaces
  File "/home/runner/.local/lib/python3.8/site-packages/galaxykit/namespaces.py", line 2, in <module>
    from .utils import logger
  File "/home/runner/.local/lib/python3.8/site-packages/galaxykit/utils.py", line 6, in <module>
    from ansible.galaxy.api import GalaxyError
ModuleNotFoundError: No module named 'ansible'
Error: Process completed with exit code 1.
```

PR #2069, opened 2022-05-15
Added signing related tests

brumik:
Issue: AAH-1599

## TODO
* [x] Test badge as an extra tests to (also it will be included in the `signing` test suite):
  * [x] Collections list
  * [x] Namespace collection list
  * [x] Collection detail versions
* [x] Test Signature download on collection detail
* [x] Test Sync signed only on remote collection edit
* [x] Test (and probably update) the described configuration here: https://galaxyng.netlify.app/config/options/
  * [x] Requires first [AAH-1622](https://issues.redhat.com/browse/AAH-1622) 
* [x] Set up CI backend signing service

## On hold because:
* [x] Feature flags are changing https://github.com/ansible/ansible-hub-ui/pull/2211
* [x] Additional work on the UI for uploading signature https://github.com/ansible/ansible-hub-ui/pull/2231

---
* [x] If #2068 merged and galaxykit gets an upload multiple versions support (https://github.com/ansible/galaxykit/pull/44) then this PR could include those tests too.
* [x] Wait for https://github.com/ansible/galaxykit/pull/45 so we can make sure that the staging collections can be removed safely too.

PR #2379, opened 2022-07-27
Revert "cypress: pip install ansible"

himdel:
Reverts ansible/ansible-hub-ui#2355

No longer needed, https://github.com/ansible/galaxykit/pull/55 removes the ansible dependency from galaxykit again :).

Cc @hendersonreed , thanks :)

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -26,7 +26,7 @@ jobs:
 
     - name: "Install galaxykit dependency"
       run: |
-        pip install ansible git+https://github.com/ansible/galaxykit.git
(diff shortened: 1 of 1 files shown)


hendersonreed:
@himdel do we think that the failing cypress tests in `signing.js` are unrelated to galaxykit? If they're unrelated, then we can merge straightaway, I say.

himdel:
Thanks, yeah, those are probably coming from the interaction of https://github.com/ansible/ansible-hub-ui/pull/2069 with https://github.com/ansible/galaxy_ng/pull/1351, and definitely happening on master as well :)

PR #2411, opened 2022-08-01
Error messages aren't displaying on the remote container create form

MilanPospisil:
https://issues.redhat.com/browse/AAH-1845

Messages were in fact displayed, but at the bottom of the form and many users dont have the bottom displayed without scrolling. Those errors could be easily missed.

This PR added the error message right under the name field.
And it also repairs another mistake - errors were not cleared when modal closed (on success or cancel).

Issue: AAH-1845

### DIFF
--- a/src/components/execution-environment/repository-form.tsx
+++ b/src/components/execution-environment/repository-form.tsx
@@ -175,6 +175,7 @@ export class RepositoryForm extends React.Component<IProps, IState> {
                   isDisabled={true}
                   type='text'
                 />
+                {this.renderNameError(formError)}
               </FormGroup>
 
               <FormGroup
@@ -207,6 +208,7 @@ export class RepositoryForm extends React.Component<IProps, IState> {
                   onChange={(value) => this.setState({ name: value })}
                   validated={this.validateName(name)}
                 />
+                {this.renderNameError(formError)}
               </FormGroup>
 
               <FormGroup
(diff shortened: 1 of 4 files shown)


src/components/execution-environment/repository-form.tsx
```diff
@@ -581,4 +583,18 @@ export class RepositoryForm extends React.Component<IProps, IState> {
...
   }
+
+  private renderNameError(formError) {
+    const nameError = formError
+      ? formError.find((e) => e.field == 'name')
+      : null;
+
+    return (
+      nameError && (
+        <Alert title={nameError.title} variant='danger' isInline>
```
himdel, line 594: so, these are eseentially validation errors, right?

We're usually using code like

```
<FormGroup
  ...
  validated={this.toError(!('token' in errorMessages))}
  helperTextInvalid={errorMessages['token']}
>
  <TextInput
    ...
   validated={this.toError(!('token' in errorMessages))}
```

with `errorMessages: ErrorMessagesType` that comes from `mapErrorMessages(error)`.

`formErrors` is used here instead, that's probably fine (for now; also used in user-form, but nowhere else) but we should still be using `helperTextInvalid` instead of an inline alert, unless there's an actual error

(this is also missing a screenshot, so I'm only guessing these are validation errors)

src/components/execution-environment/repository-form.tsx
```diff
@@ -207,6 +208,7 @@ export class RepositoryForm extends React.Component<IProps, IState> {
                   onChange={(value) => this.setState({ name: value })}
                   validated={this.validateName(name)}
                 />
+                {this.renderNameError(formError)}
```
newswangerd, line 211: This should check for errors on other fields too, not just "name"

src/components/execution-environment/repository-form.tsx
MilanPospisil: Ok, I will try to use your way.

himdel:
(replaced with https://github.com/ansible/ansible-hub-ui/pull/2601)

PR #2414, opened 2022-08-03
Remove counter from collection signing UI

ShaiahWren:
Issue: AAH-1852

Connected to https://issues.redhat.com/browse/AAH-1794
Counters for signed and unsigned collections, as well as filters to search partially signed collections, have been removed.
These are the steps for signing collections:

1. Feed your local system with some collections
2. Add to your .compose.env the settings: 
ENABLE_SIGNING=1
PULP_GALAXY_COLLECTION_SIGNING_SERVICE='ansible-default'
PULP_GALAXY_REQUIRE_CONTENT_APPROVAL=true
PULP_GALAXY_AUTO_SIGN_COLLECTIONS=true
3. Run: ./compose exec api ./entrypoint.sh manage create-test-collections --strategy=faux --ns=1 --cols=10 --prefix=signing
4. Then you can go to approval dashboard and approve & sign some collections
5. browse http://0.0.0.0:8002/ui/repo/published and http://0.0.0.0:8002/ui/repo/staging and check if the badges are showing ok

Before:
Screenshot_2022-08-02_19-22-54
Screenshot_2022-08-02_19-27-30

After:
Screen Shot 2022-08-09 at 4 41 34 PM
Screen Shot 2022-08-08 at 1 58 13 PM

### DIFF
--- a/src/components/signing/sign-all-certificates-modal.tsx
+++ b/src/components/signing/sign-all-certificates-modal.tsx
@@ -1,6 +1,5 @@
 import { t, Trans } from '@lingui/macro';
 import {
-  Badge,
   Button,
   ButtonVariant,
   Form,
@@ -27,8 +26,6 @@ interface Props {
 
 export const SignAllCertificatesModal: React.FC<Props> = ({
   name,
-  numberOfAffected,
-  affectedUnsigned,
   isOpen,
   onSubmit,
   onCancel,
@@ -57,8 +54,8 @@ export const SignAllCertificatesModal: React.FC<Props> = ({
         <GridItem span={12}>
           <p>
             <Trans>
-              You are about to sign <strong>{numberOfAffected}</strong>{' '}
-              version(s) under <strong>{name}</strong>.
+              You are about to sign <strong>all versions</strong> under{' '}
+              <strong>{name}</strong>.
             </Trans>
           </p>
         </GridItem>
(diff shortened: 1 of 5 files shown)


CHANGES/1852.bug
```diff
@@ -0,0 +1 @@
+Remove signature counter and partially signed filter from UI.
```
newswangerd, line 1: lets chance this to .misc.

src/components/signing/sign-all-certificates-modal.tsx
```diff
@@ -27,16 +24,14 @@ interface Props {
...
-  affectedUnsigned,
   isOpen,
   onSubmit,
   onCancel,
 }) => {
   return (
     <Modal
       variant={ModalVariant.medium}
-      title={t`Sign all collections`}
+      title={t`Sign collection`}
```
newswangerd, line 34: If I'm not mistaken, this is still going to sign all the collections, not the latest version (right @rochacbruno ?)

The only thing is we no longer get an easy count of the number of signed and unsigned collections here. You might be able to make an api request to a different set of apis to get this information. If that's not possible, then we should adjust the wording to reflect that you're still signing all the collections, but remove any reference to how many versions are getting signed.

newswangerd (changes requested):

src/components/signing/sign-all-certificates-modal.tsx
rochacbruno: YEs if it is on the namespace view it is signing all collections under the namespace

newswangerd (approved):

PR #2415, opened 2022-08-05
Create test for repository management

MilanPospisil:
https://issues.redhat.com/browse/AAH-626

Collections -> Repository management

Test:

    Both tabs contains default repos
    An user sees buttons in kebab according to RBAC rights

Two users - admin and no privilleges user are tested.
Admin sees all buttons, other user none of them.

Additionaly - command for delete test groups was enhanced - it now loops 4 times to delete at least 40 groups.

### DIFF
--- a/CHANGES/626.misc
+++ b/CHANGES/626.misc
@@ -1,0 +1,1 @@
+More test for repository management (collections). An user sees buttons in kebab according to RBAC rights. 
--- a/test/cypress/e2e/repo_management.js
+++ b/test/cypress/e2e/repo_management.js
@@ -26,8 +26,9 @@ describe('Repo Management tests', () => {
 
   it('admin user sees download_concurrency in remote config', () => {
     cy.visit(remoteRepoUrl);
-    cy.contains('.title-box', 'Repo Management');
+    cy.contains('.body', 'community'); // without this, sporadic failures
     cy.get('[aria-label="Actions"]:first').click(); // click the kebab menu on the 'community' repo
+    cy.contains('Edit'); // without this, sporadic failures
     cy.contains('Edit').click();
     cy.contains('Show advanced options').click();
     cy.get('#download_concurrency').should('exist');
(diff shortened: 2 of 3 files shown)


test/cypress/integration/repo_management.js
```diff
@@ -2,13 +2,33 @@ describe('Repo Management tests', () => {
...
 
+  let noPrivilegesUser0 = 'noPrivilegesUser0';
+  //let somePrivilegesUser0 = 'somePrivilegesUser0';
+
+  before(() => {
+    cy.deleteTestGroups();
+    cy.deleteTestUsers();
+    cy.galaxykit('user create', noPrivilegesUser0, noPrivilegesUser0);
+
+    /*cy.galaxykit('user create', somePrivilegesUser0, somePrivilegesUser0);
```
himdel, line 13: lots of commented out code here, why?

Should this be removed, or uncommented?

test/cypress/integration/repo_management.js
```diff
@@ -2,13 +2,33 @@ describe('Repo Management tests', () => {
...
+
   beforeEach(() => {
     cy.login();
   });
 
   it('admin user sees download_concurrency in remote config', () => {
     cy.visit(remoteRepoUrl);
+    cy.contains('.body', 'community'); // without this, sporadic failures
     cy.get('[aria-label="Actions"]:first').click(); // click the kebab menu on the 'community' repo
+    cy.contains('Edit'); // without this, sporadic failures
```
himdel, line 31: this is identical to the next line .. so unlikely to be doing anything

himdel:
missing PR description?

test/cypress/integration/repo_management.js
MilanPospisil: Removed, I forgot this. I tried to create user with some privilleges between None and All, but it seems that there is nothing between.

test/cypress/integration/repo_management.js
MilanPospisil: Yeah, I think this is some magical waiting that prevents sporadic failure, this should be better engineered...

newswangerd (approved):

PR #2245, opened 2022-06-25
Update to cypress 10

himdel:
Replaces https://github.com/ansible/ansible-hub-ui/pull/2171

Updating cypress to 10.6 ([changelog](https://docs.cypress.io/guides/references/changelog))
the main change is tests move from `cypress/integration/` to `cypress/e2e/` and slightly different syntax for config files.

PR #2432, opened 2022-08-08
Create workflow for distributed tests

MilanPospisil:
Issue: AAH-1858


himdel:
Missing description, and a couple of yaml validation errors ;)

Otherwise, looks like this was surprisingly straightforward, nice job :+1: 

I'm not happy about having to remove dirs just to skip tests, but if we can get #2245 merged first, that part can go away in favor of something like `sed -i '/specPattern/s/\*\*/${{ matrix.text }}/' test/cypress.config.js`.

himdel:
Replaced by #2474

PR #2449, opened 2022-08-15
Collection dependencies from community are broken links

MilanPospisil:
https://issues.redhat.com/browse/AAH-1574

The problem was, that context repo was used in link path to collection, which was wrong.
Now, collections in dependencies list are loaded and info about their repo is used in link href path construction.

You can fill test data using:
./compose exec api django-admin create-test-collections --strategy=faux --ns=1 --cols=6 --prefix=deptest

This PR also includes fix of error, that was also present at master before this PR (but the situation cant be reproduced in UI, only by "hijacking" the URL. When you are for example on published repo and you will try to access collection detail in different repo (manualy by inserting URL), GUI will redirect you to Not Found Page (althrough API do returns the collection detail). This is also fixed in this PR, because user can now navigate to different repos, when there is collection in dependency lists with different repo than selected.

### DIFF
--- a/src/components/collection-dependencies-list/collection-dependencies-list.tsx
+++ b/src/components/collection-dependencies-list/collection-dependencies-list.tsx
@@ -4,7 +4,7 @@ import { Link } from 'react-router-dom';
 
 import { List, ListItem, ListVariant } from '@patternfly/react-core';
 
-import { EmptyStateNoData } from 'src/components';
+import { EmptyStateNoData, HelperText } from 'src/components';
 
 import { CollectionDetailType } from 'src/api';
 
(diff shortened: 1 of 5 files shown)


src/containers/collection-detail/collection-dependencies.tsx
```diff
@@ -178,7 +182,64 @@ class CollectionDependencies extends React.Component<
...
+  }
+
+  private loadCollectionsDependenciesRepos(callback) {
+    const dependencies =
+      this.state.collection.latest_version.metadata.dependencies;
+    const dependencies_repos = [];
+    const promises = [];
+
+    Object.keys(dependencies).forEach((dependency) => {
+      const [namespace, collection] = String(dependency).split('.');
```
himdel, line 199: surprising use of String, would `dependency.split('.')` work? (Since these are key names, I don't see how it could not be a string.)

src/containers/collection-detail/collection-dependencies.tsx
```diff
@@ -178,7 +182,64 @@ class CollectionDependencies extends React.Component<
...
+          },
+          this.separateVersion(
+            dependencies[
+              dependency_repo.namespace + '.' + dependency_repo.name
+            ],
+          ),
+        );
+      })
+      .catch(() => {
+        // do nothing, dependency_repo.path stays empty
```
himdel, line 241: also `dependency_repo.repo`?

src/components/collection-dependencies-list/collection-dependencies-list.tsx
```diff
@@ -4,22 +4,20 @@ import { Link } from 'react-router-dom';
...
 
 import { CollectionDetailType } from 'src/api';
-import { formatPath, Paths } from 'src/paths';
 
 import 'src/containers/collection-detail/collection-dependencies.scss';
 
 interface IProps {
   collection: CollectionDetailType;
-  repo: string;
+  dependencies_repos: any;
```
himdel, line 15: looks like this should be an array of something .. is this just `CollectionVersion[]`?

src/containers/collection-detail/collection-dependencies.tsx
MilanPospisil: Ah, yes, it is not path but repo, you are right. I will fix the comment.

himdel (approved):

PR #2239, opened 2022-06-23
Feature/rbac roles

himdel:
Opening a PR from the upstream [feature/rbac-roles](https://github.com/ansible/ansible-hub-ui/commits/feature/rbac-roles) branch, to have a convenient way of seeing the change as a whole...

This contains all the PRs merged against the rbac branch (https://github.com/ansible/ansible-hub-ui/pulls?q=is%3Apr+base%3Afeature%2Frbac-roles+is%3Aclosed):

*  EE edit - only make the Owners tab link clickable when editing (#2376)
*  Remove unexpected digest value (#2375)
*  Add and fix permissions related to roles and group/roles (#2337)
*  group-list: filter "Group name" by name__icontains, not exact name (#2336)
*  cypress: install galaxykit branch with roles (#2268)
*  Fix additional perms (#2267)
*  Fix roles and roles/create screens for user with group perms (#2266)
*  Move rbac components to src/components/rbac/ (#2247)
*  Test Namespace/EE owners (#2230)
*  Add permission checks to group (#2210)
*  Filter roles for applicability to current namespace/EE (#2186)
*  RBAC Test rework (#2184)
*  Fix redirection to not existing group (#2118)
*  Group roles test (#2117)
*  Role lists - default sort by name, show description, pretty "Galaxy only" in AppliedFilters (#2092)
*  Optimize assigned roles to group loading and fix bugs (#2086)
*  test add roles to group (#1995)
*  Rbac testing (#1993)
*  GroupRoleAPI, RoleAPI, PulpAPI - move logic to components (#1981)
*  Drop all ObjectPermissionField logic (#1980)
*  Fix CI, unify breadcrumb types (#1978)
*  New UI for object permission assignment (#1863)
*  Role list view (#1860)
*  Rework groups view (#1797)

PR #2451, opened 2022-08-15
Fix Owners tab permissions

himdel:
https://github.com/ansible/ansible-hub-ui/pull/2378 got closed because the base branch went away during the rbac merge (#2239)

Reopening against master :)

---

Fixes AAH-1875

Set owners tab to use the permissions returned by my_permissions on container and collection namespaces, or user permissions.

Requires https://github.com/ansible/galaxy_ng/pull/1389 (merged)

the view for assigning roles to containers should use `change_containernamespace`,
the view for assigning roles to collection namespaces should use `change_namespace`.

### DIFF
--- a/CHANGES/1875.bug
+++ b/CHANGES/1875.bug
@@ -1,0 +1,1 @@
+Fix Owners tab permissions
--- a/src/api/namespace.ts
+++ b/src/api/namespace.ts
@@ -2,6 +2,10 @@ import { HubAPI } from './hub';
 
 class API extends HubAPI {
   apiPath = this.getUIPath('namespaces/');
+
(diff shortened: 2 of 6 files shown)


himdel:
(`group_roles` and `rbac_owners` are also red on master .. and both passing locally .. so if only those two fail, the issue should be  unrelated)

himdel:
Cc @newswangerd @awcrosby this should be ready

awcrosby (approved):
It seems this is checking for object or model permissions, so that looks good.

Not sure about the Cypress tests failures, I'll defer on that. Otherwise LGTM.

PR #2456, opened 2022-08-17
galaxykit: use main branch, instead of feature branch

himdel:
now that rbac is merged, we can use galaxykit main again :).

(#2239 & ansible/galaxykit#39)


himdel:
Ugh, merging right away, since merging the galaxkit PR removed the original branch

PR #2474, opened 2022-08-22
Distributed workflows for github.

MilanPospisil:
Issue: AAH-1858

Matrix was created - so cypress.yml will run multiple times, each time with different variable (defined in matrix array). Tests are organized into folders - folder names match the matrix array. Each test runs only tests in its folder.


himdel:
There are still a couple of yaml validation errors ;)

And we don't have to remove dirs just to skip tests any more, you can use something like `sed -i '/specPattern/s/\*\*/${{ matrix.text }}/' test/cypress.config.js` to make cypress run just the right folder.

Aand I think we talked about testing if the matrix has all the dirs as well? `diff -Naur <(ls test/cypress/integration) <(yq '.jobs.cypress.strategy.matrix.test[]' .github/workflows/cypress-distributed.yml | sort)` works for me ([yq](https://github.com/mikefarah/yq))

As for why everything fails .. there's a typo in the matrix list .. evnironments->environments .. probably a test failure cancels the other jobs.

MilanPospisil:
> 

Not yet, thats why its still WIP :-)

test/cypress.config.js
```diff
@@ -8,6 +8,6 @@ module.exports = defineConfig({
       return require('./cypress/plugins/console-logger').install(on);
     },
     baseUrl: 'http://localhost:8002',
-    specPattern: 'cypress/e2e/**/*.js',
+    specPattern: 'cypress/e2e/matrix_folder_to_change/*.js',
```
himdel, line 11: this would break tests everywhere except on CI

change this back, and match on `specPattern:` for the substitution instead?

himdel:
(Test failures unrelated, caused by https://github.com/ansible/galaxykit/pull/71#discussion_r972006872)

test/cypress.config.js
MilanPospisil: Ah, you are right :-(

.github/workflows/cypress.yml
```diff
@@ -208,9 +221,19 @@ jobs:
...
+      working-directory: 'ansible-hub-ui'
+      run: |
+        diff -Naur <(ls test/cypress/e2e) <(yq '.jobs.cypress.strategy.matrix.test[]' .github/workflows/cypress.yml | sort)
+
     - name: "Run cypress"
       working-directory: 'ansible-hub-ui/test'
       run: |
+        dir=${{matrix.test}}
+        pattern="s/\*\*/$dir/"
+        sed -i "$pattern" cypress.config.js
```
himdel, line 234:
```suggestion
        sed -i '/specPattern:/s/\*\*/${{matrix.test}}/' cypress.config.js
```

`**` may be present on more than just this line, let's match by the key name first

himdel (approved):
:+1: 

I'd still like to see the `specPattern:` match just to be sure,
but this is ready and working :+1: 

(restarted tests to see if working galaxykit takes care of the remaining red tests)

PR #2479, opened 2022-08-24
Release 4.5.2

newswangerd:
We're re-releasing 4.5.1 as 4.5.2 because of a dependency issue with the backend.


ironfroggy (approved):

CHANGES.rst
```diff
@@ -27,6 +29,8 @@ Misc
 
 ----
```
ShaiahWren, line 30: Is this extra file necessary? I think it's created with the `sed` command

ShaiahWren (approved):

PR #1969, opened 2022-04-25
[stable-4.5] Cypress: install galaxykit from pypi, not github

himdel:
the idea being a stable branch should use a released version of galaxykit

4.2: no `cypress.yml`
4.3: `==0.1.0` since #520
4.4: #1968, `==0.7.0`
4.5: here, `==0.8.0` (https://github.com/ansible/galaxykit/pull/60)
master: `git+https://github.com/ansible/galaxykit.git` since #1343

---

this may not work yet? not sure if we'll need to release whatever is not released in galaxykit, draft

PR #2498, opened 2022-08-29
[stable-4.5] cypress.yml - backport changes from master (#2480)

himdel:
Backports cypress.yml changes from #2480, #2478, #2240, #1979,
and the whole #2164 (because galaxykit-importer was updated to 0.4.5 in 4.5 as well).

The remaining difference between 4.5 and master `cypress.yml` are..

* #1969 
  * galaxykit 0.8.0 vs main
  * galaxykit wait_for_task patch
* #2265 
  * waits for pulp to start responding
  * galaxykit queries namespaces not collections
* build from `ghcr.io/pulp/pulp-ci-centos` version `3.18` (not `latest`)
* install `openldap-devel` before `pip install` .. `python-ldap` fails to install otherwise on 3.18
* pulpPrefix without API_BASE_PATH, that's 4.6+


himdel:
Cc @ironfroggy - this ensures 4.5 tests on github run against current 4.5 (with pulpcore 3.18, not 3.20), instead of June 4.5

PR #2499, opened 2022-08-29
add community Dockerfile

drodowic:
Add Dockerfile for community galaxy_ng instance

Issue: AAH-1818

### DIFF
--- a/community/Dockerfile
+++ b/community/Dockerfile
@@ -1,0 +1,11 @@
+FROM node:16-alpine
+WORKDIR /workspace/
+RUN mkdir -p /workspace/ && \
+    apk add --no-cache git
+COPY . /workspace/
+RUN npm ci && \
+    npm run build-community
+
+FROM registry.access.redhat.com/ubi8/nginx-120:latest
+COPY --from=0 /workspace/dist ${NGINX_APP_ROOT}/src
+CMD ["nginx", "-g", "daemon off;"]
--- a/config/community.prod.webpack.config.js
+++ b/config/community.prod.webpack.config.js
@@ -1,0 +1,16 @@
+const webpackBase = require('./webpack.base.config');
+
+// Compile configuration for stnadalone mode
+module.exports = webpackBase({
+  API_HOST: '',
+  API_BASE_PATH: '/api/',
+  PULP_API_BASE_PATH: '/api/pulp/api/v3/',
+  UI_BASE_PATH: '/ui/',
+  DEPLOYMENT_MODE: 'standalone',
+  NAMESPACE_TERM: 'namespaces',
+  UI_USE_HTTPS: false,
+  UI_DEBUG: false,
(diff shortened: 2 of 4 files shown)


community/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM node:16-alpine
+WORKDIR /workspace/
+RUN mkdir -p /workspace/ && \
+    apk add --no-cache git
+COPY . /workspace/
+RUN sed -i 's/login/login\/github\//' config/standalone.prod.webpack.config.js && \
+    sed -i 's/static\/galaxy_ng\///' config/standalone.prod.webpack.config.js && \
+    sed -i 's/api\/galaxy/api/' config/standalone.prod.webpack.config.js && \
```
himdel, line 8: can we just create a `config/community.prod.webpack.config.js`, and a corresponding `build-community` task?

community/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+    npm ci && \
```
himdel, line 9: (`npm ci` is not really suitable for development .. I'd suggest `npm install`, but maybe it's ok for the container and people can `npm install ; npm run start-community` manually when desired?)

community/Dockerfile
drodowic: I created `config/community.prod.webpack.config.js`, and the `build-community` task

community/Dockerfile
drodowic: This mulitstage Dockerfile creates the static assets and copies them for Nginx to serve. The resulting image won't have node installed and won't be able to run `npm run`. 
For this use case would `npm install` or `npm ci` be better?

community/Dockerfile
himdel: Thanks :) :+1:

community/Dockerfile
himdel: The main difference is that `npm ci` will fail if you update `package.json` (without updating `package-lock.json`), whereas `npm install` will install things accordingly.

So.. if we're only using this for testing changes done in standalone, this should work just fine.
If you want to use it for community-only features and have dependencies, it could become cumbersome.

I guess we can keep `npm ci` and address it if it ever becomes a problem :)

himdel (approved):
LGTM :)

(Ignore the `webpack-config` CI failure, it's just detecting build config change.)

config/community.prod.webpack.config.js
```diff
@@ -0,0 +1,16 @@
+const webpackBase = require('./webpack.base.config');
+
+// Compile configuration for stnadalone mode
+module.exports = webpackBase({
+  API_HOST: '',
+  API_BASE_PATH: '/api/',
+  PULP_API_BASE_PATH: '/api/pulp/api/v3/',
```
himdel, line 7: (we can drop the separate PULP_API_BASE_PATH line after #2501, but it's harmless)

PR #1857, opened 2022-04-04
PULP_API_BASE_PATH: add an env var to override the `/pulp/api/v3/` path

himdel:
and also respect API_HOST in PulpAPI, same as HubAPI

Related to AAH-1332 (#1589) and AAH-1367 - both will need to access pulp API endpoints in insights mode.

PR #2501, opened 2022-08-29
PULP_API_BASE_PATH - hardcode to use API_BASE_PATH + 'pulp/api/v3/'

himdel:
related to packaging!11 (closed), #2135, https://github.com/ansible/galaxy_ng/pull/1247
mostly undoes #1857 (+packaging!10) config changes but not code changes

`packaging/roles/publish/files/config/downstream.webpack.config.js` can't easily have version-dependent logic,
so having individual webpack configs provide `PULP_API_BASE_PATH` doesn't quite work,
but having the constant available in code is still useful.

=> ignoring any PULP_API_BASE_PATH config options or env variables, hardcoding PULP_API_BASE_PATH global to be based on API_BASE_PATH in 4.6+
(for 4.5 and older, the url is just /pulp/api/v3/, no PULP_API_BASE_PATH, no changes needed)

### DIFF
--- a/config/webpack.base.config.js
+++ b/config/webpack.base.config.js
@@ -27,7 +27,6 @@ const defaultConfigs = [
   // as a constant after it is compiled
   { name: 'API_HOST', default: '', scope: 'global' },
   { name: 'API_BASE_PATH', default: '', scope: 'global' },
-  { name: 'PULP_API_BASE_PATH', default: '', scope: 'global' },
   { name: 'UI_BASE_PATH', default: '', scope: 'global' },
   { name: 'DEPLOYMENT_MODE', default: 'standalone', scope: 'global' },
   { name: 'NAMESPACE_TERM', default: 'namespaces', scope: 'global' },
@@ -64,6 +63,11 @@ module.exports = (inputConfigs) => {
       );
     }
   });
+
+  // 4.6+: pulp APIs live under API_BASE_PATH now, ignore previous overrides
+  globals.PULP_API_BASE_PATH = JSON.stringify(
+    customConfigs.API_BASE_PATH + 'pulp/api/v3/',
+  );
 
   const isStandalone = customConfigs.DEPLOYMENT_MODE !== 'insights';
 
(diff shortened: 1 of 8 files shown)


himdel:
(`webpack-config` failure expected, this is technically changing the config, even if it's just reordering the globals)

config/webpack.base.config.js
```diff
@@ -64,6 +63,11 @@ module.exports = (inputConfigs) => {
     }
   });
 
+  // 4.6+: pulp APIs live under API_BASE_PATH now, ignore previous overrides
+  globals.PULP_API_BASE_PATH = JSON.stringify(
+    customConfigs.API_BASE_PATH + 'pulp/api/v3/',
```
newswangerd, line 68: why not just set this here https://github.com/ansible/ansible-hub-ui/blob/master/src/api/pulp.ts#L7 instead of using a global var?

newswangerd, line 68: why not just set this here https://github.com/ansible/ansible-hub-ui/blob/master/src/api/pulp.ts#L7 instead of using a global var?

himdel:
```
src/api/pulp.ts
7:    super(API_HOST + PULP_API_BASE_PATH);

src/api/role.ts
19:      newParams.for_object_type = PULP_API_BASE_PATH + for_object_type + '/';

src/components/headers/collection-header.tsx
617:    const signed_collection = `${PULP_API_BASE_PATH}content/ansible/collection_versions/${version.id}/`;

src/containers/certification-dashboard/certification-dashboard.tsx
535:    const signed_collection = `${PULP_API_BASE_PATH}content/ansible/collection_versions/${version.id}/`;

src/containers/task-management/task_detail.tsx
457:            const url = resource.replace(PULP_API_BASE_PATH, '');
```

himdel: So unless we want to update all those places..?

newswangerd: ah, I didn't realized we used this config in so many places.

newswangerd (approved):

PR #2502, opened 2022-08-30
Signing error alerts - switch title and description

himdel:
API Error details should be in the description, while the notification title should say which action failed

Cc @MilanPospisil


MilanPospisil (approved):
LGTM 👍

PR #2503, opened 2022-08-30
test test breakage

himdel:
Trying to figure out if the current cypress breakage (https://github.com/ansible/ansible-hub-ui/runs/8098212627?check_suite_focus=true) is caused by galaxy_ng#1428 or a new pulp-ci-centos


himdel:
Does NOT need a backport for now, looks like the 3.15 and 3.18 images still use the old s6 service paths.

himdel:
Ok, does seem to be needed now for 4.4: https://github.com/ansible/ansible-hub-ui/runs/8114735243?check_suite_focus=true
But not yet for 4.5 (https://github.com/ansible/ansible-hub-ui/runs/8138567958?check_suite_focus=true)

PR #2308, opened 2022-07-05
Drop master-stable, deploy stage-stable from master conditionally

himdel:
Follow-up to https://github.com/ansible/ansible-hub-ui/pull/1770

Before:
* cloud-stage-cron periodically pushes master to master-stable when $flag
* master-stable push triggers stage-stable deploy (but the cron one actually doesn't because of actions not triggering actions)
* master push triggers stage-beta deploy
* cloud-stage-{enable,disable} set $flag

After:
* master push triggers stage-beta deploy
* master push ALSO triggers stage-stable deploy *when $flag*
* cloud-stage-{enable,disable} set $flag
(no master-stable branch, no cloud-stage-cron action)

This means stage-stable updates together with stage-beta unless paused,
but it also means that it won't update when unpaused until a new PR is merged (but `deploy-cloud` can still be triggered manually)

PR #2522, opened 2022-09-01
deploy-cloud, base config: add explicit deployment config to build stage-stable and stage-beta both from master

himdel:
`npm run deploy` reads the current branch name (in cloud mode) and decides to build a beta version when the branch is `master`
but since #2308, we're deploying to stage-stable as well, leading to issues on stage-stable:

    Loading chunk 849 failed. (error: https://console.stage.redhat.com/beta/apps/automation-hub/js/849.1661953046888.a80f84b3e77abc4bde6d.js)

=> adding a `HUB_CLOUD_BETA` env var to override the autodetection,
and building with `HUB_CLOUD_BETA=true` for stage-beta, `HUB_CLOUD_BETA=false` for stage-stable,
and no override for prod builds, as before.

(and moving the `npm run deploy` call to `custom_release.sh` so we can run it twice from master)

### DIFF
--- a/.github/workflows/deploy-cloud.yml
+++ b/.github/workflows/deploy-cloud.yml
@@ -58,4 +58,4 @@ jobs:
       run: |
         npm cache verify
         npm ci || npm install
-        npm run deploy && curl -sSL https://raw.githubusercontent.com/RedHatInsights/insights-frontend-builder-common/master/src/bootstrap.sh | bash -s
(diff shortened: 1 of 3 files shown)


himdel:
Cc @awcrosby @Hyperkid123 

This should fix stage-stable deploys.. I hope :)

awcrosby (approved):

PR #2567, opened 2022-09-12
Expose signing service public keys

himdel:
Fixes AAH-1826

This is adding a Signing Keys screen, both in standalone and insights mode.

---

The last commit changes to exact name filtering only, and disables sorting options, as the API currently doesn't support those. We can revert that part once support is added.

PR #2568, opened 2022-09-12
Simplify toolbar css

himdel:
Many containers use a div class wrapper around toolbar to set display flex, justify content, align items
some also set inner toolbar content padding
(and search+namespaces use a different structure not addressed here)

This is a minimal change to dedup these into `hub-toolbar` (doesn't reset padding) and `hub-list-toolbar` (does).
(I suspect we can merge the two as well, but not enough to do it right away)

(Pulled out of #2567.)

|from|to|
|-|-|
|`detail-images-toolbar`|`hub-toolbar`|
|`hub-certification-dashboard-toolbar`|`hub-toolbar`|
|`hub-usedby-dependencies-header`|`hub-toolbar`|
|`hub-container-list-toolbar`|`hub-list-toolbar`|
|`hub-group-list-toolbar`|`hub-list-toolbar`|
|`hub-task-list`|`hub-list-toolbar`|
|`hub-user-list-toolbar`|`hub-list-toolbar`|
|`role-list`|`hub-list-toolbar`|

### DIFF
--- a/src/components/collection-dependencies-list/collection-usedby-dependencies-list.tsx
+++ b/src/components/collection-dependencies-list/collection-usedby-dependencies-list.tsx
@@ -63,7 +63,7 @@ export class CollectionUsedbyDependenciesList extends React.Component<IProps> {
 
     return (
       <>
-        <div className='hub-usedby-dependencies-header'>
+        <div className='hub-toolbar'>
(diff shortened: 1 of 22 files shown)


MilanPospisil (approved):
LGTM 👍

PR #2570, opened 2022-09-13
EE list: filter by name__icontains, not name exact

himdel:
Fixes: AAH-1913

Sort by name__icontains instead of exact name.

Before:

After:


MilanPospisil (approved):
LGTM 👍

PR #2572, opened 2022-09-14
Owners tab: go up to group list when clicking the tab

himdel:
Fixes: AAH-1733

when opening a group detail under an owners tab (EE or namespace),
it's not immediately obvious how to get back to the list of groups

there's no explicit back button, only the browser back button and breadcrumbs

this also makes it so clicking the owners tab again goes back to groups :)

### DIFF
--- a/src/containers/execution-environment-detail/base.tsx
+++ b/src/containers/execution-environment-detail/base.tsx
@@ -72,6 +72,13 @@ export function withContainerRepo(WrappedComponent) {
       this.loadRepo();
     }
 
+    componentDidUpdate() {
+      // when reloading the same tab, state doesn't reset
+      if (this.state.redirect) {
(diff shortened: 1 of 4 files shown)


src/containers/execution-environment-detail/base.tsx
```diff
@@ -92,6 +92,9 @@ export function withContainerRepo(WrappedComponent) {
       }[this.state.redirect];
 
       if (redirect) {
+        // when reloading the same tab, state doesn't reset
+        this.setState({ redirect: null });
```
himdel, line 96: react_devtools_backend.js:4026 Warning: Cannot update during an existing state transition (such as within `render`). Render methods should be a pure function of props and state

MilanPospisil (approved):
LGTM 👍

PR #2573, opened 2022-09-14
Ensure sorting, filtering, and resetting filters resets to page 1

himdel:
Issue: AAH-1848

moving the `page: 1` reset from individual containers to the shared `AppliedFilters`, `CompoundFilter`, `SortTable` and `Sort` components.

This should ensure filtering and sorting always resets to page one, while pagination still works.
(Perpage change already sets page 1 everywhere.)


MilanPospisil (approved):
LGTM 👍

PR #2574, opened 2022-09-16
Add backport labels to PRs

awcrosby:
Ensure PR with changes to app functionality gets backport labels to be looked at by PR author and reviewer to keep or remove

Related: https://github.com/ansible/galaxy_ng/pull/1449

No-Issue

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -10,3 +10,7 @@ backport-4.4:
 
 backport-4.5:
 - src/**/*
+
+backport-4.6:
+- src/**/*
+- test/**/*
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,12 @@
+# Ensure PR with changes to app functionality gets backport labels
+# to be looked at by PR author and reviewer to keep or remove
+# called by .github/workflows/labeler.yml
+
+backport-4.2:
+- src/**/*
+
+backport-4.4:
+- src/**/*
+
+backport-4.5:
+- src/**/*
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -1,11 +1,12 @@
-# Add labels to pull request
+# Add labels to pull requests against master
 # uses logic in .github/labeler.yml
 
 ---
 name: "Add labels to pull request"
 
 on:
-- pull_request_target
+  pull_request_target:
+    branches: [ "master" ]
 
 jobs:
   triage:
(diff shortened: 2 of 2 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,12 @@
+# Ensure PR with changes to app functionality gets backport labels
+# to be looked at by PR author and reviewer to keep or remove
+# called by .github/workflows/labeler.yml
+
+backport-4.2:
+- src/**/*
```
awcrosby, line 6: I assume functionality changes exist under `src/` but please review and let me know if I should include additional or different path(s)

himdel: `src` sounds reasonable :+1: 

(it might make sense to add `test` as well, except 4.2 doesn't run them, 4.4 and 4.5 use different cypress versions (differences in naming) and rbac, so maybe 4.6+)

.github/workflows/labeler.yml
```diff
@@ -0,0 +1,19 @@
+# Add labels to pull request
+# uses logic in .github/labeler.yml
+
+---
+name: "Add labels to pull request"
+
+on:
+- pull_request_target
```
himdel, line 8:
```suggestion
  pull_request_target:
    branches: [ 'master' ]
```

so that backport PRs (manual or patchback) don't trigger labelling

.github/workflows/labeler.yml
```diff
@@ -0,0 +1,19 @@
...
+
+jobs:
+  triage:
+    permissions:
+      contents: read
+      pull-requests: write
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/labeler@v4
```
himdel, line 17: we may also need to skip dependabot PRs, as those are done per-branch.

I'm guessing... ?

```suggestion
    - uses: actions/labeler@v4
      if: ${{ github.event.pull_request.user.login }} != 'dependabot[bot]'
```

himdel:
(cypress failure unrelated (https://github.com/ansible/galaxykit/pull/71/files#r972006872))

.github/labeler.yml
```diff
@@ -0,0 +1,12 @@
...
+
+backport-4.4:
+- src/**/*
+
+backport-4.5:
+- src/**/*
```
himdel, line 12:
```suggestion
- src/**/*

backport-4.6:
- src/**/*
```

(stable-4.6 was cut since you started this PR)

himdel:
One more comment.. 2.2 is entering a "critical fixes only" maintenance period .. so not sure if we should be adding `stable-4.2` and `stable-4.4` automatically at all?

.github/workflows/labeler.yml
awcrosby: This syntax worked in my testing, thanks!

awcrosby:
@himdel those older branches likely won't get many backports, but could get a security backport. Let's keep them for now and see later if we want to remove them.

himdel (approved):
Thanks, LGTM :+1: :)

PR #2575, opened 2022-09-17
Automate cloud menu updates

himdel:
Cloud menu cleanup, TODO try to automate menu changes
(moved out of #2567 as unrelated)

also TODO: there's also deploy/frontend.yaml with its own copy of the menu


himdel:
As per https://github.com/RedHatInsights/chrome-service-backend/blob/main/docs/cloud-services-config.md,
this will all move to a `frontend.yml` file in this repo with the container switch. Closing.

PR #2599, opened 2022-09-21
Add a start script for the community profile.

jctanner:
Need a way to dev & debug the local ui stack in community mode.


himdel (approved):

himdel:
The linter fails on the new config file, otherwise LGTM :)

(the webpack-config fail is expected as this is an intentional config change)

PR #2601, opened 2022-09-22
Error messages aren't displaying on the remote container create form

MilanPospisil:
https://issues.redhat.com/browse/AAH-1845

This PR refactors error handling using utils. Every error (including field validation and server side validation together) is now in one associative array formErrors, which can hold error message per form field (for example formErrors['name']). 

Also, validation from server in attempt to save the form is now included.

### DIFF
--- a/src/components/execution-environment/repository-form.tsx
+++ b/src/components/execution-environment/repository-form.tsx
@@ -15,7 +15,13 @@ import {
 } from '@patternfly/react-core';
 import { ExternalLinkAltIcon, TagIcon } from '@patternfly/react-icons';
 import { Link } from 'react-router-dom';
-import { AlertType, APISearchTypeAhead, HelperText } from 'src/components';
+import {
+  AlertType,
+  APISearchTypeAhead,
+  HelperText,
+  AlertList,
+  closeAlertMixin,
+} from 'src/components';
 import {
   ContainerDistributionAPI,
   ExecutionEnvironmentRegistryAPI,
@@ -25,8 +31,10 @@ import { Paths, formatPath } from 'src/paths';
 import {
   errorMessage,
   ErrorMessagesType,
-  mapErrorMessages,
   isFieldValid,
+  isFormValid,
+  mapErrorMessages,
+  alertErrorsWithoutFields,
 } from 'src/utilities';
 
 interface IProps {
@@ -47,13 +55,12 @@ interface IProps {
   upstreamName?: string;
   remoteId?: string;
   addAlert?: (variant, title, description?) => void;
-  formError: { title: string; detail: string }[];
 }
 
 interface IState {
   name: string;
   description: string;
-
+  alerts: AlertType[];
   addTagsInclude: string;
   addTagsExclude: string;
   excludeTags?: string[];
@@ -61,9 +68,6 @@ interface IState {
   registries?: { id: string; name: string }[];
   registrySelection?: { id: string; name: string }[];
   upstreamName: string;
-  /*formErrors?: {
-    registries?: AlertType;
-  };*/
   formErrors: ErrorMessagesType;
 }
 
@@ -81,10 +85,8 @@ export class RepositoryForm extends React.Component<IProps, IState> {
       registries: null,
       registrySelection: [],
       upstreamName: this.props.upstreamName || '',
-      /*formErrors: {
-        registries: null,
-      },*/
       formErrors: {},
+      alerts: [],
     };
   }
 
@@ -101,10 +103,17 @@ export class RepositoryForm extends React.Component<IProps, IState> {
             });
           }
         })
-        .catch((res) => {
-          debugger;
-          const error = mapErrorMessages(res);
-          this.setState({ formErrors: error });
+        .catch((e) => {
+          const { status, statusText } = e.response;
+          const errorTitle = t`Registries list could not be displayed.`;
+          this.addAlert({
+            variant: 'danger',
+            title: errorTitle,
+            description: errorMessage(status, statusText),
+          });
+          this.setState({
+            formErrors: { ...this.state.formErrors, registries: errorTitle },
+          });
         });
     }
   }
@@ -123,8 +132,6 @@ export class RepositoryForm extends React.Component<IProps, IState> {
       addTagsExclude,
       formErrors,
     } = this.state;
-
-    this.validateName(this.state.name);
 
     return (
       <Modal
@@ -148,6 +155,10 @@ export class RepositoryForm extends React.Component<IProps, IState> {
           </Button>,
         ]}
       >
+        <AlertList
+          alerts={this.state.alerts}
+          closeAlert={(i) => this.closeAlert(i)}
+        />
         <Form>
           {!isRemote ? (
             <>
@@ -187,7 +198,10 @@ export class RepositoryForm extends React.Component<IProps, IState> {
                   id='name'
                   value={name}
                   isDisabled={!isNew}
-                  onChange={(value) => this.setState({ name: value })}
+                  onChange={(value) => {
+                    this.setState({ name: value });
+                    this.validateName(value);
+                  }}
                   validated={isFieldValid(this.state.formErrors, 'name')}
                 />
               </FormGroup>
@@ -216,10 +230,16 @@ export class RepositoryForm extends React.Component<IProps, IState> {
                 label={t`Registry`}
                 className='hub-formgroup-registry'
                 isRequired={true}
+                helperTextInvalid={
+                  this.state.formErrors['registries'] ||
+                  this.state.formErrors['registry']
+                }
+                validated={isFieldValid(this.state.formErrors, [
+                  'registries',
+                  'registry',
+                ])}
               >
-                {formErrors?.registries ? (
-                  <></>
-                ) : (
+                {!formErrors?.registries && (
                   <>
                     {registries ? (
                       <APISearchTypeAhead
@@ -230,6 +250,7 @@ export class RepositoryForm extends React.Component<IProps, IState> {
                             registrySelection: registries.filter(
                               ({ name }) => name === value,
                             ),
+                            formErrors: { ...formErrors, registry: null },
                           })
                         }
                         placeholderText={t`Select a registry`}
@@ -241,37 +262,6 @@ export class RepositoryForm extends React.Component<IProps, IState> {
                     )}
                   </>
                 )}
-
-                {/* formErrors?.registries ? (
-                  <Alert
-                    title={formErrors.registries.title}
-                    variant='danger'
-                    isInline
-                  >
-                    {formErrors.registries.description}
-                  </Alert>
-                ) : (
-                  <>
-                    {registries ? (
-                      <APISearchTypeAhead
-                        loadResults={(name) => this.loadRegistries(name)}
-                        onClear={() => this.setState({ registrySelection: [] })}
-                        onSelect={(event, value) =>
-                          this.setState({
-                            registrySelection: registries.filter(
-                              ({ name }) => name === value,
-                            ),
-                          })
-                        }
-                        placeholderText={t`Select a registry`}
-                        results={registries}
-                        selections={registrySelection}
-                      />
-                    ) : (
-                      <Spinner />
-                    )}
-                  </>
-                    )*/}
               </FormGroup>
 
               <FormGroup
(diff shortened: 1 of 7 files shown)


himdel:
There's a lot of commented-out code, and no  PR description.

What is this supposed to do?

src/containers/execution-environment-list/execution_environment_list.tsx
```diff
@@ -467,15 +467,8 @@ class ExecutionEnvironmentList extends React.Component<
...
-            .catch((err) => {
-              this.setState({
-                formError: err.response.data.errors.map((error) => {
-                  return {
-                    title: error.title,
-                    detail: error.source.parameter + ': ' + error.detail,
-                  };
-                }),
-              });
+            .catch(() => {
```
himdel, line 470: we're not returning the promise anywhere, so an empty catch is not needed?

src/utilities/map-error-messages.ts
```diff
@@ -26,3 +26,54 @@ export function mapErrorMessages(err): ErrorMessagesType {
...
+
+  let valid = true;
+  // if any key contains error text inside, its invalid
+  Object.keys(errorMessagesType).forEach((error) => {
+    if (error) {
+      valid = false;
+    }
+  });
+  return valid;
+}
```
himdel, line 60: so, `ErrorMessagesType` is the type, that sounds ok, but why is the variable called `errorMessagesType`? This is just the regular `errorMessages` field from the containers, right? Can we call it `errorMessages`?

src/components/execution-environment/repository-form.tsx
```diff
@@ -462,6 +470,18 @@ export class RepositoryForm extends React.Component<IProps, IState> {
...
         exclude_tags,
+      }).catch((err) => {
+        /*const { status, statusText } = err.response;
+        const errorTitle = t`Name already exists.`;
+        this.addAlert({
+          variant: 'danger',
+          title: errorTitle,
+          description: errorMessage(status, statusText),
+        });*/
+
```
himdel, line 481:
```suggestion
        /*const { status, statusText } = err.response;
        const errorTitle = t`Name already exists.`;
        this.addAlert({
          variant: 'danger',
          title: errorTitle,
          description: errorMessage(status, statusText),
        });*/

```

src/utilities/map-error-messages.ts
```diff
@@ -26,3 +26,54 @@ export function mapErrorMessages(err): ErrorMessagesType {
...
+
+// this function suppose that state contains formErrors variable, if not, it will not work
+export function setError(state, setState, field, error) {
+  // if error is already present, dont set value again
+  if (state.formErrors[field] == error) {
+    return;
+  }
+
+  state.formErrors[field] = error;
+  setState({ formErrors: state.formErrors });
```
himdel, line 70: The inplace editing of formErrors in state right before calling setState feels wrong,
and not resetting the field sounds suspicious. 

Why are we not creating a new `formErrors` object here?

src/utilities/map-error-messages.ts
```diff
@@ -26,3 +26,54 @@ export function mapErrorMessages(err): ErrorMessagesType {
...
+}
+
+export function mapErrorsAndSet(state, setState, err) {
+  const errObj = mapErrorMessages(err);
+
+  Object.keys(errObj).forEach((key) => {
+    setError(state, setState, key, errObj[key]);
+  });
```
himdel, line 78: isn't this exactly what `mapErrorMessages` does by itself?

This whole thing seems unnecessary when the caller can just do `this.setState({ formErrors: mapErrorMessages(err) })`

src/utilities/map-error-messages.ts
```diff
@@ -26,3 +26,54 @@ export function mapErrorMessages(err): ErrorMessagesType {
...
+  }
+
+  let valid = true;
+  // if any key contains error text inside, its invalid
+  Object.keys(errorMessagesType).forEach((error) => {
+    if (error) {
+      valid = false;
+    }
+  });
+  return valid;
```
himdel, line 59: this is just `return Object.keys(errorMessagesType).length > 0;` right?

himdel: (unless you meant to look at `formErrors[key]` instead of the field names?)

src/utilities/map-error-messages.ts
MilanPospisil: Yes, that was only my laziness :-D But with class components, there must be quite a lot params passed to function, so this not saves coding too much...

src/utilities/map-error-messages.ts
MilanPospisil: Thanks for catching that, I forgot that forEach returns keys and not the values.

src/containers/execution-environment-list/execution_environment_list.tsx
MilanPospisil: I now catched the second promise that contains 'return Promise.all([...'. So this outside catch is now not needed.

src/containers/execution-environment-list/execution_environment_list.tsx
```diff
@@ -443,50 +443,36 @@ class ExecutionEnvironmentList extends React.Component<
...
-                      </Trans>
-                    ),
-                  }),
-                },
-                () => this.queryEnvironments(),
-              );
-            })
-            .catch((err) => {
-              this.setState({
-                formError: err.response.data.errors.map((error) => {
```
himdel, line 472: `formError` is dead in EE list now, can you remove the rest?

src/components/execution-environment/repository-form.tsx
```diff
@@ -215,16 +228,10 @@ export class RepositoryForm extends React.Component<IProps, IState> {
...
-                {formErrors?.registries ? (
-                  <Alert
-                    title={formErrors.registries.title}
-                    variant='danger'
-                    isInline
-                  >
-                    {formErrors.registries.description}
-                  </Alert>
-                ) : (
+                {!formErrors?.registry && (
```
himdel, line 234: `registry` vs `registries` was a meaningful difference

the field should NOT be invalid when `formErrors.registries`, you just can't choose a differernt registry
and vice versa, when a validation error for `registry` comes from the API, the field should be invalid but you need to be able to chose a different registry

src/utilities/map-error-messages.ts
```diff
@@ -26,3 +26,36 @@ export function mapErrorMessages(err): ErrorMessagesType {
...
+
+  let valid = true;
+  // if any key contains error text inside, its invalid
+  Object.keys(errorMessages).forEach((field) => {
+    if (errorMessages[field]) {
+      valid = false;
+    }
+  });
+  return valid;
+}
```
himdel, line 61: dead code

src/components/execution-environment/repository-form.tsx
himdel: So... now I'm not seeing any references to `formErrors.registry` .. what happened there?

We should be showing errors from both sources.

himdel:
One more thing noticed during testing... this doesn't deal with unknown errors from the server in any way, no alerts, no messages, just a stuck form.

Try adding a EE with a name like `foo/.` - this passes the name validation, but the server responds with a `base_path` validation message .. we should be displaying unknown messages on error.

(The response looks like...
```
{
  "errors": [
    {
      "status": "400",
      "code": "invalid",
      "title": "Invalid input.",
      "detail": "The provided base path contains forbidden characters.",
      "source": {
        "parameter": "base_path"
      }
    }
  ]
}
```
)

MilanPospisil:
TODO - delete validation errors when alert is shown and error does not belong to any field.

src/containers/execution-environment-list/execution_environment_list.tsx
```diff
@@ -448,9 +446,8 @@ class ExecutionEnvironmentList extends React.Component<
         excludeTags={exclude_tags || []}
         includeTags={include_tags || []}
         permissions={namespace?.my_permissions || []}
-        remoteId={remoteId}
+        remotePulpId={remoteId}
```
himdel, line 449: bad rebase? There's no reason to rename the prop in this PR, right?

src/containers/execution-environment-list/execution_environment_list.tsx
```diff
@@ -476,15 +473,8 @@ class ExecutionEnvironmentList extends React.Component<
...
-                formError: err.response.data.errors.map((error) => {
-                  return {
-                    title: error.title,
-                    detail: error.source.parameter + ': ' + error.detail,
-                  };
-                }),
-              });
+            .catch((e) => {
+              console.log(e);
             });
```
himdel, line 478: Empty catch with forgotten console.log?

```suggestion
```

src/containers/execution-environment-detail/base.tsx
```diff
@@ -251,17 +248,9 @@ export function withContainerRepo(WrappedComponent) {
...
-                            title: error.title,
-                            detail:
-                              error.source.parameter + ': ' + error.detail,
-                          };
-                        }),
-                      }),
-                    );
+                    .catch((e) => {
+                      console.log(e);
+                    });
```
himdel, line 253:
```suggestion
                    .catch((e) => {
                      console.log(e);
                    });
```

src/utilities/map-error-messages.ts
```diff
@@ -26,3 +26,72 @@ export function mapErrorMessages(err): ErrorMessagesType {
...
+    if (!fields.includes(field) && errorMessages[field]) {
+      errors += errorMessages[field] + ' ';
+      errorMessages[field] = '';
+    }
+  });
+
+  if (errors) {
+    addAlert({ variant: 'danger', title: title, description: errors });
+  }
+  return;
```
himdel, line 96:
```suggestion
  const errors = Object.keys(errorMessages)
    .filter((field) => !fields.includes(field))
    .map((field) => errorMessages[field]);
    
  if (errors.length) {
    addAlert({ variant: 'danger', title: title, description: errors.join("\n") });
  }
```

src/utilities/map-error-messages.ts
```diff
@@ -26,3 +26,72 @@ export function mapErrorMessages(err): ErrorMessagesType {
...
+  }
+
+  let valid = true;
+  // if any key contains error text inside, its invalid
+  Object.keys(errorMessages).forEach((field) => {
+    if (errorMessages[field]) {
+      valid = false;
+    }
+  });
+  return valid;
```
himdel, line 68:
```suggestion
  return !Object.values(errorMessages).find(Boolean);
```

src/utilities/map-error-messages.ts
```diff
@@ -26,3 +26,72 @@ export function mapErrorMessages(err): ErrorMessagesType {
...
+  }
+
+  let res = 'default';
+  names.forEach((n) => {
+    if (errorMessagesType[n]) {
+      res = 'error';
+    }
+  });
+
+  return res;
```
himdel, line 49:
```suggestion
  return names.find((n) => errorMessagesType[n]) ? 'error' : 'default';
```

src/components/execution-environment/repository-form.tsx
```diff
@@ -179,15 +191,18 @@ export class RepositoryForm extends React.Component<IProps, IState> {
...
               >
                 <TextInput
                   id='name'
                   value={name}
                   isDisabled={!isNew}
-                  onChange={(value) => this.setState({ name: value })}
-                  validated={this.validateName(name)}
+                  onChange={(value) => {
+                    this.setState({ name: value });
+                    this.validateName(value);
```
himdel, line 203: shouldn't this also be unsetting `formErrors.base_path` (as per https://github.com/ansible/ansible-hub-ui/pull/2601#issuecomment-1273714360)?

src/containers/execution-environment-list/execution_environment_list.tsx
MilanPospisil: All errors are handled in previous promises, this would fail on unexpected error (in setState), so I will show something like Unexpected error alert with error message.

MilanPospisil: So, this is problem. This catch pops up whenever form is invalid, so it does really nothing, because those errors are displayed in the RepositoryForm component. With alert message, the alert message is displayed whenever forms throw an error (aditionaly to its own error handling) and shows alert behind the form modal, which is bad. I will delete the catch for now and we can discuss it later.

src/components/execution-environment/repository-form.tsx
```diff
@@ -40,25 +53,22 @@ interface IProps {
   includeTags?: string[];
   registry?: string; // pk
   upstreamName?: string;
-  remoteId?: string;
   addAlert?: (variant, title, description?) => void;
-  formError: { title: string; detail: string }[];
+  remotePulpId?: string;
```
himdel, line 57: please don't rename unrealted props  here .. `remoteId` is right

src/utilities/map-error-messages.ts
```diff
@@ -26,3 +26,65 @@ export function mapErrorMessages(err): ErrorMessagesType {
 
   return messages;
 }
+
+export function isFieldValid(errorMessagesType: ErrorMessagesType, name): any {
```
himdel, line 30:
```suggestion
export function isFieldValid(errorMessagesType: ErrorMessagesType, name): 'default' | 'error' {
```

himdel:
Looks like this works :+1: 







(just needs a rebase to get tests green, #3018 )

himdel (approved):
(once tests go green)

PR #2602, opened 2022-09-22
deploy-cloud: allow custom manual deploys given a target branch name, source commit ref, and a bool for /beta/ vs /

himdel:
The intent is to be able to deploy a custom commit to a cloud branch, so we can notify users of any upcoming changes.

This updates the workflow_dispatch event to take user inputs specifying the source, taget and wheteher to build for /beta/ urls or not.

In the event of a custom run, `TRAVIS_BRANCH` is set to `custom` (used by the `custom_release.sh` script only (except when set to `"nightly"` which we don't)), `BRANCH` is set to the target branch name in case there's any webpack logic still looking there,  and `CUSTOM_BRANCH` is the name of the target branch used by the `release.sh` script.

### DIFF
--- a/.github/workflows/deploy-cloud.yml
+++ b/.github/workflows/deploy-cloud.yml
@@ -14,10 +14,10 @@ on:
         required: true
         type: choice
         options:
-          - prod-beta
-          - prod-stable
-          - qa-beta
-          - qa-stable
+        - prod-beta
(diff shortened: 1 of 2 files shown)


himdel:
Cc @awcrosby this should allow us to deploy any banners outside the standard process if needed :)

PR #2605, opened 2022-09-23
Use `/ui/:repo/:namespace` paths for namespace

jerabekjiri:
Issue: [AAH-1945](https://issues.redhat.com/browse/AAH-1945)

Filtering by repository doesn't work if user is on `/ui/namespaces/<namespace>` or `/ui/my-namespaces/<namespace>`. We should use `/ui/:repo/:namespace` paths for namespaces instead.

User was redirected to the wrong path after:
- creating namespace
- editing the namespace
- breadcrumbs

- I added a test `Filter by repository` to test correct functionality (could get handy in the future with new repo management features)

### DIFF
--- a/src/components/namespace-form/namespace-form.tsx
+++ b/src/components/namespace-form/namespace-form.tsx
@@ -20,6 +20,7 @@ import { NamespaceType } from 'src/api';
 import { NamespaceCard } from 'src/components';
 import { Paths, formatPath } from 'src/paths';
 import { ErrorMessagesType, validateURLHelper } from 'src/utilities';
+import { AppContext } from 'src/loaders/app-context';
 
 interface IProps {
   namespace: NamespaceType;
@@ -29,6 +30,8 @@ interface IProps {
 }
 
 export class NamespaceForm extends React.Component<IProps> {
+  static contextType = AppContext;
+
   render() {
     const { namespace, errorMessages } = this.props;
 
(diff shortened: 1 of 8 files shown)


src/components/namespace-form/namespace-form.tsx
```diff
@@ -82,8 +82,9 @@ export class NamespaceForm extends React.Component<IProps> {
                 <Link
                   target='_blank'
                   to={formatPath(
-                    Paths.myCollections,
+                    Paths.namespaceByRepo,
                     {
+                      repo: 'published',
```
himdel, line 87: should this alos use `this.context.selectedRepo`?

src/containers/namespace-list/namespace-list.tsx
```diff
@@ -165,8 +165,9 @@ export class NamespaceList extends React.Component<IProps, IState> {
           onCreateSuccess={(result) =>
             this.setState({
               redirect: formatPath(
-                Paths.myCollections,
+                Paths.namespaceByRepo,
                 {
+                  repo: 'published',
```
himdel, line 170: (and here?)

src/components/namespace-form/namespace-form.tsx
jerabekjiri: you're right, I thought we could only edit the published repo, but I was wrong.

src/containers/namespace-list/namespace-list.tsx
jerabekjiri: I think after creating a new namespace, user should be redirected to the published repo in UI. I don't think we should redirect to the previous or latest selected repo.

himdel: Aah, makes sense, works for me :+1: :)

himdel (approved):
LGTM :+1: :)

(I just restarted tests, there was an imports failure, but looks unrelated .. I think :))

PR #2623, opened 2022-09-26
hasPermission - add a context helper to check user permissions

himdel:
Cc @ShaiahWren 

Something like this should help split up the https://github.com/ansible/galaxy_ng/pull/1430 work into 2 phases:

1. refactor UI to use the `hasPermission` function for every `model_permissions` check
  * (possibly adding a temporary `const oldNames = { "new_name": "old_name" }; name = oldNames[name];` mapping)
  * merge, part by part, without any api dependencies
2. change `hasPermission` to consume the format returned by https://github.com/ansible/galaxy_ng/pull/1430 instead of the original
  * merge together with the API change

### DIFF
--- a/src/containers/group-management/group-detail.tsx
+++ b/src/containers/group-management/group-detail.tsx
@@ -124,11 +124,9 @@ class GroupDetail extends React.Component<RouteComponentProps, IState> {
   }
 
   componentDidMount() {
-    if (
-      !this.context.user ||
-      this.context.user.is_anonymous ||
-      !this.context.user.model_permissions.view_group
-    ) {
+    const { user, hasPermission } = this.context;
+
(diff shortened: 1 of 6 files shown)


himdel:
So in step 1, we'd be changing `this.context.user.model_permissions.move_collection` to `this.context.hasPermission('ansible.modify_ansible_repo_content')`,
but it'd be the role of `hasPermission` to know that `move_collection` and `ansible.modify_ansible_repo_content` are the same thing.

So step 1 changes all the UI but doesn't depend on the API.

Step 2 would just switch the API implementation, forget the old names, and not touch any other UI code.

himdel:
Done in https://github.com/ansible/ansible-hub-ui/pull/2682, closing this one :)

PR #2660, opened 2022-09-29
Automerge test for dependabots PRs

MilanPospisil:
https://issues.redhat.com/browse/AAH-1959

This PR solves one par - the result is either 0 (pass) or 1 (fail) if all PRs finished with ok status and if dependabot PR can be merged by automaticaly.


.github/workflows/automerge.js
```diff
@@ -0,0 +1,137 @@
...
+{
+    console.log('Actor argument (third) was not specified');
+    return;
+}
+
+console.log(branch);
+console.log(prName);
+console.log(actor);
+
+if (actor != 'dependabot')
```
himdel, line 33: the user should be called `dependabot[bot]`

(also.. you can add `if: ${{ github.event.pull_request.user.login == 'dependabot[bot]' }}` to any action in the workflow, possibly to the whole workflow as well)

.github/workflows/automerge.js
```diff
@@ -0,0 +1,137 @@
...
+if (prName.includes('patternfly'))
+{
+    console.log('Automerge cant merge patternfly Prs.');
+    return;
+}
+
+if (prName.includes('types/node'))
+{
+    console.log('Checking for types node version.');
+    let pattern = /from 16[.]\d\d[.]\d\d to 16[.]\d\d[.]\d\d/;
```
himdel, line 48:
```suggestion
    let pattern = /from 16[.]\d+[.]\d+ to 16[.]\d+[.]\d+/;
```

(because `\d\d` wont' match 16.0.1 or 16.17.189)

.github/workflows/automerge.js
```diff
@@ -0,0 +1,137 @@
...
+    }else
+    {
+        console.log('Version does not match the pattern ' + pattern);
+        return;
+    }
+
+
+}
+
+const { exec } = require("child_process");
```
himdel, line 62: `child_process` is a node api, not an npm package

so, this works, but the new convention (since node16) is 
```suggestion
const { exec } = require("node:child_process");
```

package.json
```diff
@@ -17,11 +17,13 @@
         "@types/react-dom": "^16.9.15",
         "@types/react-router-dom": "^5.3.3",
         "axios": "~0.27.2",
+        "child_processes": "^0.1.0",
```
himdel, line 20: and we should remove this one, `child_processes` != `child_process`

https://nodejs.org/api/child_process.html - the node api you're using
https://www.npmjs.com/package/child_process - a placeholder package because you're not the first one to get there :)
https://www.npmjs.com/package/child_processes - unrelated, apparently harmless

.github/workflows/automerge.yml
```diff
@@ -0,0 +1,37 @@
...
+
+    - name: "Install node 16"
+      uses: actions/setup-node@v2
+      with:
+        node-version: '16'
+
+    - name: "install libraries for node"
+      run: |
+        npm install child_processes
+
```
himdel, line 30: and here
```suggestion
```

.github/workflows/automerge.js
```diff
@@ -0,0 +1,137 @@
...
+    return new Promise( (resolve, reject) => {
+        exec(`gh run list -b "${branch}" --json name,status,conclusion,event `, (error, stdout, stderr) => {
+            if (error) {
+                console.log(`error: ${error.message}`);
+                reject(error);
+                return;
+            }
+            if (stderr) {
+                console.log(`stderr: ${stderr}`);
+                reject(error);
```
himdel, line 75:
```suggestion
                reject(stderr);
```
?

(error will be empty here)

.github/workflows/automerge.js
```diff
@@ -0,0 +1,137 @@
...
+
+
+}
+
+const { exec } = require("child_process");
+
+function loadLatestRuns(branch)
+{
+    return new Promise( (resolve, reject) => {
+        exec(`gh run list -b "${branch}" --json name,status,conclusion,event `, (error, stdout, stderr) => {
```
himdel, line 67: up to you if you want to change it or not, but you were looking for a wait to run exec with promise, right?

```
const { promisify } = require('node:util');
const { exec: origExec } = require('node:child_process');
const exec = promisify(origExec);

exec(`command`).then(({ stdout, stderr }) => ..., (error) => ...)
```

https://nodejs.org/docs/latest-v18.x/api/util.html#utilpromisifyoriginal

himdel:
yamllint complains about the final whitespace,
and we should probably run prettier for these scripts as well,
but otherwise looks good, see comments above :)

himdel:
Oh, feel free to address this in a separate PR if you like ,but ... we also need to filter the runs by `headSha`...

the API returns all the runs, not just the one for the last commit .. so we need to look at the branch head sha and compare it to the `headSha` field

.github/workflows/automerge.js
```diff
@@ -0,0 +1,137 @@
...
+if (process.argv.length >= 3) { branch = process.argv[2]; }
+if (process.argv.length >= 4) { prName = process.argv[3]; }
+if (process.argv.length >= 5) { actor = process.argv[4]; }
+
+console.log('--------------------------------------------------------');
+
+if (!branch)
+{
+    console.log('Branch name argument (first) was not specified');
+    return;
```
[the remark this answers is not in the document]
MilanPospisil: Oh, you are right. Im not used to return nonzero value as success :-D

.github/workflows/automerge.js
MilanPospisil: Oh, ok, I though that versiosns are strictly defined as 2 digits.

.github/workflows/automerge.js
MilanPospisil: Hey, promisify is so cool!

.github/workflows/automerge.js
MilanPospisil: But I rather have the all rules in one place in the js file for clarity.

.github/workflows/automerge.js
MilanPospisil: Hm, I see, that was probably situation, when I had opened only dependabots PRs with two digits, today, I have many single digits :)

package.json
MilanPospisil: Oh, little confusing...

himdel (approved):
Looks ready to be tested against real dependabot PRs before we add any actual merging logic :+1:

PR #2682, opened 2022-10-03
Create permissions function

ShaiahWren:
Issue: AAH-1714

See https://github.com/ansible/ansible-hub-ui/pull/2623

This pr starts the first part of the Permissions UI refactor. 

The `model_permissions` references in `standalone-loader.tsx` I have kept - they probably need to go as well?

I also wasn't sure what to touch in the `constants.ts` file.

Thanks in advance for a review!

### DIFF
--- a/CHANGES/1714.feature
+++ b/CHANGES/1714.feature
@@ -1,0 +1,1 @@
+Utilize hasPermission function to accomodate new permission names from the API.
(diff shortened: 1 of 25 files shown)


himdel:
Pretty sure this is missing the part where you rename the permissions to use the new name instead of the existing name?
Or do you want to do this in a separate step?

himdel (approved):
Looks good, I just added AppContext to RemoteRepositoryTable which was missing it, if the tests pass now LGTM :+1: :)

(except for the repo test which is likely still failing)

PR #2686, opened 2022-10-06
The ui for dependencies doesn't show collection namespace

MilanPospisil:
https://issues.redhat.com/browse/AAH-1807

Added namespace to the link caption. Also namespace is also added to used by dependencies links captions.


himdel (approved):
LGTM :+1:

PR #2712, opened 2022-10-13
Path in the collection dependencies does not contains / at the end of the url

MilanPospisil:
Issue: AAH-1973

https://issues.redhat.com/browse/AAH-1973

### DIFF
--- a/CHANGES/1973.bug
+++ b/CHANGES/1973.bug
@@ -1,0 +1,1 @@
+Slash at the end of the url of dependencies of collections. 
--- a/src/paths.ts
+++ b/src/paths.ts
@@ -6,7 +6,7 @@ export function formatPath(
   data,
   params?: Record<string, string | boolean>,
 ) {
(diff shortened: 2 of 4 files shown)


himdel:
(the issue is insights-only, removing backport labels)

himdel:
:+1: definitely works, but if we can get away with `formatPath` always returning a trailing slash, that'd be a better solution, as none of these urls are generated with a trailing slash by default. Looking into whather that is something we can do.. :hourglass:

MilanPospisil:
> 👍 definitely works, but if we can get away with `formatPath` always returning a trailing slash (or always in insights mode), that'd be a better solution, as none of these urls are generated with a trailing slash by default. Looking into whather that is something we can do.. ⌛

That was my first try, but it broke the tests. So it needs more work.

himdel (approved):
LGTM :+1: 

(assuming tests pass :))

PR #2713, opened 2022-10-13
Collection dependencies - link has to check also collection version

MilanPospisil:
Issue: AAH-1972

https://issues.redhat.com/browse/AAH-1972

### DIFF
--- a/CHANGES/1972.bug
+++ b/CHANGES/1972.bug
@@ -1,0 +1,1 @@
+Collection dependencies - link has to check also collection version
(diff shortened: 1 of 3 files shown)


himdel:
OK, I think we can merge this one and wait for development in https://issues.redhat.com/browse/AAH-1968.

This PR does ensure that we don't generate 404ing dependency links by checking for the existence of the same version we'd be linking to :+1: 

Here's the problem - we're getting the version from the dependent's `metadata.dependencies` field, and parsing out the first thing that looks like a version. So, given `>=2.4.0,!=2.6.0` as input, we'll only test for 2.4.0, instead of accepting 2.4.7, 2.5.0 or 2.6.8 as well. There's no API *yet* for that (see AAH-1968), so unless we want to parse ansible version strings in JS and compare that with a full list of versions (or give up no versions altogether), we need to wait for that. But now we wait without 404s :).

himdel (approved):

PR #2740, opened 2022-10-18
Insights tests

himdel:
Issue: AAH-1271 (first part)

Add `insights` to cypress test matrix,
Run pulp-galaxy-ng in insights mode (same container, different settings), but on port 5001, not using static assets,
together with `npm run start` for dev mode insights UI, serving UI on 8002, proxying api to 5001.
Also move cypress env, pulp settings and pulp-galaxy-ng Containerfile to files from the workflow, and parametrize api path and port in workflow

~~galaxykit unable to authenticate (it can with ansible/galaxykit#79, or https://github.com/RedHatInsights/frontend-components/pull/1659)~~

TODO for sharing test code (separate PRs):
/ui/ vs /beta/ansible/automation-hub/ - need uiRoot setting or cypress basePath?
lot of /ui/ refs in tests .. also different urls .. change to menus where doable, logic to rest


himdel:
:tada: 
Collections list -- lists collections -- before each hook
Collections list -- lists collections -- before each hook (1)

himdel:
Cc @newswangerd can you review?

This enables us to write tests against an insights env running on CI.
(I'll add the actual tests in a separate PR.)

I originally wanted to replace this with the oci-env based version and close, but this one works while the other one still needs work to let standalone tests pass (#2769) .. so might as well start writing tests with against this one, and switch when the other one is ready, WDYT?

.github/workflows/cypress/settings.py.standalone
```diff
@@ -0,0 +1,14 @@
+ANSIBLE_API_HOSTNAME='http://localhost:8002'
+ANSIBLE_CONTENT_HOSTNAME='http://localhost:8002/api/galaxy/v3/artifacts/collections'
+CONTENT_ORIGIN='http://localhost:8002'
+GALAXY_API_PATH_PREFIX='/api/galaxy/'
+GALAXY_AUTHENTICATION_CLASSES=['rest_framework.authentication.SessionAuthentication', 'rest_framework.authentication.TokenAuthentication']
```
newswangerd, line 5: Don't override the auth classes for standalone. We want to test the defaults (which should work)

newswangerd (approved):
Looks good to me. I'll let you know when I get oci-env running with the sync tests.

.github/workflows/cypress/settings.py.standalone
himdel: Updated :) .. it seems to default to `['rest_framework.authentication.SessionAuthentication', 'rest_framework.authentication.TokenAuthentication', 'rest_framework.authentication.BasicAuthentication']` ... is that the intended default? (I thought basic auth was supposed to be disabled?)

PR #2741, opened 2022-10-18
update hasPermission function to work with the /me api endpoint refactor

ShaiahWren:
Issue: AAH-1714

To be merged in concert with https://github.com/ansible/galaxy_ng/pull/1430 for completion of the `/me` endpoint refactor. 

This pr simply adjusts the `hasPermission` function to read the `has_model_permission` boolean attribute of a given permission from the newly refactored `/me` endpoint.

### DIFF
--- a/src/utilities/has-permission.tsx
+++ b/src/utilities/has-permission.tsx
@@ -3,29 +3,6 @@ export function hasPermission({ user, settings, featureFlags }, name) {
   if (!user?.model_permissions) {
     return false;
   }
-  const newToOld = {
-    'ansible.modify_ansible_repo_content': 'move_collection',
-    'ansible.delete_collection': 'delete_collection',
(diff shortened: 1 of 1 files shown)


himdel (approved):
LGTM! :)

We could do a bit more cleanup of the old names in src/constants, but it doesn't have to be a part of this PR :)

(This will now throw an exception if the ui tries to call hasPermission with a permission name that the API doesn't know about, instead of just returing false .. but that sounds like a good thing :+1: :))

ShaiahWren:
@himdel I'd be happy to do a little cleanup of the old names in constants - shall I get rid of the commented code?

ShaiahWren:
Closing & merging pr. We have a bit of digging to do to see why the cypress tests are failing, but it's not because of this pr:)

PR #2742, opened 2022-10-19
Screenshots tests

MilanPospisil:
Screenshot tests works like this:

It creates screenshots using cypress test.

It runs either as master (using cron) mode or in pull request mode (thus variable COMPARE SCREENSHOTS true for PR and false for master mode).

 In master mode, it only creates screenshots and cache them.
In PR mode, it compares screenshots that were made on the PR with retrieved master screenshots from cache. 

Then, the test fails if there is some difference (with minimum treshold tolerance) between master and PRs screenshots. Image differences are part of uploaded artifacts data.


himdel:
@MilanPospisil I've cleaned up the whitespace issues, removed some commented out code here, fixed the cache invocation, added support for SHORT_BRANCH other than master, and made the tool keep the visual diffs so we can see them in screenshots when this fails.

I think we could test this on master, WDYT?

himdel:
The history here is broken, there are commits that should not be here doing things different from what they're saying, clean it up please in the next rebase.

And looking at some of the changes we should set a specific goal here, which should IMO be:

* when running on master, run screenshots test to create screenshots, ensure they end up in cache
* when running against a PR, fail on missing cache, run screenshots tests to create screenshots, fail on differences

MilanPospisil:
> 

You mean I should rebase -i master and squash those commits?I dont know why your commits got into this PR. Was that something about rebasing onto your PR?

himdel:
> You mean I should rebase -i master and squash those commits

Yeah, sounds reasonable. Actually reading those commits and removing the ones that are not needes is probably better, but whichever works.

> The goals - I dont think it should fail on non-dependabots PRs, 

The point is to get something done and merged so we can use it. The check failing is the easiest way of doing that, and we can deal with commenting on the PR in a separate PR. If the check doesn't do anything user-visible when there are changes, it's useless.

himdel (approved):
LGTM, not seeing any issues, should work :+1: 

(maybe let's wait for the start of January before merging, in case there are any issues after merge)

himdel:
:tada: worked in https://github.com/ansible/ansible-hub-ui/actions/runs/3819922593/jobs/6498007657 (#3126)

But there are still some changes [visible](https://github.com/ansible/ansible-hub-ui/suites/10135965831/artifacts/494889337)..

```
screenshots.js/__repositories.png
4713.93

screenshots.js/__users.png
1024.67
```

`__repositories`: last updated N minutes ago number mismatch
`__users`: created N minutes ago number mismatch

PR #2743, opened 2022-10-19
[prod-beta] insights - add maintenance notice (2022-11-15)

himdel:
Adds a "Automation Hub has a scheduled downtime between 8am-9am ET on 2022-10-26." banner to the top of all hub insights pages.

Cc @awcrosby @jlmitch5

--- 

EDIT: updated to Nov 15


himdel:
@awcrosby do we want this in (with the correct date) or should I close this in favor of the https://status.redhat.com message?

PR #2744, opened 2022-10-20
insights: use UI_BASE_PATH over computing basename from pathname

himdel:
when clicking a link that goes to /ansible/automation-hub/,
but the current page is /ansible/automation-dashboard/ or /ansible/automation-analytics/ (not hub),
the basename gets computed wrong, and a hub router tries to resolve the path using the wrong basepath,
and ends up parsing /ansible/automation-hub/ as matching Paths.collection (/:namespace/:collection).

Which triggers rendering collection-detail, querying the API with a wrong collection name, and after the 404 a redirect to /ansible/ansible-dashboard/not-found (or analytics).

=> Hardcoding the basename to be either /ansible/automation-hub/ or /beta/ansible/automation-hub/ based on build config, instead of computing it from the pathname.

Issue: [AA-1417](https://issues.redhat.com/browse/AA-1417) (renamed from AAH-1990)

Cc @h-kataria


h-kataria:
Thanks for looking into this @himdel

awcrosby (approved):

PR #2765, opened 2022-10-24
ci: build backend container with UI

himdel:
Master tests are failing (https://github.com/ansible/ansible-hub-ui/actions/runs/3276489348/jobs/5477502564), originally seen in #2741

```
$ echo 'expecting /static/galaxy_ng/js/App.'"$BUILD_HASH"'.js'
$ curl http://localhost:8002/static/galaxy_ng/index.html | tee /dev/stderr | grep '/static/galaxy_ng/js/App.'"$BUILD_HASH"'.js'

expecting /static/galaxy_ng/js/App.1666644881606.ef88a718cd5f9228a206.js
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100   179  100   179    0     0   2671      0 --:--:-- --:--:-- --:--:--  2632
100   179  100   179    0     0   2632      0 --:--:-- --:--:-- --:--:--  2594

<!doctype html>
<html lang="en">
<head>
  <title>Not Found</title>
</head>
<body>
  <h1>Not Found</h1><p>The requested resource was not found on this server.</p>
</body>
</html>

Error: Process completed with exit code 1.
##[debug]Finishing: Ensure index.html uses the new js
```

---

https://github.com/pulp/pulp-oci-images/pull/311 - `collectstatic` no longer happens on api restart, only build time, which is too soon for us
=> moving this to build time as well, losing caching (TODO clean up caching more)

---

Except this doesn't help, there doesn't seem to be a pulp oci container that serves UI now. Will find out if intentional or not.
Opening #2769 as backup.

---

```
django.db.utils.ProgrammingError: relation "galaxy_user" does not exist
LINE 1: ...er"."is_active", "galaxy_user"."date_joined" FROM "galaxy_us...`
```

I guess the container fails to run galaxy migrations.

But that doesn't happen for vanilla pulp/pulp-galaxy-ng , that one just doesn't serve the UI (https://github.com/pulp/pulp-oci-images/issues/340).

Maybe we should just use the container for APIs and take the "static server/proxy outside the container" logic from #2769.
But then again oci-env has signing working out of the box.

PR #2769, opened 2022-10-30
ci: use oci-env as backend

himdel:
Related to #2765.. tries to solve the same "no ui served by the container anymore" issue

But this time, by switching from a manual pulp-oci-images build to pulp/oci-env.

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -37,7 +37,7 @@ jobs:
 
     steps:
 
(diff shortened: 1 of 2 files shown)


himdel:
closing for #2795 
may still be relevant for #2740

himdel:
Will be reopening to switch both standalone and insights over to oci-env.
(profiles are merged now, so TODO drop the extra code pulling profiles)

PR #2770, opened 2022-10-30
ci: serve UI assets separately, use container only for backend

himdel:
Potential alternative to #2769 and #2765, to fix the current CI failures (details in #2765)

#2765 rebuilds backend container after building UI
  * :-1: more container shenanigans
  * :+1: closest to current test env and to production

#2769 switches backend container to oci-env
  * :-1: least prod-like
  * :+1: likely more stable interface, signing out of the box

This is a smaller change than either, use the same container but only for backend
  * :-1: serving UI is less prod-like, won't fix repo signing
  * :+1: minimal change

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -172,7 +172,7 @@ jobs:
         cd www
         echo '{}' > package.json
         npm install local-web-server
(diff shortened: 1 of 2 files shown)


himdel:
Closing for #2795 
May still be relevant for #2740

PR #2771, opened 2022-10-30
Permissions - clean up

himdel:
Follow-up to #2682 and #2741

Clean up old permission names

### DIFF
--- a/src/constants.tsx
+++ b/src/constants.tsx
@@ -26,78 +26,69 @@ export class Constants {
       name: 'namespaces',
       label: defineMessage({ message: `Collection Namespaces` }),
       object_permissions: [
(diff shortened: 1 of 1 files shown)


ShaiahWren (approved):
Looks good! :) Though I'm not seeing a difference in the `galaxy.add_containerregistryremote` name. Is that the one needing a change?

himdel:
Ah, thanks, good catch :) 
The change was originally here, but I moved it to https://github.com/ansible/ansible-hub-ui/pull/2795 to fix all the red tests in one PR - so that's already merged through https://github.com/ansible/ansible-hub-ui/pull/2795/commits/581f44a4646118b1f374bef56239fe061ed0a5fa :)

PR #2822, opened 2022-11-09
cypress: run standalone and insights against oci-env

himdel:
Replaces #2769, follows #2740

Switching CI over from a custom build of https://github.com/pulp/pulp-oci-images to https://github.com/pulp/oci_env
(using `galaxy_ng/base` & `galaxy_ng/insights` profiles)

Also switching away from `cy.settings` that changes settings and restarts API during tests, towards running each config in a different run - matrix dirs can now have a `compose.env` fragment, which gets applied before compose build and compose up.

Updated tests to expect signatures enabled, added a `compose.env` with 
`PULP_GALAXY_REQUIRE_CONTENT_APPROVAL='false'` to test dirs that needed it (the default changes to true with the env switch), fixed missing `uiPrefix` for `cy.visit('/')`, and split/moved tests around to fit the settings.

Also ensures both standalone and insights have a working galaxykit, feature-flags and settings check.


himdel:
Merging to get tests green

PR #2824, opened 2022-11-09
Pushing updated strings for localization

rooftopcellist:

PR #2895, opened 2022-11-16
Update localization strings

himdel:
bringing strings from https://github.com/ansible/ansible-hub-ui/pull/2824 which was a stable-4.6 PR to master

    git checkout upstream/stable-4.6 -- locale/
    npm run gettext:extract
    npm run gettext:compile

(also API: https://github.com/ansible/galaxy_ng/pull/1542)


himdel:
Merging before the autorefresh causes conflicts

PR #2899, opened 2022-11-19
automerge workflow: skip on non-dependabot PRs, fail on bad args, and patternfly

himdel:
The automerge workflow is surprising on PRs not created by dependabot  => adding a step-level condition, as per https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution

And making sure the script fails on missing/empty args, skips patternfly PRs with a failure, and handles single-digit node versions.


himdel:
Cc @MilanPospisil when you're back :)

Merging to test with Monday's dependabot batch.

PR #2900, opened 2022-11-20
Test ansible/galaxykit#81 fix

himdel:
ansible/galaxykit#80 causes `galaxykit container create ...` to error with `Cannot parse expected JSON response (http://localhost:8002/api/galaxy/v3/plugin/execution-environments/remotes/)` (bad url)
https://github.com/ansible/galaxykit/pull/81 should fix that, testing.


himdel:
Merging for now, will revert when galaxykit PR gets merged

PR #2930, opened 2022-11-24
[stable-4.5] Pin pulp-ci-centos to 66dc18a - fix CI container build

himdel:
trying if this fixes 4.5 failure to build container

from ansible/galaxy_ng#1525

test failure https://github.com/ansible/ansible-hub-ui/actions/runs/3536978652/jobs/5936539038


himdel:
Nope, not it. Seems like some version of setuptools is incompatible with some version of pip. We'll see...

PR #2952, opened 2022-11-28
[stable-4.4] ci: override GALAXY_NG_COMMIT to use a cached container that works

himdel:
temporary fix pending a real one or the oci-env switch

(failing in https://github.com/ansible/ansible-hub-ui/pull/2926)


himdel:
Oh never mind, the failure has resolved itself, no more build errors.

PR #2953, opened 2022-11-29
EE tests - switch to quay as source, drop unique suffix

himdel:
Closes #1986 

We may not need to run a separate docker api server to test sync, we're already using quay to pull containers during CI, so we can find testable containers there.

This should fix EE use in controller tests failing when many tests run at the same time.

The smallest container I've found on quay is `quay.io/ansible/docker-test-containers:busybox` (and 
 `:hello-world`) (https://quay.io/repository/ansible/docker-test-containers?tab=tags), using that.

---

Also dropping unique `num` suffix from EE/RR tests - no longer needed with working EE cleanup (AAH-1095)


himdel:
Closing, not needed now, not sure what changed.

PR #2452, opened 2022-08-15
Unify RolePermissions component - replace other versions

himdel:
Replaces https://github.com/ansible/ansible-hub-ui/pull/2246 now that rbac (#2239) is merged.

---

`RolePermissions` comes from d8e4f75446c5f3965ea24c0970c269bec9648311 (ansible/ansible-hub-ui#1863),
taken from `src/containers/group-management/group-detail-role-management/group-detail-role-management.tsx`

But the same logic also appears in:
`src/components/role-management/role-form.tsx`
`src/components/role-management/role-permissions.tsx`
`src/containers/role-management/role-list.tsx`

with variations on editability, showing extra permissions, and hiding empty permission groups.

=> `setPermissions`, `showCustom`, `showEmpty` on `RolePermissions` :)

also TODO ensure changes from https://github.com/ansible/ansible-hub-ui/pull/2228 survive

PR #2955, opened 2022-11-30
Use permissions from api not constants

ShaiahWren:
AAH-2016

Use permissions from API /me endpoint in lieu of using permissions from constants.


src/components/rbac/permission-categories.tsx
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  private getNicenames(permission) {
+    const { model_permissions } = this.context.user;
+    if (model_permissions[permission].name !== undefined) {
+      return model_permissions[permission].name;
+    } else {
+      return undefined;
+    }
+  }
```
himdel, line 121: this doesn't work with `showCustom={true}` .. you can try by switching to true in RoleList and expanding the `galaxy.synclist_owner` role ... it will call `getNicenames('galaxy.add_synclist')` and fail because it's missing from `model_permissions`

I guess the quick fix is

```suggestion
  private getNicenames(permission) {
    const { model_permissions } = this.context.user;
    return model_permissions[permission]?.name || permission;
  }
```

, or we could call nicenames only on known permissions instead, up to you :)

src/containers/role-management/role-list.tsx
```diff
@@ -132,8 +129,6 @@ export class RoleList extends React.Component<RouteComponentProps, IState> {
     const noData =
       roleCount === 0 && !filterIsSet(params, ['name__icontains', 'locked']);
 
-    const groups = Constants.PERMISSIONS;
-
     const { featureFlags } = this.context;
     let isUserMgmtDisabled = false;
     const filteredPermissions = { ...Constants.HUMAN_PERMISSIONS };
```
himdel, line 134: `filteredPermissions` is unused now

And also line 148/143 .. USER_GROUP_MGMT_PERMISSIONS is a list of permissions to hide when in keycloak/insights modes.

Is that already filtered API-side or do we need to add this logic to PermissionCategories? (probably with a `showUserMgmt={!featureFlags.external_authentication}`)

src/components/rbac/permission-categories.tsx
```diff
@@ -0,0 +1,143 @@
...
+      <React.Fragment>
+        {groupsToMap.map((group) => (
+          <Flex
+            style={{ marginTop: '16px' }}
+            alignItems={{ default: 'alignItemsCenter' }}
+            key={group.label}
+            className={group.label}
+          >
+            <FlexItem style={{ minWidth: '200px' }}>
+              {i18n._(group.label)}
```
himdel, line 67: no `i18n._` calls should be needed anymore .. the data is coming translated from the API

src/components/rbac/permission-categories.tsx
```diff
@@ -0,0 +1,143 @@
...
+    this.setState({ groups: this.formatPermissions(model_permissions) });
+  }
+
+  render() {
+    const { groups } = this.state;
+    const { role, showEmpty, showCustom } = this.props;
+
+    const origGroups = groups.map((group) => ({
+      ...group,
+      label: i18n._(group.label),
```
himdel, line 40: nor here

src/components/rbac/permission-categories.tsx
```diff
@@ -0,0 +1,143 @@
...
+                  )
+                  .map((permission) => this.getNicenames(permission))
+                  .sort()}
+                selectedPermissions={role.permissions
+                  .filter((selected) =>
+                    group.object_permissions.find((perm) => selected === perm),
+                  )
+                  .map((permission) => this.getNicenames(permission))}
+                menuAppendTo='inline'
+                multilingual={true}
```
himdel, line 84: this causes PermissionChipSelector to call i18n as well, shouldn't be needed either

src/containers/role-management/role-list.tsx
ShaiahWren: @newswangerd can you weigh in on this?

src/components/rbac/permission-categories.tsx
```diff
@@ -0,0 +1,145 @@
...
+    const { role, showEmpty, showCustom, showUserMgmt } = this.props;
+    const { model_permissions } = this.context.user;
+
+    if (!showUserMgmt) {
+      Constants.USER_GROUP_MGMT_PERMISSIONS.forEach((perm) => {
+        if (perm in model_permissions) {
+          delete model_permissions[perm];
+        }
+      });
+    }
```
himdel, line 45: oh this is problematic because we're mutating state in context by doing it this way... after this deletes a permission, permission checks for that permission will start failing because it's no longer a known permission

.. so we need to copy the object first, before changing it..

(and delete doesn't mind if the permission doesn't exist)

```suggestion
    const model_permissions = { ...this.context.user.model_permissions };

    if (!showUserMgmt) {
      Constants.USER_GROUP_MGMT_PERMISSIONS.forEach(
        (perm) => delete model_permissions[perm],
      );
    }
```

himdel:
I remember you're on PTO in January and didn't want to leave this hanging, so I added a couple of commits to:

* resolve my feedback above
* tested `showEmpty`, `showCustom`, `showUserMgmt` working - added missing constants to `USER_GROUP_MGMT_PERMISSIONS`
* add a `ModelPermissionsType` for `context.user.model_permissions`
* removed (now unused) `RolePermissions`
* move permission names logic from `PermissionCategories` to `PermissionChipSelector`
* make `PermissionChipSelector` use names from `model_permissions` when available (`PermissionCategories` only knows about the translated names for categories, passes a list of permission ids to chip selector, which looks those up to show the translated version) (only done `viewOnly==true && multilingual==false` to keep edit working as before)
* switched the RoleList view to showCustom=true showEmpty=false now that we can

If the tests pass and all that, this should be ready, leaving only the edit-mode bits unfinished (#3025).

More things to check before both of these are done-done:

* `USER_GROUP_MGMT_PERMISSIONS` - merge with `PermissionCategories` logic; still used in
  * `src/containers/group-management/group-detail-role-management/group-detail-role-management.tsx`
* `HUMAN_PERMISSIONS` - should be using `model_permissions` now; still used in 
  * `src/containers/group-management/group-detail-role-management/group-detail-role-management.tsx`
  * `src/components/rbac/preview-roles.tsx`
  * `src/components/rbac/role-form.tsx`
* `PERMISSIONS` - should be using `model_permissions` now; still used in 
  * `src/components/rbac/role-form.tsx`
* `twoWayMapper` - should be gone, PermissionChipSelector will work with permission ids, and use `model_permissions` to display the translated names; still used in
  * `src/components/rbac/role-form.tsx`

himdel:
Actually, the edit part was easy after that O:-)

So, adding that here as well, this just adds a `setSelected` prop to `PermissionCategories`, and uses that to set edit mode and make it work.  And cleaned up `PermissionChipSelector` a bit since there's no longer multiple modes.

The checklist is done now, removing those constants and `twoWayMapper`.

Closes #3025
Closes #2452 

TODO(me): test #2228 user/group management hiding really still works in edit mode

PR #2980, opened 2022-12-05
Release 4.6.3

jerabekjiri:
No-Issue


newswangerd (approved):

himdel (approved):
LGTM :+1: 

The test failures are expected for now, all of those are tests that need to be able to change settings, which fails now, master fix in #2822, I don't have a good way of fixing this for 4.4-4.6 yet.

PR #2984, opened 2022-12-08
Use version range when comparing dependencies.

newswangerd:
Use version range when looking up if a dependency exists on the collection dependencies tab.

Issue: AAH-2018

Requires: https://github.com/ansible/galaxy_ng/pull/1571

### DIFF
--- a/src/containers/collection-detail/collection-dependencies.tsx
+++ b/src/containers/collection-detail/collection-dependencies.tsx
@@ -197,13 +197,12 @@ class CollectionDependencies extends React.Component<
 
     Object.keys(dependencies).forEach((dependency) => {
       const [namespace, collection] = dependency.split('.');
-      let version = dependencies[dependency];
-      version = this.separateVersion(version).version;
+      const version_range = dependencies[dependency];
 
       const dependency_repo = {
         name: collection,
         namespace: namespace,
-        version: version,
+        version_range: version_range,
         repo: '',
         path: '',
       };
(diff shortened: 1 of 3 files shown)


himdel:
There's a `version_range` filter?

We talked about this for https://github.com/ansible/ansible-hub-ui/pull/2713, and went with the simple fix because there was no way to filter by version, leaving a TODO in https://issues.redhat.com/browse/AAH-1968.

Was it always there? :)

src/containers/collection-detail/collection-dependencies.tsx
```diff
@@ -222,19 +221,16 @@ class CollectionDependencies extends React.Component<
...
       .then((result) => {
         dependency_repo.repo = result.data.data[0].repository_list[0];
-        dependency_repo.path = formatPath(
-          Paths.collectionByRepo,
-          {
-            collection: dependency_repo.name,
-            namespace: dependency_repo.namespace,
-            repo: dependency_repo.repo,
-          },
-          dependency_repo.version,
```
himdel, line 236: intentionally not linking to a matching version?

jerabekjiri (approved):

jerabekjiri:
API and UI backports don't match, API has labels `backport-4.5` and `backport-4.6`, but this PR only `backport-4.5`.

And I'm pretty sure `repo` and `imports` test fails are unrelated.

himdel (approved):
LGTM, assuming the missing version in the link is intentional (or fixed) :)

Tests are green on master now, so rebasing should fix them here as well :)

src/containers/collection-detail/collection-dependencies.tsx
newswangerd: yeah, the version link was broken before, and I figure it's best to not pick a random collection that matches the version range.

PR #2985, opened 2022-12-09
Tests for insights mode

MilanPospisil:
Issue: AAH-2054

Create or adapt existing tests for insights mode.


test/cypress/e2e/insights/login.js
```diff
@@ -0,0 +1 @@
+/home/milan/software/ansible-hub-ui/test/cypress/e2e/groups_and_users/login.js
```
himdel, line 1: relative vs absolute symlinks ;)

MilanPospisil: Ah, make sense, dumb mistake :-)

.github/workflows/cypress/cypress.env.json.insights
```diff
@@ -5,5 +5,7 @@
   "username": "admin",
   "password": "admin",
   "galaxykit": "galaxykit --ignore-certs --auth-url 'http://localhost:8002/auth/realms/redhat-external/protocol/openid-connect/token'",
-  "insightsLogin": true
+  "insightsLogin": true,
+  "disableRepoSwitch" : true,
+  "namespaceName" : "Partners"
```
himdel, line 10: what is this trying to do?

I only see it used as an url fragment in goToNamespaces .. but we definitely don't have any routes with capital letters

If you're already relying on insights mode, why not just use the whole route?
Or, if you want this clean, drop the logic based on mode and have a setting for the namespaces url to go to

test/cypress/e2e/namespaces/namespace_detail.js
```diff
@@ -77,28 +77,30 @@ describe('Namespace detail screen', () => {
     // The test for success are impmeneted in the collection_upload file
   });
 
-  it('should filter by repositories', () => {
-    const namespace = 'coolestnamespace';
-    cy.galaxykit('-i namespace create', namespace);
-    cy.visit(`${uiPrefix}repo/published/${namespace}`);
+  if (!disableRepoSwitch) {
```
himdel, line 80: would it make more sense to move repo switcher tests to a separate file instead?

(They will go away once repository management is in, so .. either works.)

.github/workflows/cypress/cypress.env.json.insights
```diff
@@ -5,5 +5,6 @@
   "username": "admin",
   "password": "admin",
   "galaxykit": "galaxykit --ignore-certs --auth-url 'http://localhost:8002/auth/realms/redhat-external/protocol/openid-connect/token'",
-  "insightsLogin": true
+  "insightsLogin": true,
+  "disableRepoSwitch" : true,
```
himdel, line 9: trailing commas are not valid json

test/cypress/e2e/namespaces/namespace_detail.js
MilanPospisil: So if they will go away soon, why just not leave them here for now?

himdel (approved):

PR #2850, opened 2022-11-14
Merge Galaxy Doc Builder code into ansible-hub-ui

himdel:
Taking components and docs from https://github.com/ansible-community/galaxy-doc-builder and moving into ansible-hub-ui src/components/render-plugin-doc

This unblocks #2095, by removing a react 16 version constraint.

PR #3012, opened 2022-12-12
RenderPluginDoc: convert bools to true/false, not yes/no

himdel:
Issue: AAH-1859

Before:

After:

Change the default way of displaying boolean options from yes/no to true/false, as per https://github.com/ansible-community/community-topics/issues/116


himdel:
(backporting this will need https://github.com/ansible/ansible-hub-ui/pull/2850 as well)

newswangerd (approved):

PR #3016, opened 2022-12-13
workflows - use rh-galaxy-droid key to allow pushing to protected branches

himdel:
Enabling the branch push protection stops i18n, and cloud-stage-* workflows from being able to push to stable-*/master

there's no way to use GITHUB_TOKEN together with push protection, but we can use a different user
=> going with @rh-galaxy-droid, used for manifests

wip.. need to check that's the right key, and the right account

fixes https://github.com/ansible/ansible-hub-ui/actions/runs/3667585193/jobs/6200136062#step:5:16


himdel:
Closing as not needed now, I'm still not sure about the keys, might want to create a new set for the specific stuff if needed.

PR #3017, opened 2022-12-13
Fix reserved resources on task management

newswangerd:
Issue: AAH-2055

This seems to have broken when we updated the pulp api base path.


himdel (approved):
:+1: 

Given a random `pulp_container.app.tasks.synchronize.synchronize` task...

Before



After:



LGTM :)

PR #3020, opened 2022-12-13
Configurable legacy prefix

jctanner:
The community has raised various complaints over using the term "legacy" to refer to roles. This PR makes that term configurable in the display so that the community can debate amongst themselves and change it to whatever they want when they reach a consensus.

Due to the lack of support for computed values in enums, the configurable term can't influence the paths in paths.tx, at least not until https://github.com/microsoft/TypeScript/pull/50528 is extended for strings.

image

NOTE: The "standalone" prefix doesn't really make sense in the context of the left nav menu. It would be better if the standalone roles and standalone namespaces fell under the "Collections" NAV and that was renamed.

### DIFF
--- a/src/constants.tsx
+++ b/src/constants.tsx
@@ -1,6 +1,5 @@
 import { t, defineMessage } from '@lingui/macro';
 
-
 export class Constants {
   static readonly SEARCH_VIEW_TYPE_LOCAL_KEY = 'search_view_type';
   static readonly DEFAULT_PAGE_SIZE = 10;
--- a/src/constants.tsx
+++ b/src/constants.tsx
@@ -1,5 +1,6 @@
 import { t, defineMessage } from '@lingui/macro';
 
+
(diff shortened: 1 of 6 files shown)


gotmax23:
> NOTE: The "standalone" prefix doesn't really make sense in the context of the left nav menu.

How about

- Stanadlone Roles
  - Roles
  - Namespaces

himdel:
From the l10n pov, this is not so great (translating `{foo} Roles` separately from `Standalone/Legacy` is a problem in any language with a concept of gendered nouns), and we will need stable URLs at *some* point (note that whichever prefix we end up chosing will become an invalid namespace name for collection namespaces, beacuse of the `/:namespace/:collection` route).

But if we need to evolve this incrementaly with feedback, LGTM :+1:

himdel:
Closing, development is moving to https://github.com/ansible/ansible-ui/

PR #3021, opened 2022-12-14
Use page_size for v1 instead of limit.

jctanner:
Pagination isn't working because v1 pagination parameters differ from v3, so we need to use page & page_size.

https://github.com/ansible/galaxy_ng/pull/1575

### DIFF
--- a/src/api/legacy.ts
+++ b/src/api/legacy.ts
@@ -12,16 +12,8 @@ export class LegacyAPI extends BaseAPI {
   }
 
   public mapPageToOffset(p) {
-    // Need to copy the object to make sure we aren't accidentally
-    // setting page state
-    const params = { ...p };
-
-    const pageSize =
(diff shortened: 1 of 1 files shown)


src/api/legacy.ts
```diff
@@ -10,6 +11,19 @@ export class LegacyAPI extends BaseAPI {
...
   }
 
+  public mapPageToOffset(p) {
+    // Need to copy the object to make sure we aren't accidentally
+    // setting page state
+    const params = { ...p };
+
+    const pageSize =
+      parseInt(params['page_size']) || Constants.DEFAULT_PAGE_SIZE;
+    const page = parseInt(params['page']) || 1;
```
himdel, line 21: page is unused now

also.. if this is just supposed to make sure we're not rewriting page&page_size to offset&limit, the whole function could just be `return p;`, right?

PR #3025, opened 2022-12-14
Editable role form component

ShaiahWren:
Issue: AAH-2016

This ajdusts the `RoleForm` component to handle object permission names coming from the API `/me` endpoint instead of the `Constants.js` file. I have reworked the `twoWayMapper` util function so it no longer works with the constants. 

Any suggestions welcome:) There may be a merge conflict after [pr#2955](https://github.com/ansible/ansible-hub-ui/pull/2955) goes through, but I figure that can be ironed out. :)


src/components/rbac/permission-chip-selector.tsx
```diff
@@ -11,10 +11,10 @@ import {
 } from '@patternfly/react-core';
 
 interface IProps {
-  availablePermissions?: string[];
-  selectedPermissions: string[];
+  availablePermissions?: object[];
```
himdel, line 14: `object` is rather generic, can we do

```
type Permission = {
  global_description: string;
  has_model_permission: boolean;
  name: string;
  object_description: string;
  ui_category: string;
];
```

and use `Permission[]` here?

himdel:
Looks like there's a type conflict with some of the components .. that's probably because they're still being used the old way in all the read-only places addressed in #2955 .. so it would go away once that one is done?

Second, now that the component knows about both the `galaxy.foobar` id names and the `View users` nice names, I'd like to get rid of those `twoWayMapper` calls if possible.

In places using twoWayMapper to map from ids to nice names, we can just do that lookup directly (and should also have the nice name available already),
and any places mapping from nice names to ids should be obsolete now, because we also have the id available.

himdel:
I might have misled you here with how to fix `twoWayMapper`, sorry about that...
Turns out the translation bits can be done entirely inside `PermissionChipSelector`, without any of the parent components knowing about these .. so `selectedPermissions` and `availablePermissions` can still be `string[]` - arrays of permission ids - onSelect can be called with just the keys, and we only need to ensure that the nice name lookup happens when displaying the permission here.
(Resolved for view-only in #2955, more details there.)

Also, I created a checklist of things to check for before this is done done - see https://github.com/ansible/ansible-hub-ui/pull/2955#issuecomment-1368087784.

src/api/response-types/permission.ts
```diff
@@ -0,0 +1,4 @@
+export class PermissionType {
```
himdel, line 1: (won't be needed, I replaced it with the ModelPermissionType in #2955)

himdel:
Implemented directly in #2955

PR #3056, opened 2022-12-19
Divide tests into more folders and make them faster

MilanPospisil:
No-Issue

Divide approval and collection tests into 2 separate folders. Those two tests are slowest and slow down the whole PR test runs.


himdel:
This won't help much anymore,
`approval` took 14m, now it takes 10m + 11m in parallel,
`collections` took 16m, now it's 10m + 12m.

So we're saving 4m in clock time by investing 13m of run time.

We can get better savings by cleaning up the duplicate tests and using fewer items in setup.

MilanPospisil:
> This won't help much anymore, `approval` took 14m, now it takes 10m + 11m in parallel, `collections` took 16m, now it's 10m + 12m.
> 
> So we're saving 4m in clock time by investing 13m of run time.
> 
> We can get better savings by cleaning up the duplicate tests and using fewer items in setup.

Yes, I see, collections tests are the main delay. You are right, lets lower the number of items. Those test should be reliable enough with considerable time saving.

PR #3063, opened 2022-12-21
[beta] UX: give a helpful message about roles with no versions

MilanPospisil:
Issue: AAH-2063

https://issues.redhat.com/browse/AAH-2063

### DIFF
--- a/src/containers/legacy-roles/legacy-role.tsx
+++ b/src/containers/legacy-roles/legacy-role.tsx
@@ -132,6 +132,7 @@ class LegacyRoleVersion extends React.Component<
 
 interface RoleVersionsIProps {
   role_versions: LegacyRoleVersionDetailType[];
+  isLoading: boolean;
 }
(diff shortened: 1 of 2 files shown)


src/containers/legacy-roles/legacy-role.tsx
```diff
@@ -147,13 +149,20 @@ class LegacyRoleVersions extends React.Component<RoleMeta, RoleVersionsIProps> {
...
     });
   }
 
   render() {
     return (
       <div id='versions-div'>
+        {!this.state.loading &&
+        this.state.role_versions &&
+        this.state.role_versions.length == 0
+          ? t`The role is versionless and will always install from the head/main/master branch.`
```
himdel, line 163: thous should be an `EmptyStateNoData`, not just a raw string. (But the message itself looks good.)

himdel (approved):

PR #2982, opened 2022-12-08
Finish automerge of dependabots PR

MilanPospisil:
Issue: AAH-2053

Finish automerge of dependabots PR.

PR #3065, opened 2022-12-21
Automerge with --auto

MilanPospisil:
Related to #3026, #2982, #2899, #2660

This attempts to solve the problem of "gh merge won't happen while the automerge workflow is still running".


himdel (approved):
:crossed_fingers:

PR #3118, opened 2023-01-02
Convert SignatureKeysList & TaskList to ListPage

himdel:
The idea here is to convert list pages to an intermediate representation between current state and ansible-ui-framework by moving the page layout and common logic out to a shared ListPage component factory, separate the page-specific logic from the common one, while not changing how it works nor how it renders for testability.
(Any visual changes, or conversion filter and field definitions to framework style will be a separate step as well, in or after #3147.)

Converted `SignatureKeysList` and `TaskList` to use it.

---

`ListPage` params - returns a router-aware container

* `condition` - rbac, a permission checker function - from `src/permissions` (`{ featureFlags, settings, user } => bool`)
* `didMount` (optional) - extra code to run on mount
* `displayName` - component name for debugging, should be the same as the export name
* `defaultPageSize` - initial page size
* `defaultSort` (optional) - initial sort ordering
* `errorTitle` - alert on query failure
* `extraState` (optional) - extra initial state
* `filterConfig` - filters
* `headerActions` - buttons to show after filter
* `noDataDescription`, `noDataTitle` - `EmptyStateNoData` props
* `query` - `({ params }) => Promise<{ data: { count, results[] } }>`
* `renderModals` (optional) - `({ addAlert, state, setState, query }) => <ConfirmationModal... />`
* `renderTableRow` - `(item, index) => <tr>...</tr>`
* `sortHeaders` - table headers
* `title` - container title

`Action` params:
* `button` and `dropdownItem` render the right thing for `ListActionItems`, respecting a `visible(item)` check for both rbac and item-specific logic, and doing `onClick`
* `modal` renders a modal

`src/permissions`:
* trying to change all permision checks into a `canViewWhatever` call - action and model specific checks defined in one place

### DIFF
--- a/src/actions/action.tsx
+++ b/src/actions/action.tsx
@@ -1,10 +1,15 @@
 import * as React from 'react';
-import { DropdownItem } from '@patternfly/react-core';
+import { Button, DropdownItem } from '@patternfly/react-core';
 
 export const Action = ({ title, onClick }) => ({
   dropdownItem: (item) => (
     <DropdownItem key={title} onClick={() => onClick(item)}>
       {title}
     </DropdownItem>
+  ),
+  button: (item) => (
+    <Button key={title} onClick={() => onClick(item)}>
+      {title}
+    </Button>
   ),
 });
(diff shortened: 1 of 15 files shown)


himdel:
Most of this was merged in #3144 with less history .. closing to keep that... and opening a new PR with the remaining changes - conversion of signature keys & tasks

PR #3144, opened 2023-01-03
Repositories & Remotes

himdel:
wip, also includes changes from #3118,
missing forms, modals, details&tabs


himdel:
(screenshot test failures expected - new menu items, renames)

MilanPospisil:
I see some minor things, that maybe should be upgraded after this PR is merged (it is not essential) and maybe you have different opinion:

After deletion, task appears, but after task continues, UI could react accordingly - display success message, reload the table, or redirect back to the table from detail. We also dont have this standardised, for example collection and namespaces dont alert about task, but they do the redirection/reload.

Update alerts the task, but not success message. User can be confused that it is still updating, while page was clearly reloaded. But it does not reloads after distribution is created (and no success message). The same for adding the remote.

Proxy URL is not validated in edit remotes like URL is, but this is problem probably carried from old remote edit modal and probably not imporant.

MilanPospisil:
In the repositories detail, there is list of labels. when you select pipeline, and also hide from search and private, it will show expandable "1 more" button, but after button click, the whole detail is reloaded and it will not show expanded list, it will show "1 more" button again (it will show the whole list for nearly a second though, but it will reload to "1 more" button after).

src/components/page/list-page.tsx
```diff
@@ -0,0 +1,360 @@
...
+      <>
+        {headerActions?.length &&
+          headerActions.map((action) => action?.modal?.(actionContext))}
+        {listItemActions?.length &&
+          listItemActions.map((action) => action?.modal?.(actionContext))}
+      </>
+    );
+  };
+
+  const klass = class extends React.Component<RouteProps, IState<T>> {
```
MilanPospisil, line 140: Why the weird name klass?

src/containers/ansible-remote/detail.tsx
```diff
@@ -0,0 +1,142 @@
...
+import {
+  CopyURL,
+  Details,
+  LazyRepositories,
+  PageWithTabs,
+} from 'src/components';
+import { Paths, formatPath } from 'src/paths';
+import { isLoggedIn } from 'src/permissions';
+
+const wip = '🚧 ';
```
MilanPospisil, line 26: What is that? I have never seen that before.

himdel: [construction sign emoji](https://emojipedia.org/construction/) :)

src/components/page/list-page.tsx
himdel: Yeah, mentioned in the description, this is a component factory until I can convert it to a functional layout component and separate out the state management bits. The why is just time :)

himdel:
Merging to unblock the rest,
and adding the PulpLabels fix to #3430 :+1: 

(As for task alerts, valid point, we should extend taskAlert to start a polling task, but this matches the other task handling code in other places for now.)

PR #3153, opened 2023-01-05
Container signing feature flag fix

himdel:
Fixes AAH-2013, AAH-2015
(AAH-1989 will come as a separate PR)

Before:
used the `can_create_signatures`, `can_upload_signatures`, `display_signatures`, `require_upload_signatures`, `signatures_enabled` feature flags for both container(EE) and collection signatures, and only treated the `container_signing` / `collection_signing`, `collection_auto_sign` flags as model specific

After:
* collection/namespace/approvals signing uses `can_create_signatures`, `can_upload_signatures`, `collection_auto_sign`, `collection_signing`, `display_signatures`, `require_upload_signatures` (as documented in https://galaxyng.netlify.app/config/collection_signing/#feature-flags)
* EE signing uses `container_signing` only
* `SignatureKeys` use `collection_signing || container_signing`
* (nothing uses `signatures_enabled` since it doesn't mean what it says, as it's an alias for `collection_signing` only)

AAH-2013: `canSignEE` no longer checks `can_create_signatures`, or `signatures_enabled`, only `container_signatures`
AAH-2015: `SignatureBadge` now gets a `displaySignatures` check using either `display_signatures` or `container_signing`, depending on where we are.

(And cleaned up some obsolete defensive programming around feature flags availability.)

Cc @hendersonreed @rochacbruno

### DIFF
--- a/CHANGES/2013.bug
+++ b/CHANGES/2013.bug
@@ -1,0 +1,1 @@
+Show container signing button based only on container_signing, not collection_signing
--- a/CHANGES/2015.bug
+++ b/CHANGES/2015.bug
@@ -1,0 +1,1 @@
+Show container signature badge based only on container_signing, not collection_signing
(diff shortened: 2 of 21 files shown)


MilanPospisil (approved):
LGTM 👍

PR #3154, opened 2023-01-05
Make approval collection list test faster

MilanPospisil:
No-Issue

Instead of 23 items, this PR now creates only 11 items. This is enough for the test, only paging is affected and the test is still covering most functionality.


himdel (approved):
Takes the approvals test "Run cypress" step from 11m to 7m :tada:

PR #3168, opened 2023-01-09
Add test to click on left navigation menu items

MilanPospisil:
https://issues.redhat.com/browse/AAH-2011

### DIFF
--- a/test/cypress/e2e/insights/menu.js
+++ b/test/cypress/e2e/insights/menu.js
@@ -1,15 +1,19 @@
 const uiPrefix = Cypress.env('uiPrefix');
 
 describe('Insights Menu Tests', () => {
-
-  function menuPresent(item)
-  {
-    cy.contains(`[data-quickstart-id="Automation-Hub"] [data-ouia-component-id="${item}"] a`, item);
+  function menuPresent(item) {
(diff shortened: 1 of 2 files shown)


test/cypress/e2e/insights/menu.js
```diff
@@ -0,0 +1,82 @@
...
+  });
+
+  it('can navigate to Task Management', () => {
+    menuClick('Task Management');
+    cy.contains('main', 'Task Management');
+    cy.contains('main .body', 'Task name');
+  });
+
+  it('can navigate to Connect to Hub', () => {
+    cy.on('uncaught:exception', (err, runnable) => {
```
himdel, line 70: I wonder if we can `cy.intercept` the problematic call instead

MilanPospisil: Does not seems like that. Intercept only waits for successful api call. I tried to filter messages for the one api call (metrics) that fails. Other api failures will fail.

MilanPospisil: It seems that cy on really ignores only api calls inside it. If anything happens in next it, it is not ignored anymore.

himdel (approved):

PR #3172, opened 2023-01-13
[stable-4.2] Dead code cleanup, sync prettier config

himdel:
(4.2: here, 4.3: eol, 4.4: #3173, 4.5: #3174, 4.6: #3175, master: #3176)

Follows #2413 & #1918

We're still depending on redux transitively and have a bunch of configs that are always set to the same value because most but not all insights-related code was removed from stable branches.

Removing unused dependencies, code and configs.

Clean up unused exports - #3178 
Convert imports to src/ - #319
Update prettier, plugin sort imports - #3198


himdel:
```
[webpack-cli] Failed to load '/home/himdel/ansible-hub-ui-4.2/config/standalone.prod.webpack.config.js' config
[webpack-cli] TypeError: Cannot read properties of undefined (reading 'appname')
```

himdel:
TODO add #3178 and #3198

PR #1920, opened 2022-04-19
[stable-4.4] Cleanup github files, README

himdel:
This drops unused `.firebaserc` (#1917),

drops insights mode, Deploying and Workflow sections from README and remove `profiles/` (insights mode updated in #1526, deploying & workflow is up-to-date only on master),

remove workflows not relevant outside master: `deploy-cloud.yml`, `dev-release.yml`, `i18n.yml`, `update-manifest.yml`
remove github files that only work on master: `PULL_REQUEST_TEMPLATE.md`, `ISSUE_TEMPLATE/`, `dependabot.yml`, `patchback.yml`

removes husky (only needed on master)

PR #3173, opened 2023-01-13
[stable-4.4] Dead code cleanup, sync prettier config

himdel:
(4.2: #3172, 4.3: eol, 4.4: here, 4.5: https://github.com/ansible/ansible-hub-ui/pull/3174, 4.6: https://github.com/ansible/ansible-hub-ui/pull/3175, master: https://github.com/ansible/ansible-hub-ui/pull/3176)

Follows https://github.com/ansible/ansible-hub-ui/pull/3141 & https://github.com/ansible/ansible-hub-ui/pull/1920

We're still depending on redux transitively and have a bunch of configs that are always set to the same value because most but not all insights-related code was removed from stable branches.

Removing unused dependencies, code and configs.

Clean up unused exports - #3178 
Update prettier, plugin sort imports - #3198


himdel:
TODO add #3178 and #3198

PR #3176, opened 2023-01-13
Clean up more old code

himdel:
(4.2: #3172, 4.3: eol, 4.4: https://github.com/ansible/ansible-hub-ui/pull/3173, 4.5: https://github.com/ansible/ansible-hub-ui/pull/3174, 4.6: https://github.com/ansible/ansible-hub-ui/pull/3175, master: here)

Follows #1263 

We're still depending on redux transitively and have some old related code and configs, removing.
Also fixed some unused exports and relative path imports.


himdel:
This is not touching any actual code, but many files .. merging right after #1263 so everything doesn't need to get rebased twice.

PR #3198, opened 2023-01-17
Prettier: sort imports, bump to 2.8

himdel:
Bump prettier from 2.3 to 2.8
Clean up deprecated prettier config, config that is already default and config already set in `.editorconfig`.

Add prettier plugin to sort js/ts imports:
* (force `../app.scss` to go before `patternfly.css` - reordering breaks token page paddings)
* sort external imports first,
* then src/,
* then any relative paths

(within those groups, sort imports by source name; sort specifiers within import asciibetically)

### DIFF
--- a/.prettierrc.yaml
+++ b/.prettierrc.yaml
@@ -5,3 +5,7 @@ tabWidth: 2
 jsxSingleQuote: true
 jsxBracketSameLine: false
 bracketSpacing: true
+importOrder:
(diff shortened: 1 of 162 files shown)


himdel:
(merging after #1263 and #3176 to cluster any disruptions)

himdel:
(merging after #1263 and #3176 to cluster any disruptions)

PR #3227, opened 2023-01-27
Switch frontend l10n implementation to i18next

MilanPospisil:
https://issues.redhat.com/browse/AAH-2081¨

Extraction: i18next "src/**/*.{js,tsx}"


himdel:
If it helps, here's my checklist of things we need to test with i18next..

* simple strings
  * works with "key" but will it work with whole sentences? (i18next might be separating things by `.` and `:` by default)
  * does it try to do too much/too little escaping? (is `<div>{t('4 < 5')}</div>` showing a `<`, or nothing or a doubly-escaped `&gt;`?)
  * what happens for unknown strings (fallback to the original string) - technically, I think this means falling back to the 'dev' language, so we should test if it still works in production (=CI)
* interpolation
  * check escaping in interpolated values
  * unless there's a way to keep using template strings, we might need a linter rule to check for issues like `t('Foo {{bar}}',  { baz: "not bar" })`
* plurals
  * what's the syntax, how much do we need to change ours?
  * how does the plural format look when extracted in the json? (for conversion, even the po converter might not understand lingui flavor plural syntax)
* deferred translations (lingui defineMessage)
  * what's the syntax, does the extractor handle it correctly
  * (or do we need to fake it by having a `t` with forced translation to the original language? (or a PR to convert them to functions with normal translations))
* react
  * does it make sense to have our own Trans component that calls the hook so we don't have to always pass in `t`?
  * is escaping consistent between `t` and `Trans`?
  * does it re-render on language change? (we don't change languages at runtime, but rendering components might happen before the dynamic import finishes - make sure it either waits before rendering, or switches when ready (import just returns a Promise, so you can just add a `.then` that returns a new promise that resolves with a large  timeout to test this))
* template strings
  * if we can make ``t`template ${'strings'}` `` work, switching would be easier (probably needs the extractor parser to know about them, rest is easy)
* extraction - there seem to be multiple parsers, use the same one as ansible-ui, if possible
* pseudolocalization - preserve french quotes, the browser language detector plugin may be able to read the same cookies
(* conversion - there may already be a plugin to convert po to json)
* (later: think about using namespaces to keep framework strings in a separate json)

locale/en.json
```diff
@@ -0,0 +1,624 @@
...
+  "Yes, stop": "Yes, stop",
+  "Task &quot;{taskName}&quot; stopped successfully.": "Task &quot;{taskName}&quot; stopped successfully.",
+  "You do not have permission to view all tasks. Only tasks created by you are visible.": "You do not have permission to view all tasks. Only tasks created by you are visible.",
+  "Task Management": "Task Management",
+  "No tasks yet": "No tasks yet",
+  "Tasks will appear once created.": "Tasks will appear once created.",
+  "Started at": "Started at",
+  "Task list": "Task list",
+  "Stop task?": "Stop task?",
+  "Task &quot;{name}&quot; stopped successfully.": "Task &quot;{name}&quot; stopped successfully.",
```
himdel, line 564: escaping is probably misconfigured, these should be 
```
  "Task \"{name}\" stopped successfully."
```

i18next-parser.config.mjs
```diff
@@ -0,0 +1,106 @@
+export default {
+  contextSeparator: '_',
+  // Key separator used in your translation keys
+
+  createOldCatalogs: true,
```
himdel, line 5:
```suggestion
  createOldCatalogs: false,
```

i18next-parser.config.mjs
```diff
@@ -0,0 +1,106 @@
...
+  // If you want to use plain english keys, separators such as `.` and `:` will conflict. You might want to set `keySeparator: false` and `namespaceSeparator: false`. That way, `t('Status: Loading...')` will not think that there are a namespace and three separator dots for instance.
+
+  // see below for more details
+  lexers: {
+    hbs: ['HandlebarsLexer'],
+    handlebars: ['HandlebarsLexer'],
+
+    htm: ['HTMLLexer'],
+    html: ['HTMLLexer'],
+
```
himdel, line 32:
```suggestion
    hbs: ['HandlebarsLexer'],
    handlebars: ['HandlebarsLexer'],

    htm: ['HTMLLexer'],
    html: ['HTMLLexer'],

```

i18next-parser.config.mjs
```diff
@@ -0,0 +1,106 @@
...
+    jsx: ['JsxLexer'],
+    tsx: ['JsxLexer'],
+
+    default: ['JavascriptLexer'],
+  },
+
+  lineEnding: 'auto',
+  // Control the line ending. See options at https://github.com/ryanve/eol
+
+  locales: ['en'],
```
himdel, line 45:
```suggestion
  locales: ["en", "es", "fr", "ko", "nl", "ja", "zh"],
```

i18next-parser.config.mjs
```diff
@@ -0,0 +1,106 @@
...
+  output: 'locale/$LOCALE.json',
+  // Supports $LOCALE and $NAMESPACE injection
+  // Supports JSON (.json) and YAML (.yml) file formats
+  // Where to write the locale files relative to process.cwd()
+
+  pluralSeparator: '_',
+  // Plural separator used in your translation keys
+  // If you want to use plain english keys, separators such as `_` might conflict. You might want to set `pluralSeparator` to a different string that does not occur in your keys.
+
+  input: undefined,
```
himdel, line 61:
```suggestion
  input: 'src/**/*.{js,jsx,ts,tsx}',
```

i18next-parser.config.mjs
```diff
@@ -0,0 +1,106 @@
...
+  // An array of globs that describe where to look for source files
+  // relative to the location of the configuration file
+
+  sort: false,
```
himdel, line 65:
```suggestion
  sort: true,
```

src/i18n.ts
```diff
@@ -0,0 +1,35 @@
...
+    messages[key] = '»' + messages[key] + '«';
+  });
+}
+
+i18next.use(initReactI18next);
+i18next.use(LanguageDetector);
+i18next.init({
+  interpolation: { escapeValue: false },
+  debug: true,
+  fallbackLng: 'en',
```
himdel, line 27:
```suggestion
```

src/i18n.ts
```diff
@@ -0,0 +1,35 @@
...
+  resources: {
+    en: {
+      translation: en,
+    },
+  },
```
himdel, line 32:
```suggestion
  },
    returnEmptyString: false,
    returnNull: false,
    supportedLngs: availableLanguages,
```

(availableLanguages from src/l10n)

src/i18n.ts
```diff
@@ -0,0 +1,35 @@
...
+const messages = en;
+
+window.localStorage.test_l10n = true;
+
+if (window.localStorage.test_l10n === 'true') {
+  Object.keys(messages).forEach((key) => {
+    // simple string
+    messages[key] = '»' + messages[key] + '«';
+  });
+}
```
himdel, line 20: this is breaking empty translation handling (beause it's never empty)

```suggestion
if (window.localStorage.test_l10n === 'true') {
  Object.keys(messages).forEach((key) => {
    // simple string
    if (key) {
      messages[key] = '»' + messages[key] + '«';
    }
  });
}
```

himdel:
```diff
diff --git a/package.json b/package.json
index 3cd792c0..2c10692a 100644
--- a/package.json
+++ b/package.json
@@ -76,7 +76,7 @@
         "eslint": "eslint --ext .js,.jsx,.ts,.tsx src/ config/ test/",
         "eslint:fix": "eslint --ext .js,.jsx,.ts,.tsx --fix src/ config/ test/",
         "gettext:compile": "lingui compile",
-        "gettext:extract": "lingui extract",
+        "gettext:extract": "lingui extract; i18next",
         "lint": "npm-run-all lint:*",
         "lint-fix": "npm-run-all lint:*:fix",
         "lint-setup": "npm-run-all lint:*:setup",
```

himdel:
Closing, we'll stick with lingui after all :)

PR #3262, opened 2023-02-02
[wisdom] Community EULA - authenticated user

himdel:
Issue: AAH-2144, AAP-8138 (with #3311)

A community mode user will see this modal the first time they visit the UI.

`EulaModalUser` can accept or decline, confirms to imaginary API, reads confirmation from imaginary user field
TODO: what happens after decline?

---

UX update:

spinner:

error:

### DIFF
--- a/CHANGES/2144.feature
+++ b/CHANGES/2144.feature
@@ -1,0 +1,1 @@
+Community mode EULA modals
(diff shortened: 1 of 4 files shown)


himdel:
Replaced by a Terms of Use menu link

PR #3311, opened 2023-02-12
Community EULA - unauthenticated user

himdel:
Issue: AAH-2144

When running in viewonly mode (feature flag / community version), a user with is_anonymous will see this modal the first time they visit the UI. And then on any subsequent visits from different computers / whenever their browser forgets localStorage.

`EulaModalViewonly` only has the confirm action, saves confirmation in `localStorage.viewonly_eula_confirmed`.


himdel:
Replaced by a Terms of Use menu link

PR #3329, opened 2023-02-14
Bump cypress 11 to 12

MilanPospisil:
Fixes - https://github.com/ansible/ansible-hub-ui/pull/3271


himdel (approved):
LGTM :+1: 

The clearField logic probably undoes #3014 .. so this may start failing again if it tries to clear a previously cleared field .. but we can fix that once it happens, and the form probably gets rewritten for repo management anyway... merging :)

PR #3346, opened 2023-02-21
Remove inbound repo logic

jerabekjiri:
Issue: AAH-1778

- Updated collection upload path to `v3/artifacts/collections/`
- removed inbound repo bits

### DIFF
--- a/src/containers/namespace-detail/namespace-detail.tsx
+++ b/src/containers/namespace-detail/namespace-detail.tsx
@@ -260,7 +260,7 @@ export class NamespaceDetail extends React.Component<IProps, IState> {
         : null,
     ].filter(Boolean);
 
-    const repositoryUrl = getRepoUrl(namespace.name);
+    const repositoryUrl = getRepoUrl();
 
     const noData = itemCount === 0 && !filterIsSet(params, ['keywords']);
 
--- a/src/containers/namespace-detail/namespace-detail.tsx
+++ b/src/containers/namespace-detail/namespace-detail.tsx
@@ -260,7 +260,7 @@ export class NamespaceDetail extends React.Component<IProps, IState> {
         : null,
     ].filter(Boolean);
 
-    const repositoryUrl = getRepoUrl('inbound-' + namespace.name);
+    const repositoryUrl = getRepoUrl(namespace.name);
 
     const noData = itemCount === 0 && !filterIsSet(params, ['keywords']);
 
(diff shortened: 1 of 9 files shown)


himdel:
API: https://github.com/ansible/galaxy_ng/pull/1602 (merged now)

Oh, some of the remaining test failures may be because `galaxykit collection upload` will still be doing it the old way..

https://github.com/ansible/galaxykit/blob/main/galaxykit/collections.py#L151
but it probably also needs a version check to work with both pre-4.7 versions and this one.

jerabekjiri:
@himdel thanks, totally forgot about galaxykit, opened a fix https://github.com/ansible/galaxykit/pull/85 and tested with UI

himdel (approved):
LGTM, thanks! :) :green_circle:

src/containers/namespace-detail/namespace-detail.tsx
```diff
@@ -260,7 +260,7 @@ export class NamespaceDetail extends React.Component<IProps, IState> {
         : null,
     ].filter(Boolean);
 
-    const repositoryUrl = getRepoUrl('inbound-' + namespace.name);
+    const repositoryUrl = getRepoUrl(namespace.name);
```
newswangerd, line 263: this should be the api endpoint for the service (`/api/galaxy/v3` or `/api/automation-hub/v3`, not `/api/galaxy/content/<namespace>/v3`)

newswangerd (changes requested):

src/containers/namespace-detail/namespace-detail.tsx
jerabekjiri: so currently the URL is `http://localhost:8002/api/automation-hub/v3/content/my_namespace/` and looking at URLs

```
pulpcore-manager show_urls | grep /api/automation-hub/content/
/api/automation-hub/content/<str:path>/ galaxy_ng.app.api.views.ApiRootView     galaxy:api:v3:content-root
```
I think this is correct endpoint, there is no endpoint like `/api/automation-hub/v3`?

jerabekjiri: Also, this is wrong, this should not point to the namespace, but selectedRepo (`inbound-<namespace>` is repo, not namespace, I confused it)

himdel:
LGTM :+1: 

(ignore the insights test failures, it seems there's a new insights API we're not mocking, causign all tests to fail there)

newswangerd (approved):

PR #3362, opened 2023-02-28
[stable-4.6] CI: run collectstatic manually, fix 404

himdel:
Test failure: https://github.com/ansible/ansible-hub-ui/actions/runs/4293339560/jobs/7480897727

`/var/lib/operator/static/` ends up empty in the container, we need to run `pulpcore-manager collectstatic` (but it asks and fails when it's non-empty, hence `--clear --no-input`)


himdel:
Merging as is, this fixes all tests except for container signing.

> `container-sign.sh: line 12: /tmp/key-password.txt: Permission denied`

Adressing that one in a separate PR

PR #3392, opened 2023-03-02
[stable-4.5] AuthHandler - wait for user, settings, featureFlags before deciding where to find the login screen

himdel:
Issue: AAH-2141
(and AAP-9768)

Tentative fix for ^.
This covers the screnario where SSO is configured correctly, but user API responds much sooner than feature flags, causing AuthHandler to redirect to /ui/login, ignoring the external auth feature flag, because isLoading is only checking user.


ironfroggy:
lgtm, minus the cypress failures.

himdel:
Ah, unlike `Promise.all(...).catch(...)`, `Promise.allSettled(...)` won't mark exceptions as handled, so any visit to the login screen when not logged in blows up because the 403s are not marked as handled.

=> adding a `.catch((e) => null);` to the end of the `Promise.all` chain handles that

PR #3428, opened 2023-03-15
Collection approval workflow UI

MilanPospisil:
https://issues.redhat.com/browse/AAH-2169


MilanPospisil:
Before:
image

After:
image

image

src/components/approve-modal/approve-modal.tsx
```diff
@@ -0,0 +1,370 @@
...
+          <div className='footer'>
+            <Pagination
+              params={params}
+              updateParams={(p) => setParams(p)}
+              count={props.allRepositories.length}
+            />
+          </div>
+        </section>
+
+        <AlertList alerts={alerts} closeAlert={() => closeAlert()} />
```
himdel, line 366: :eyes:

src/components/approve-modal/approve-modal.tsx
```diff
@@ -0,0 +1,370 @@
...
+                    ]}
+                  />
+                </ToolbarItem>
+              </ToolbarGroup>
+            </Toolbar>
+
+            <Pagination
+              params={params}
+              updateParams={(p) => setParams(p)}
+              count={props.allRepositories.length}
```
himdel, line 337: wrong count, allRepositories.lenght will be max 1 page

you're looking for `count` returned by the Repositories list api

(same issue on line 361)

src/components/approve-modal/approve-modal.tsx
```diff
@@ -0,0 +1,370 @@
...
+    // this handles the possible inconsistent state
+    selectedRepos.forEach((repo) => {
+      if (!fixedRepos.includes(repo)) {
+        reposToApprove.push(repo);
+      }
+    });
+
+    const repositoriesRef = props.allRepositories
+      .filter((repo) => reposToApprove.includes(repo.name))
+      .map((repo) => repo.pulp_href);
```
himdel, line 81: this won't work either when the repo is not in the first page of the unfiltered results

`repositoryList` does include the href when selecting the item (but may not contain the item when submitting)

so onSelect should be saving the href and this won't be needed?

src/components/approve-modal/approve-modal.tsx
```diff
@@ -0,0 +1,370 @@
...
+  }, [params]);
+
+  useEffect(() => {
+    const fixedReposLocal = [];
+    const selectedReposLocal = [];
+
+    // check for approval repos that are already in collection and select them in UI
+    // this is handling of situation when collection is in inconsistent state
+    props.collectionVersion.repository_list.forEach((repo) => {
+      const count = props.allRepositories.filter((r) => r.name == repo).length;
```
himdel, line 221: and here, I guess you'll just have to search by exact `?name=` for each

(and you can remove the allRepositories prop)

himdel:
(screenshot test failure expected, new column in approvals :+1:)

Works for me (both with modal and directly), looks great :+1: 

* This is missing the "Select all" logic in the modal
* it should use the `taskAlert` alert for rejection (and approval without modal), otherwise you can get in a state where the collection version is correctly Rejected and the UI looks done, but also showing the old approved repos because that task is still running. (It works correctly, once the task is done and the list reloads the approved repos are gone, so just adding the alert while the task is running is enough.)
* closeAlert closes all alerts, use the shared one
* things break when there are more approval repos than the defaut page size (see code comments)
  * but that size is 100 .. so this pretty much works for most realistic use cases (=> not blocking this PR)

But it probably makes sense to fix the pagination issue before merging.

LGTM :+1: let's merge this now, and address the ux issues ^ in a separate PR?

src/containers/certification-dashboard/certification-dashboard.tsx
```diff
@@ -120,7 +130,21 @@ class CertificationDashboard extends React.Component<RouteProps, IState> {
...
       this.setState({ unauthorized: true });
     } else {
-      this.queryCollections();
+      this.setState({ loading: true });
+
+      Repositories.listApproved()
+        .then((data) => {
+          this.setState({ repositoryList: data.data.results });
+          this.queryCollections();
+        })
```
himdel, line 139: queryCollections doesn't depend on listApproved, can we call them in paralel?
(is this just about `loading`?)

(Also, if loading the repository list fails, the rest of the UI could still work.)

himdel: (Also, querying just the page 1 of repository list is not a good long-term solution, we may need to switch to some kind of memoized lazy loaded per item thing, or go through all the pages.) (But works when <100 repos, so good enough for now)

src/containers/certification-dashboard/certification-dashboard.tsx
```diff
@@ -565,6 +611,140 @@ class CertificationDashboard extends React.Component<RouteProps, IState> {
...
+            Constants.NOTCERTIFIED,
+          )
+            .then(() => {
+              removedRepos.push(repo);
+              return Promise.resolve();
+            })
+            .catch(() => {
+              failedRepos.push(repo);
+              return Promise.resolve();
+            });
```
himdel, line 710:
```suggestion
            .then(() => {
              removedRepos.push(repo);
            })
            .catch(() => {
              failedRepos.push(repo);
            });
```

src/components/approve-modal/approve-modal.tsx
MilanPospisil: But all repositories are from props, from Certification dashboard, when Im loading all the repositories. The actualy loaded repos for the page are in the state variable.

MilanPospisil:
UX was late with the design of select/unselect all. I was tallking to them and we agreed that we can merge this and then repair it later in sepearate PR. Its nothing that will prevent user from using the modal for now. And I have also to solve one cosmetical issue - UX wants to have bottom paging always visible and not part of the scrollable content. I think we should also do this in separate PR.

src/components/approve-modal/approve-modal.tsx
MilanPospisil: allRepositories are actualy really all repositories loaded by certification dashboard.

himdel:
> UX wants to have bottom paging always visible and not part of the scrollable content. I think we should also do this in separate PR.

Don't worry about it too much. The user is selecting the number of items per page, they can always select less, and they won't scroll down if they see pagination on the bottom.

And none of our other screens have sticky pagination - we're not doing UX redesign as part of this effort.

PR #3429, opened 2023-03-16
Repositories & Remotes - Access tab

himdel:
Follows #3144, #3430

Implements the Access tab for Repositories & Remotes,
depends on https://github.com/pulp/pulp_ansible/pull/1338 & https://github.com/ansible/galaxy_ng/pull/1647

Testing:

* remotes - create a new role with all the permissions from the "Collection Remotes" section, but only from that section
* repositories - there's a built-in `galaxy.ansible_repository_owner` role

---

Note, this only adds a working Access tab to both.
Adressing permission checks separately in https://github.com/ansible/ansible-hub-ui/pull/3463

### DIFF
--- a/src/api/ansible-remote.ts
+++ b/src/api/ansible-remote.ts
@@ -44,6 +44,22 @@ class API extends PulpAPI {
   update(_id, _obj) {
     throw 'use smartUpdate()';
   }
+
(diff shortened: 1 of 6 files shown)


himdel:
(reviewed in #3463)

PR #3430, opened 2023-03-16
Repositories - Versions tab

himdel:
Follows https://github.com/ansible/ansible-hub-ui/pull/3144

Implements the Versions tab for Repositories

### DIFF
--- a/src/actions/ansible-repository-version-revert.tsx
+++ b/src/actions/ansible-repository-version-revert.tsx
@@ -1,0 +1,113 @@
+import { Trans, t } from '@lingui/macro';
+import { Button, Modal, Spinner } from '@patternfly/react-core';
+import React, { useState } from 'react';
+import { AnsibleRepositoryAPI } from 'src/api';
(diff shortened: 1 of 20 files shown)


MilanPospisil:
When reverting, there is issue with collections. You can end up with collections in system that are in 0 repositories and I dont see any way in UI to delete them. What about checking for them after revert is done and delete them? I can see them in approval dashboard.

But however, deletion API us using repo, so this may not be even possible:

deleteCollectionVersion(repo, collection) {
    return this.http.delete(
      `v3/plugin/ansible/content/${repo}/collections/index/${collection.namespace.name}/${collection.name}/versions/${collection.latest_version.version}/`,
    );
  }

  deleteCollection(repo, collection) {
    return this.http.delete(
      `v3/plugin/ansible/content/${repo}/collections/index/${collection.namespace.name}/${collection.name}/`,
    );
  }

himdel:
Good catch, @newswangerd mentions there is a cleanup task, so we don't have to worry about orphans now :)

PR #3463, opened 2023-03-27
Repositories & Remotes - access checks

himdel:
Follows #3429 , AAH-1998, AAH-1999

This adds permission checks to Repository and Remotes screens and actions.

Permissions come from https://github.com/newswangerd/galaxy_ng/blob/fe0470ee087c97560579e005a28788570536c334/galaxy_ng/app/access_control/statements/pulp.py#L198

For list views, only superadmin will see actions when object permissions are required (users will still see them if they have the model level permission)
for detail/edit views, /my_permissions will always get loaded before evaluating permissions.
(The assumption is that query() yields an object with .my_permissions[].)

### DIFF
--- a/src/actions/action.tsx
+++ b/src/actions/action.tsx
@@ -1,38 +1,41 @@
 import { Button, DropdownItem } from '@patternfly/react-core';
 import React from 'react';
 import { Tooltip } from 'src/components';
+import { type PermissionContextType } from 'src/permissions';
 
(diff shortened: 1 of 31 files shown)


newswangerd (approved):
@himdel functionality looks good. Sounds like we need to add a distribution owner role.

newswangerd:
> @himdel functionality looks good. Sounds like we need to add a remote owner role.

https://github.com/ansible/galaxy_ng/pull/1654

PR #3464, opened 2023-03-27
Repositories - add helper text, remove obsolete `approved_for_use`, change `is_private` label to `private` field

himdel:
> You don’t need to add the “approved for use” label on the repo screen. The pipeline labels are all we need
> informational boxes/descriptions would be helpful on all the fields on the new forms to let people know what they do (this is especially important with private, pipeline etc on the repo page)
> Having something on the retained versions to let users know to leave it blank to retain all versions would be good

Follows #3144 

This removes the "approved_for_use" label logic, replaced by the Pipelines section.

And adds help text to the fields in AnsibleRepositoryForm:

### DIFF
--- a/src/api/response-types/ansible-repository.ts
+++ b/src/api/response-types/ansible-repository.ts
@@ -7,6 +7,7 @@ export class AnsibleRepositoryType {
   pulp_labels?: { [key: string]: string };
   remote?: string;
   retain_repo_versions: number;
+  private?: boolean;
 
   // gpgkey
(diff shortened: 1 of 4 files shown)


newswangerd (changes requested):

himdel:
Thanks, updated the labels,
and moved "Make private" to a separate section, with its own help text, and using the `private` field, instead of a `is_private` label

newswangerd (approved):

PR #3466, opened 2023-03-28
Repository version - add list of collections

himdel:


newswangerd (approved):
Perfect. Thanks

himdel:
(I'm still hoping to split these into tabs or something like that, but good enough for now, merging :).)

PR #3468, opened 2023-03-28
Repair approval process and modal

MilanPospisil:
No-Issue

After merging of new CollectionAPISearch, Approval modal has now different philosophy. Each row now represents tuple Collection and repo. So for one collection in multiple repos, there will be one row for each repo.

Approve modal not affected, data for it was transformed in master. However this PR is adding ability to select/deselect all/on page.

Certification dashboard now contains those changes:

Rejected repo being loaded using pipeline:rejected filter, no longer contains hardwired 'rejected' repo. However, it reads first rejected repo, completely ignoring another if there are more. But user cant still create rejected repos, so this is probably ok.

Staging repos loaded, so it now supports multiple staging repos.

Rejection process reworked - it now rejecting only one collection from one approved repo into rejected repo. If collection is already in rejected repo (it can be both in rejected and in approved repos), collection is deleted from that approved repo, rather than moved to rejected repo (this will fail if collection is already in rejected repo).

### DIFF
--- a/src/components/approve-modal/approve-modal.tsx
+++ b/src/components/approve-modal/approve-modal.tsx
@@ -17,7 +17,12 @@ import {
   ToolbarItem,
 } from '@patternfly/react-core';
 import React, { useEffect, useState } from 'react';
-import { CollectionVersion, CollectionVersionAPI, Repositories } from 'src/api';
+import {
+  CollectionVersion,
+  CollectionVersionAPI,
+  Repositories,
+  SigningServiceAPI,
+} from 'src/api';
 import { Repository } from 'src/api/response-types/repositories';
 import {
   AlertList,
@@ -29,6 +34,7 @@ import {
   SortTable,
   closeAlert,
 } from 'src/components';
+import { useContext } from 'src/loaders/app-context';
 import {
   errorMessage,
   parsePulpIDFromURL,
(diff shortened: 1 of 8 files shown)


src/components/approve-modal/approve-modal.tsx
```diff
@@ -89,11 +107,24 @@ export const ApproveModal = (props: IProps) => {
...
+
+              if (reapprove) {
+                // reapprove takes first
+                promiseCopyOrMove = Repositories.copyCollectionVersion(
+                  pulp_id,
+                  [data.data.pulp_href],
+                  repositoriesRef,
+                );
+              } else {
+                promiseCopyOrMove = Repositories.moveCollectionVersion(
```
newswangerd, line 120: When auto signing is enabled, you need to include the signing service HREF in the post request.

You can get the name of the signing service from `_ui/v1/settings/` and look up the href on `pulp/api/v3/signing-services/?name=<NAME>`

newswangerd (changes requested):

src/components/approve-modal/approve-modal.tsx
MilanPospisil: Thank you for catching that.

MilanPospisil: Done. But I dont know if this is correct. I have also rewrited approval method to async (handling so many API calls including conditions with then was crazy), so its easier to read.

newswangerd (approved):

src/containers/certification-dashboard/certification-dashboard.tsx
```diff
@@ -675,26 +715,63 @@ class CertificationDashboard extends React.Component<RouteProps, IState> {
...
-        'danger',
-        t`Collection not found in any repository.`,
-      );
-      return;
-    }
+    const originalRepo = collection.repository.name;
+    const version = collection.collection_version;
+
+    this.transformToCollectionVersion(collection)
+      .then((versionWithRepos: any) => {
```
himdel, line 722:
```suggestion
      .then((versionWithRepos) => {
```

himdel:
Merging, so I can use the modal in #3431 :)

PR #3470, opened 2023-03-30
basic landing page

ShaiahWren:
Issue: AAH-2172

This work is the first version of the Beta Galaxy landing page dashboard with footer. The design follows and adds to the mockups created by Anastasia Ratti. 

The 'Cookie preferences' link on the footer (seen on redhat.com) is NOT included in this pr. 

Screenshots

Body: 
Screenshot 2023-04-24 at 11 15 05 AM

Footer:
Screenshot 2023-04-28 at 12 18 14 PM

### DIFF
--- a/src/containers/landing/landing-page.tsx
+++ b/src/containers/landing/landing-page.tsx
@@ -74,7 +74,11 @@ export class LandingPage extends React.Component<RouteProps, IState> {
                   <p>
                     <Trans>
                       Use the{' '}
-                      <a href='https://galaxy.ansible.com/search?deprecated=false&keywords=&order_by=-relevance'>
+                      <a
+                        href='https://galaxy.ansible.com/search?deprecated=false&keywords=&order_by=-relevance'
+                        target='_blank'
+                        rel='noreferrer'
+                      >
                         Search page
                       </a>{' '}
                       to find content for your project, then download them onto
@@ -134,7 +138,11 @@ export class LandingPage extends React.Component<RouteProps, IState> {
                   </b>
                   <br />
                   <p>
-                    <a href='https://www.redhat.com/en/summit/ansiblefest?intcmp=7013a0000034lvmAAA'>
+                    <a
+                      href='https://www.redhat.com/en/summit/ansiblefest?intcmp=7013a0000034lvmAAA'
+                      target='_blank'
+                      rel='noreferrer'
+                    >
                       <img
                         width='100%'
                         alt='Ansible Fest at Red Hat Summit May 23rd to 25th 2023'
@@ -179,24 +187,32 @@ export class LandingPage extends React.Component<RouteProps, IState> {
                       <a
                         className='footer-link'
                         href='https://www.redhat.com/en/about/privacy-policy'
+                        target='_blank'
+                        rel='noreferrer'
                       >{t`Privacy statement`}</a>
                     </span>
                     <span>
                       <a
                         className='footer-link'
                         href='https://www.redhat.com/en/about/terms-use'
+                        target='_blank'
+                        rel='noreferrer'
                       >{t`Terms of use`}</a>
                     </span>
                     <span>
                       <a
                         className='footer-link'
                         href='https://www.redhat.com/en/about/all-policies-guidelines'
+                        target='_blank'
+                        rel='noreferrer'
                       >{t`All policies and guidelines`}</a>
                     </span>
                     <span>
                       <a
                         className='footer-link'
                         href='https://www.redhat.com/en/about/digital-accessibility'
+                        target='_blank'
+                        rel='noreferrer'
                       >{t`Digital accessibility`}</a>
                     </span>
                   </div>
--- a/src/containers/landing/landing-page.tsx
+++ b/src/containers/landing/landing-page.tsx
@@ -1,17 +1,17 @@
 import { t } from '@lingui/macro';
 import * as React from 'react';
-import { SigningServiceAPI, SigningServiceType } from 'src/api';
+import { SigningServiceType } from 'src/api';
 import {
   AlertList,
   AlertType,
   BaseHeader,
-  LoadingPageSpinner,
+  LandingPageCard,
   Main,
   closeAlertMixin,
 } from 'src/components';
 import { AppContext } from 'src/loaders/app-context';
 import { RouteProps, withRouter } from 'src/utilities';
-import { ParamHelper, errorMessage } from 'src/utilities';
+import { ParamHelper } from 'src/utilities';
 
 interface IState {
   params: {
@@ -50,16 +50,8 @@ export class LandingPage extends React.Component<RouteProps, IState> {
     };
   }
 
-  componentDidMount() {
-    if (!this.context.user || this.context.user.is_anonymous) {
-      this.setState({ loading: false, unauthorised: true });
-    } else {
-      this.query();
-    }
-  }
-
   render() {
-    const { loading, alerts } = this.state;
+    const { alerts } = this.state;
 
     return (
       <React.Fragment>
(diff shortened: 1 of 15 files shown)


himdel:
Thanks, yeah, I think this looks good :+1:,  enough to start discussions about what *should* be there :)

Still, when you have time, it might make sense to hide/remove the filter & pagination bits and add a section card with text, whether lorem ipsum or copied from https://galaxy.ansible.com/, because it's more likely we'll only have static content here now.

Then, to get this reading for actually being the page .. it would not get its own menu item .. but clicking the top left logo would go to `Paths.landingPage` instead of `Paths.collections`

And it's still not a landing page if it's not the default page .. which is always the one with `/` for url in our case...
So, you'd then replace any use of `Paths.search` with `Paths.collections` (for all the links that should still be going to collections),
remove `Paths.search`, and change `Paths.landingPage` from `/landing-page` to `/`.

That should get us to the state where this is the landing page everywhere, if we need this to be different per mode,
either we have logic in the router assigning the `/` route to either the landing page component or the collections component, or the landing page component could `navigate` to a the right place based on logic in `componentDidMount`.

src/loaders/insights/routes.tsx
```diff
@@ -155,7 +155,6 @@ const routes = [
   { path: Paths.collection, component: CollectionDetail },
   { path: Paths.namespace, component: NamespaceDetail },
   { path: Paths.collections, component: Search },
-  { path: Paths.search, component: Search },
```
himdel, line 158: (this means no default route for insights mode - it will either need to be the landing page, or a `{ path: '/', component: Search }`)

src/paths.ts
```diff
@@ -84,7 +84,7 @@ export enum Paths {
   myImports = '/my-imports',
   login = '/login',
   logout = '/logout',
-  search = '/',
+  landingPage = '',
```
himdel, line 87:
```suggestion
  landingPage = '/',
```

himdel:
The changes look good :+1:.

But the cards now take up all horizontal space instead of going side by side...
this should work..

```diff
--- a/src/components/cards/landing-page-card.tsx
+++ b/src/components/cards/landing-page-card.tsx
@@ -8,13 +8,14 @@ interface IProps {
 
 export const LandingPageCard = ({ title, body }: IProps) => {
   return (
-    <React.Fragment>
-      <Card className='landing-page-card' style={{ marginBottom: '20px' }}>
+    <Card
+      className='landing-page-card'
+      style={{ margin: '0 0 24px 24px', flex: '30%' }}
+    >
       <CardHeader>
         <CardTitle>{title}</CardTitle>
       </CardHeader>
       <CardBody>{body}</CardBody>
     </Card>
-    </React.Fragment>
   );
 };
--- a/src/containers/landing/landing-page.tsx
+++ b/src/containers/landing/landing-page.tsx
@@ -61,6 +35,14 @@ export class LandingPage extends React.Component<RouteProps, IState> {
         ></AlertList>
         <BaseHeader title={t`Home`} />
         <Main>
+          <div
+            style={{
+              display: 'flex',
+              flexWrap: 'wrap',
+              alignContent: 'flex-start',
+              marginLeft: '-24px',
+            }}
+          >
             <LandingPageCard
               title={t`Lorem Ipsum`}
               body={t`Lorem ipsum dolor sit amet, consectetur adipiscing elit. Aliquam nibh odio, semper non ex vitae, semper convallis tellus. Praesent et ipsum erat. Praesent hendrerit urna eget mattis vestibulum. Maecenas dictum orci vitae nisl sagittis laoreet id et mauris. Sed pharetra accumsan nibh a viverra. Duis tincidunt eros at maximus sodales. Fusce gravida tellus ligula eu posuere lorem placerat ut.`}
@@ -77,6 +59,7 @@ export class LandingPage extends React.Component<RouteProps, IState> {
               title={t`Lorem Ipsum`}
               body={t`Lorem ipsum dolor sit amet, consectetur adipiscing elit. Aliquam nibh odio, semper non ex vitae, semper convallis tellus. Praesent et ipsum erat. Praesent hendrerit urna eget mattis vestibulum. Maecenas dictum orci vitae nisl sagittis laoreet id et mauris. Sed pharetra accumsan nibh a viverra. Duis tincidunt eros at maximus sodales. Fusce gravida tellus ligula eu posuere lorem placerat ut.`}
             />
+          </div>
         </Main>
       </React.Fragment>
     );
```

And you may want to clean up the state logic in LandingPage a bit more .. `alerts` still make sense, but there would be no paging params, no loading/unauthorized state, no items, no inputText...

Oh and can you also change `Paths.search` to `Path.collections` in `developer_guidelines.md`? Just to keep the example up to date :)

---

After that, the only remaining thing will be adding a redirect to collections when not running in community mode, we'll need to add a feature flag to enable the landing page. (similar to https://github.com/ansible/galaxy_ng/pull/1669)

himdel:
Looks great! :) 

The only visual change I'd suggest is moving the "Welcome to Beta Galaxy" text to replace the "Home" title.

Second thing, localization.. most of it is :+1: , but we should not be splitting strings in the middle of sentences,
"Search page", "Use the", "ansible-galaxy", "to find content for your project..", ", the command line tool..." as 5 separate strings are hard to translate, this needs to be one string .. we have the `Trans` component (from lingui/macro) for that, `<Trans>Use the <a href=...>Search page</a> to find....</Trans>` should do just that :)

Otherwise I think this is ready except for the community vs standalone mode logic.. 
I think the final final landing page won't be tied to wisdom/lightspeed, but this one kind of is... so maybe all we need for that is...

```
diff --git a/src/containers/landing/landing-page.tsx b/src/containers/landing/landing-page.tsx
index e63ba73e..64ecb03c 100644
--- a/src/containers/landing/landing-page.tsx
+++ b/src/containers/landing/landing-page.tsx
@@ -10,10 +10,12 @@ import {
   closeAlertMixin,
 } from 'src/components';
 import { AppContext } from 'src/loaders/app-context';
+import { Paths, formatPath } from 'src/paths';
 import { RouteProps, withRouter } from 'src/utilities';
 
 interface IState {
   alerts: AlertType[];
+  redirect: boolean;
 }
 
 export class LandingPage extends React.Component<RouteProps, IState> {
@@ -22,11 +24,24 @@ export class LandingPage extends React.Component<RouteProps, IState> {
 
     this.state = {
       alerts: [],
+      redirect: false,
     };
   }
 
+  componentDidMount() {
+    const { ai_deny_index } = this.context.featureFlags;
+    if (!ai_deny_index) {
+      this.setState({ redirect: true });
+    }
+  }
+
   render() {
-    const { alerts } = this.state;
+    const { alerts, redirect } = this.state;
+
+    if (redirect) {
+      setTimeout(() => this.props.navigate(formatPath(Paths.collections)));
+      return null;
+    }
 
     return (
       <React.Fragment>
```

src/containers/landing/landing-page.tsx
```diff
@@ -0,0 +1,174 @@
...
+                  <p>{t`Maybe you have automation for installing and configuring a popular software package, or for deploying software built by your company. Whatever it is, use Galaxy to share it with the community.`}</p>
+                  <br />
+
+                  <p>
+                    {t`Red Hat is working on exciting new Ansible content development capabilities within the context of`}{' '}
+                    <a
+                      href='https://www.redhat.com/en/engage/project-wisdom?extIdCarryOver=true&sc_cid=701f2000001OH6uAAG'
+                      target='_blank'
+                      rel='noopener noreferrer'
+                    >{t`Project Wisdom`}</a>{' '}
```
himdel, line 92: Project Wisdom -> Ansible Lightspeed

the name of the project changed.. the link stays the same for now, until a better link exists :)

src/containers/landing/landing-page.tsx
```diff
@@ -0,0 +1,215 @@
...
+                    <Trans>
+                      Use the{' '}
+                      <a href='https://galaxy.ansible.com/search?deprecated=false&keywords=&order_by=-relevance'>
+                        Search page
+                      </a>{' '}
+                      to find content for your project, then download them onto
+                      your Ansible host using{' '}
+                      <a
+                        href='https://docs.ansible.com/ansible/latest/reference_appendices/galaxy.html#the-command-line-tool'
+                        target='_blanck'
```
himdel, line 69:
```suggestion
                        target='_blank'
```

src/containers/login/login.tsx
```diff
@@ -29,7 +29,7 @@ class LoginPage extends React.Component<RouteProps, IState> {
     };
 
     const params = ParamHelper.parseParamString(this.props.location.search);
-    this.redirectPage = params['next'] || formatPath(Paths.search);
+    this.redirectPage = params['next'] || formatPath(Paths.collections);
```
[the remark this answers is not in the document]
ShaiahWren: Hmm, I changed this to `Paths.landingPage` and it still redirects to collections - both on login and clicking the GALAXY icon...

src/containers/landing/landing-page.tsx
```diff
@@ -0,0 +1,225 @@
...
+
+                  <p>
+                    {t`Content from roles and collections can be referenced in Ansible PlayBooks and immediately put to work. You'll find content for provisioning infrastructure, deploying applications, and all of the tasks you do everyday.`}{' '}
+                  </p>
+                  <br />
+
+                  <p>
+                    <Trans>
+                      Use the{' '}
+                      <a href='https://galaxy.ansible.com/search?deprecated=false&keywords=&order_by=-relevance'>
```
himdel, line 77: oh, we'll need to update the Search page link to link to the new galaxy Search page, not the old one...

can you use a Link to Paths.collections here?

jerabekjiri:
I wonder if this would be a good idea to be a clickable link to landing page, but I guess not really since the logo is right above it 
Screenshot from 2023-05-03 15-25-52

jerabekjiri:
just a small nitpicks, other than LGTM :+1: :)

newswangerd (approved):

PR #3504, opened 2023-04-04
Approval Modal test

MilanPospisil:
Requires: https://github.com/ansible/galaxykit/pull/87

Issue: https://issues.redhat.com/browse/AAH-2248

Tests for approval process. Two files - one for testing the repo table in modal - paging, filtering, ordering and selection. Other for approval and rejection tests. Those are both added in new folder, so the tests can run more in parallel (approval tests were very slow even before).

### DIFF
--- a/src/containers/certification-dashboard/certification-dashboard.tsx
+++ b/src/containers/certification-dashboard/certification-dashboard.tsx
@@ -454,9 +454,9 @@ class CertificationDashboard extends React.Component<RouteProps, IState> {
 
   private renderRow(collectionData: CollectionVersionSearch, index) {
     const { collection_version: version, repository } = collectionData;
-    const data_cy = `${collectionData.repository.name}-${collectionData.collection_version.namespace}-${collectionData.collection_version.name}`;
+    const data_cy = `CertificationDashboard-row-${collectionData.repository.name}-${collectionData.collection_version.namespace}-${collectionData.collection_version.name}`;
     return (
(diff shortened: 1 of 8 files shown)


src/containers/certification-dashboard/certification-dashboard.tsx
```diff
@@ -454,8 +454,9 @@ class CertificationDashboard extends React.Component<RouteProps, IState> {
 
   private renderRow(collectionData: CollectionVersionSearch, index) {
     const { collection_version: version, repository } = collectionData;
+    const data_cy = `${collectionData.repository.name}-${collectionData.collection_version.namespace}-${collectionData.collection_version.name}`;
     return (
-      <tr key={index} data-cy='CertificationDashboard-row'>
+      <tr key={index} className='CertificationDashboard-row' data-cy={data_cy}>
```
himdel, line 459: wrong direction? Our tests should be using data-cy for selectors, not clasnames.

What about ``data-cy={`CertificationDashboard-row-${collectionData.repository.name}-${collectionData.collection_version.namespace}-${collectionData.collection_version.name}`}`` and using `[data-cy^="CertificationDashboard-row"]` for the old selectors ([`^=`](https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors#attrvalue_4))?

MilanPospisil: Yeah, but I was lazy to rewrite the rest of the tests :-)

PR #3558, opened 2023-04-10
Use `display_repositories` feature flag

himdel:
Depends on https://github.com/ansible/galaxy_ng/pull/1669
Issue AAH-2269

There's a new `display_repositories` feature flag,
without it:
hide repo link in collections view,
hide repo selector,
hide menu item

with it:
show menu item in anonymous mode too


himdel:
Backporting for code consistency.. but the actual feature flag is NOT available in 4.7, hardcoding to true there.

PR #3559, opened 2023-04-10
CollectionDetail: use routeParams.repo, not routeParams.published

himdel:
Fixes typo from #3467 

there's no `:published` param, there is a `:repo` param .. use that one :)


jerabekjiri (approved):
Nice, thanks :+1:

PR #3567, opened 2023-04-11
Fix logo size in collection list

jerabekjiri:
Issue: AAH-2272

before:
Screenshot from 2023-04-11 14-42-14

after:
Screenshot from 2023-04-11 14-10-02

### DIFF
--- a/CHANGES/2272.bug
+++ b/CHANGES/2272.bug
@@ -1,0 +1,1 @@
+Unlock logo size in collection list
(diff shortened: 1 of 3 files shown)


himdel:
We still need to limit the size of the image somehow.

With this change, tall images break - you can try with http://placekitten.com/40/400 (/width/height)

Before:

After:

himdel:
One more note... the `default-logo` image we're using is a 214x214 logo inside a 300x300 transparent image

So there's a chance this particular logo will always look too small until we crop it?

newswangerd (approved):

jerabekjiri:
Screenshot from 2023-04-12 16-28-53
Screenshot from 2023-04-12 16-29-11
Screenshot from 2023-04-12 16-29-26
Screenshot from 2023-04-12 16-29-39

himdel (approved):

PR #3575, opened 2023-04-12
Use mirror:true for repository syncs

himdel:
Issue: AAH-2286

Use `{ mirror: true }` when calling the repository sync endpoint.

When enabled, all content that is not present in the remote repository will be removed from the local repository, otherwise sync will be add missing content.


himdel:
TODO(me): test this works with requirements.yaml

PR #3580, opened 2023-04-13
Prevent deleting protected repositories

jerabekjiri:
Issue: AAH-2300

- close modal if the repository deletion fails 
- Disable delete button if the repository is protected (default) 
Screenshot from 2023-04-13 11-08-06

Protected repositories: https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/access_control/access_policy.py#L465 (https://github.com/ansible/galaxy_ng/pull/1671)


himdel (approved):
Good idea! :+1:

PR #3581, opened 2023-04-13
Select collection screen - multi select, repository field, repo filter

jerabekjiri:
Issue: [AAH-2274](https://issues.redhat.com/browse/AAH-2274)

- multi select and repository column:
Screenshot from 2023-04-13 14-19-51

Screenshot from 2023-04-13 14-18-07

### DIFF
--- a/src/actions/ansible-repository-collection-version-add.tsx
+++ b/src/actions/ansible-repository-collection-version-add.tsx
@@ -75,6 +75,7 @@ const AddCollectionVersionModal = ({
   const renderTableRow = (item: CollectionVersionSearch, index: number) => {
     const {
       collection_version: { name, namespace, version, description },
+      repository,
     } = item;
 
     return (
@@ -91,6 +92,7 @@ const AddCollectionVersionModal = ({
           {namespace}.{name} v{version}
         </td>
         <td>{description}</td>
+        <td>{repository.name}</td>
       </tr>
     );
   };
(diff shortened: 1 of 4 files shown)


src/actions/ansible-repository-collection-version-add.tsx
```diff
@@ -8,29 +8,43 @@ import {
...
       );
-      setState((ms) => ({ ...ms, addCollectionVersionModal: null }));
-      query({});
     })
     .catch(
       handleHttpError(
-        t`Failed to add ${namespace}.${name} v${version} to repository "${repositoryName}".`,
+        t`Failed to add collection${
+          collections.length === 1 ? '' : 's'
+        } to repository "${repositoryName}".`,
```
himdel, line 47: this is problematic when translating to languages which do not use -s for plurals

I guess the proper way with plurals would be...

`import { plural } from '@lingui/macro';`

```suggestion
    plural(collections.length, {
      one: 'Failed to add collection to repository "{repositoryName}".',
      other: 'Failed to add collections to repository "{repositoryName}".',
    }).replace('{repositoryName}', repositoryName)
```

(unless I'm missing a nice way to add interpolation to plurals)

himdel:
The rest looks good :+1:

src/actions/ansible-repository-collection-version-add.tsx
jerabekjiri: Forgot about this, this is much better better, thanks :+1:

jerabekjiri: seems like `.replace('{repositoryName}', repositoryName),` doesn't replace variable in the string, but it works like this  
```
 plural(collections.length, {
    one: `Failed to add collection to repository "${repositoryName}".`,
    other: `Failed to add collections to repository "${repositoryName}".`,
})
```

PR #3595, opened 2023-04-14
Fix collection version selector

himdel:
When a collection has more than 10 versions, accessing http://localhost:8002/ui/repo/comu_yaml/ansible/netcommon/?version=0.0.3-dev8 gives a 404 when the requested version is not in the top 10 (by latest) version results.

=> changing `loadCollection` to load the requested version (by `version` or `is_highest`) separately from the list of versions

but then it either works and the version loads, but there are no other versions in the version selector,
or it 404s and there would be .. because `getCached` will only return the value from one request for both .. caching works :)

=> moving `getCached` logic to `loadCollection`, caching current collection, collection versions and content separately.

And version picker (when more than 10 versions) stopped working, because we're only ever getting 10 results per page.
=> wiring in a separate `collectionsCount` variable with the full version count, and basing the versions modal logic on that.

### DIFF
--- a/src/api/collection-version.ts
+++ b/src/api/collection-version.ts
@@ -1,11 +1,7 @@
 import axios from 'axios';
-import { CollectionVersionSearch } from 'src/api';
 import { HubAPI } from './hub';
 
 export class API extends HubAPI {
-  // contains collection versions
-  cachedCollection: CollectionVersionSearch[] = [];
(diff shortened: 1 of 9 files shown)


jerabekjiri:
I have some issues with version ordering, but that's definitely an API issue.

PR #3597, opened 2023-04-14
Fix EmptyStateFilter not appearing in Namespace detail when filtering by repo/tags/sign state

himdel:
Issue: AAH-2304

Previously `EmptyStateFilter` would only render when no data and filtering by keywords, adding the rest of `CollectionFilter` params.

### DIFF
--- a/CHANGES/2304.bug
+++ b/CHANGES/2304.bug
@@ -1,0 +1,1 @@
+Fix EmptyStateFilter not appearing in Namespace detail when filtering by repo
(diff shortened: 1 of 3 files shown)


jerabekjiri (approved):
LGTM :+1:

jerabekjiri:
Not related to this PR, but adding more than one tag breaks with 
```
compound-filter.tsx:257 
Uncaught TypeError: newParams.push is not a function
```

himdel:
Ah, when 0 or 2+ tags, they get treated as an array of tags,
but with only one tag, it becomes a string.

Wrapping in an array :)

PR #3603, opened 2023-04-14
Restyle collection list item controls to keep alignment

himdel:
Issue: AAH-2303

When repository names vary in length, collection list items (both in namespace detail and in collection list view (but not cards view)) end up with misaligned controls:

##### after

Fixing by unifying the rendering between the two, and rendering as a right-aligned block of column flex items, aligned to the right, with text inside aligned to the left.

* `flex-direction: column` `align-items: flex-end`
  * `flex-direction: column` `align-items: flex-start`
    * synclist switch
    * upload button + dropdown menu
    * 2 lines of text
  * `flex-direction: row` `align-self: flex-start`
    * repository
    * signature badge

### DIFF
--- a/CHANGES/2303.bug
+++ b/CHANGES/2303.bug
@@ -1,0 +1,1 @@
+Restyle collection list item controls to keep alignment despite varying repository name lengths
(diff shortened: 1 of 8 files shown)


jerabekjiri (approved):
very nice, didn't think of this, LGTM :+1:

himdel:
Backported via https://github.com/ansible/ansible-hub-ui/pull/3624 (for some reason the bot didn't catch it, adding label manually)

PR #3607, opened 2023-04-16
Use antsibull-docs-ts to render semantic markup

felixfontein:
Uses the TypeScript library [antsibull-docs](https://www.npmjs.com/package/antsibull-docs) (https://github.com/ansible-community/antsibull-docs-ts) to render Ansible markup, including the new semantic markup.

(I marked this as a draft since I want to wait a bit until antsibull-docs-ts reaches 1.0.0 before making this ready.)

FYI, I haven't managed to test this locally, so no idea how the output actually looks...

### DIFF
--- a/CHANGES/2156.feature
+++ b/CHANGES/2156.feature
@@ -1,0 +1,1 @@
+Collection documentation: support semantic markup.
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,6 +19,7 @@
                 "@types/node": "^16.18.16",
                 "@types/react": "^17.0.2",
                 "@types/react-dom": "^17.0.2",
+                "antsibull-docs": "^1.0.0",
                 "axios": "~1.4.0",
                 "classnames": "^2.3.2",
                 "csstype": "^3.1.2",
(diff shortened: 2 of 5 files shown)


felixfontein:
My *guess* is that https://issues.redhat.com/browse/AAH-2156 is the correct issue to reference here. Can someone confirm that? (CC @awcrosby)

felixfontein:
> Does this depend on any backend/importer changes? Or was this just a dev env issue? Can we help? :)

It does not, I simply didn't got around yet to try to run this.

felixfontein:
I tested this with https://github.com/ansible-collections/community.dns/pull/23, and it works! Would be even cooler if `O()` and `RV()` would be turned into links (whenever possible), but I think that's something for a follow-up PR if someone wants to work on this.

felixfontein:
Screenshot of community.dns.hetzner_dns_record (from https://github.com/ansible-collections/community.dns/pull/23) with this PR:
With this PR
Same module rendered with `master` branch:
Without this PR
You can see that `E(...)`, `O(...)`, and `V(...)` aren't interpreted.

himdel (approved):

felixfontein:
@himdel thanks a lot for reviewing and merging!

While backporting there was a conflict in package-lock.json (where else ;-) ); I've created a manual backport in #3754.

PR #3648, opened 2023-04-24
Fix version modal - load data on pagination

himdel:
Follows #3595,
open a collection detail when the collection has more than 10 versions available.
Open version selector, more versions, go to page 2, :boom: 

the version selector modal still operates under the assumption that collections contains all the versions, not just the first page

making the modal query the api on pagination, have a loading spinner, default page size to 10, not 20 to match the loadCollections request

Cc @jctanner , thanks for finding this! :)

---

Before:
image (1)

After:


jctanner:
ty!

PR #3652, opened 2023-04-25
Namespace detail - fix Sign all

himdel:
Issue: AAH-2308

Before: "sign all" would only sign collections in the repository of the first collection of the currently displayed list.

After: "sign all" only works when a repository is selected

Also updates `findDistroBasePathByRepo` to fix fallback,
allows SignCollectionAPI.sign to find repository by name,
and when using the `myCollectionsByRepo` route, use the repo to preset the repository filter

### DIFF
--- a/src/api/collection.ts
+++ b/src/api/collection.ts
@@ -19,7 +19,8 @@ export function findDistroBasePathByRepo(distributions, repository) {
   const distro = distributions.find(
     (distro) => distro.name === repository.name,
   );
-  return distro ? distro.base_path : distro[0].base_path;
+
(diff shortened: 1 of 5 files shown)


src/api/collection.ts
```diff
@@ -19,7 +19,8 @@ export function findDistroBasePathByRepo(distributions, repository) {
   const distro = distributions.find(
     (distro) => distro.name === repository.name,
   );
-  return distro ? distro.base_path : distro[0].base_path;
```
jerabekjiri, line 22: oh :D, sorry for this :facepalm:

jerabekjiri (approved):
very nice :+1: LGTM

PR #3654, opened 2023-04-26
Add collection upload repository selector

MilanPospisil:
https://issues.redhat.com/browse/AAH-2250


MilanPospisil:
> 

I cleared the error state when repo selection is done, but Im not sure if this is the good way or you intend to do that this way. Maybe we should clear only error if it comes from the API and not validation?

himdel (approved):
Works for me :+1:

PR #3655, opened 2023-04-26
Fix manual approve/reject when repository name is different from distribution base_path

himdel:
when calling the collection version copy/move API, the endpoint accepts source base path and destination base path, those are usually identical to repository name, but not always .. any repo with spaces in the name will have a different base path

This ensures we call the endpoint with actual distribution base paths, not repository names.

(There may be room for improvement, the first thing the API does with the base path is look up the repository ([galaxy_ng/app/api/v3/viewsets/collection.py:239](https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/api/v3/viewsets/collection.py#L239)), but not sure that would be backportable.)

Cc @MilanPospisil good catch finding this :)

### DIFF
--- a/src/api/collection-version.ts
+++ b/src/api/collection-version.ts
@@ -4,25 +4,25 @@ import { HubAPI } from './hub';
 export class API extends HubAPI {
   apiPath = 'v3/plugin/ansible/search/collection-versions/';
 
-  setRepository(
+  move(
     namespace: string,
     name: string,
     version: string,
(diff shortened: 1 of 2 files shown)


MilanPospisil (approved):
LGTM :+1:

MilanPospisil:
Interesting is that error was here for very long time and nobody noticed, because we were working only with default repos :-)

PR #3657, opened 2023-04-27
Disable checkbox for adding collection if collection is already assigned to repository

jerabekjiri:
Issue: AAH-2323

User experience improvement, but for performance cost. It could be a little inefficient to fetch collections from the collection versions list again and match them with modal collection versions. So I think we need to think this through if we want to include this. :)

before:
Screenshot from 2023-04-27 13-49-31

after:
Screenshot from 2023-04-27 13-41-55

### DIFF
--- a/CHANGES/2323.feature
+++ b/CHANGES/2323.feature
@@ -1,0 +1,1 @@
+Disable checkbox for adding collection if collection is already assigned to repository
--- a/src/actions/ansible-repository-collection-version-add.tsx
+++ b/src/actions/ansible-repository-collection-version-add.tsx
@@ -3,6 +3,7 @@ import { Button, Checkbox, Modal } from '@patternfly/react-core';
 import React, { useState } from 'react';
 import {
   AnsibleRepositoryAPI,
+  AnsibleRepositoryType,
   CollectionVersionAPI,
   CollectionVersionSearch,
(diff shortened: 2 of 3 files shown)


himdel:
Looks like a nice improvement :+1:, but yeah, you're right the logic can get a bit ..complex :)

This will only work if the "collection version + repo" entry is in the same page of results when filtering by repo and when not, right?

But .. then again.. we're already showing "repository" in the modal list ...so is there a need for the extra query? Or could we just check that repository is not the current repository when the version matches?

(Though, if collection C version V is already in staging and validated, we should probably be greying out both entries, since that same collection version is already in our repo.)

((I wonder if we could add this as an api filter, or if it would make sense to fire a specific query to check for each C-V pair ...))

jerabekjiri:
> This will only work if the "collection version + repo" entry is in the same page of results when filtering by repo and when not, right?
Yep, that's too bad, thanks to pagination, this will only work sometimes...  


> But .. then again.. we're already showing "repository" in the modal list ...so is there a need for the extra query? Or could we just check that repository is not the current repository when the version matches?

make sense, checking if repository matches should be sufficient enough for now and the cleanest solution, more improvements should be made on API (remove duplicates or c-v pair match)

himdel (approved):

PR #3691, opened 2023-05-01
Fix getRepoUrl so that it takes a parameter and returns the correct url.

jctanner:
Before ...

image

After ...

image

### DIFF
--- a/src/utilities/get-repo-url.ts
+++ b/src/utilities/get-repo-url.ts
@@ -1,10 +1,13 @@
 // Returns the API path for a specific repository
-export function getRepoUrl() {
+export function getRepoUrl(reponame) {
   // If the api is hosted on another URL, use API_HOST as the host part of the URL.
   // Otherwise use the host that the UI is served from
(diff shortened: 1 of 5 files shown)


newswangerd (approved):

src/utilities/get-repo-url.ts
```diff
@@ -1,10 +1,13 @@
...
+export function getRepoUrl(reponame) {
   // If the api is hosted on another URL, use API_HOST as the host part of the URL.
   // Otherwise use the host that the UI is served from
   const host = API_HOST ? API_HOST : window.location.origin;
 
-  return `${host}${API_BASE_PATH}`;
+  if (reponame === 'published') {
+    return `${host}${API_BASE_PATH}`;
+  }
+  return `${host}${API_BASE_PATH}content/${reponame}/`;
```
himdel, line 10: base_path != reponame for repositories with spaces in the name .. which one should it be?  
(ansible-repository-copy may need to query the distribution for the base_path)

also.. is `published` special, or should we treat any `pipeline:published` repo the same way?

PR #3715, opened 2023-05-14
Collection documentation: do not wrongly interpret plugins/modules whose name starts with `_` as deprecated

felixfontein:
I'm not sure whether an issue exists for this I can reference. I spotted this while working on #3607 and noticing that a plugin in a collection I was uploading for testing was wrongly marked as deprecated.

In ansible-core itself, an underscore was interpreted this way also in ansible-doc and the validate-modules sanity test (but nowhere else), but that was changed for 2.15.

Ref: https://github.com/ansible/ansible/pull/79362

### DIFF
--- a/CHANGES/3715.bug
+++ b/CHANGES/3715.bug
@@ -1,0 +1,1 @@
+Collection documentation: do not wrongly interpret plugins/modules whose name starts with `_` as deprecated.
--- a/src/components/render-plugin-doc/render-plugin-doc.tsx
+++ b/src/components/render-plugin-doc/render-plugin-doc.tsx
@@ -316,8 +316,8 @@ export class RenderPluginDoc extends React.Component<IProps, IState> {
     }
   }
 
-  private renderDeprecated(doc: PluginDoc, pluginName: string) {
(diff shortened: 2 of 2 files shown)


felixfontein:
Not sure whether the changelog fragment's filename is OK like this; I took the PR's # since I'm not aware of a AH Jira issue.

himdel:
Good catch :+1: 

AAH 4.6 / AAP 2.3 is core 2.14, so we'll only want this in 4.7+, right?

himdel (approved):

felixfontein:
> AAH 4.6 / AAP 2.3 is core 2.14, so we'll only want this in 4.7+, right?

They weren't really deprecated in ansible-core 2.14 either, it was just that valide-modules somehow assumed so and that ansible-doc showed them as deprecated and at the same time printed their name wrong (it removed the underscore).

Not backporting to 4.6 is probably saving some discussions so ... it's probably better :)

felixfontein:
@himdel thanks for reviewing and merging!

himdel:
Any time :)

PR #3735, opened 2023-05-16
Copy collection version in collection detail screen

jerabekjiri:
Issue: AAH-2255

Screenshot from 2023-05-16 17-43-47
Screenshot from 2023-05-16 17-45-25

### DIFF
--- a/src/components/headers/collection-header.tsx
+++ b/src/components/headers/collection-header.tsx
@@ -35,6 +35,7 @@ import {
   BaseHeader,
   BreadcrumbType,
   Breadcrumbs,
+  CopyCollectionToRepositoryModal,
   DeleteCollectionModal,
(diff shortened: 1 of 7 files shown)


src/components/headers/collection-header.tsx
```diff
@@ -282,6 +287,13 @@ export class CollectionHeader extends React.Component<IProps, IState> {
       >
         {t`Upload new version`}
       </DropdownItem>,
+      <DropdownItem
```
himdel, line 290: do we have a permission to check here?

Either way, I think we need at least `featureFlags.display_repositories` to hide this.
(It might make sense to hide this in insights too if there isn't a permission check.)

himdel (approved):
Works for me :)

PR #3737, opened 2023-05-18
Repository URL - always use distro base_path, show in detail screen

himdel:
The `getRepoUrl` helper went through some changes recently, #3346 & #3691,
but it only works when repository.name === distribution.base_path now.

That's not true for repositories without a distributions, repositories with spaces in name, or any repository where there was a distribution name clash and we created a new name.

=> make Repository Copy CLI configuration query the distribution api and use the actual base_path / fail when no distribution.
=> move loading distributions on the repo detail screen from a LazyDistributions component in DetailTab to the query method, allowing Copy CLI to be disabled on detail screen when no distributions,
=> change page addAlert to support alert ids - used to replace a "loading" alert with an "error"/"success" alert

And start showing the repository URL on the detail screen.

### DIFF
--- a/src/actions/ansible-repository-copy.tsx
+++ b/src/actions/ansible-repository-copy.tsx
@@ -1,25 +1,60 @@
 import { t } from '@lingui/macro';
 import React from 'react';
-import { getRepoUrl } from 'src/utilities';
+import { AnsibleDistributionAPI } from 'src/api';
+import { getRepoURL } from 'src/utilities';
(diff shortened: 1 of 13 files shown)


himdel:
Cc @newswangerd this should add copiable repository URLs in detail screen.

newswangerd:
> Cc @newswangerd this should add copiable repository URLs in detail screen.

This looks good! Thanks for doing this

PR #3714, opened 2023-05-12
Create tests for repositories

MilanPospisil:
https://issues.redhat.com/browse/AAH-2245

PR #3756, opened 2023-05-22
Repository tests - unskip clipboard test, grant permissions

himdel:
Fixes #3736

the repo test was skipped because in headless chrome, copying to clipboard fails with 'Write permission denied.'

grant the permission in before, so the test can pass
(https://github.com/cypress-io/cypress-example-recipes/blob/master/examples/testing-dom__clipboard/cypress/e2e/permissions-spec.cy.js)

(this only works for chrome/chromium)


himdel:
(https://github.com/ansible/ansible-hub-ui/pull/3734 backports both #3714 and this one)

PR #3759, opened 2023-05-24
Delete collection all or single repo

MilanPospisil:
Issue: https://issues.redhat.com/browse/AAH-2261

Depends on https://github.com/ansible/galaxykit/pull/93

This PR Adds:

- Delete collection from repository - this is added into all collection menus (collection list, namespace collection list and collection detail).

- Delete collection version from repository - this is added into collection detail only.

- Both deletes from repository are only visible in standalone, since Insights mode does not have any repo management, this operation is meaningless.

- Added tests for those operations.

- existing "should copy collection version to validated repository" test is fixed

- Delete entire collection menu item name slightly renamed from "Delete entire collection" to "Delete entire collection from system" to better distinguish from "Delete collection from repository"

- deleteNamespacesAndCollections test command slightly changed. Now it does not removes orphan collections (otherwise it fails on them) and does not removes namespaces that has orphan collections. Also task wait all is used instead of more complicated waiting.


src/components/delete-modal/delete-collection-modal.tsx
```diff
@@ -50,6 +53,13 @@ export const DeleteCollectionModal = (props: IProps) => {
                     {deleteCollection.collection_version.name} v
                     {collectionVersion}
                   </b>{' '}
+                  {deleteFromRepo ? (
```
himdel, line 56: these will be impossible to translate. . you have to copy the whole sentence

MilanPospisil: This going to be too much combinations given that there are already several ifs. I will try something different that will not bloat into many combinations.

MilanPospisil: I implemented it as separate sentence at the end of the other sentences, if user deletes from repo, it will add this text: `The collection will be deleted only from repository ${deleteFromRepo}`

src/components/delete-modal/delete-collection-modal.tsx
```diff
@@ -50,8 +52,8 @@ export const DeleteCollectionModal = (props: IProps) => {
                     {deleteCollection.collection_version.name} v
                     {collectionVersion}
                   </b>{' '}
-                  and its data will be lost and this will cause the entire
-                  collection to be deleted.
+                  <>from the system </> and its data will be lost and this will
```
himdel, line 55: bad rebase of some kind? 

What about simplifying these to something like this...

```js
  const lastCollectionVersion = collectionVersion && collections.length === 1;
```

```jsx
        <Text style={{ paddingBottom: 'var(--pf-global--spacer--md)' }}>
          {collectionVersion ? (
            <Trans>
              Deleting{' '}
              <b>
                {deleteCollection.collection_version.name} v{collectionVersion}
              </b>
              , its data will be lost.
            </Trans>
          ) : (
            <Trans>
              Deleting <b>{deleteCollection.collection_version.name}</b>, its
              data will be lost.
            </Trans>
          )}{' '}
          {lastCollectionVersion
            ? t`This will cause the entire collection to be deleted.`
            : null}
          {deleteFromRepo
            ? t`The collection will be deleted only from repository ${deleteFromRepo}.`
            : null}
        </Text>
```

himdel:
We should check the `display_repositories` feature flag for these as well (and then may not need the standalone mode checks).

src/components/delete-modal/delete-collection-modal.tsx
MilanPospisil: You are right, I have to rewrite this better. This way, it starting to be combinatory explosion, 2 types of repos (all/from repo), 2 types of version/whole and 2 types one version/last version. I will try to separate some strings, let me then know what you think.

MilanPospisil: I now looked more deeper into your example and this is exactly what I ment in previous comment, so I will use your suggestion, thanks :-)

himdel (approved):

PR #3778, opened 2023-05-29
Lazy-load src/containers in both modes

himdel:
Insights mode has been lazy-loading containers from the start, whiile standalone and community load everything in one big bundle.

This unifies the modes, making all imports from src/containers lazy (except for router importing notfound),
adds suspense to standalone routes to deal with the lazy wait.


himdel:
the screenshot diff should be fine, we're back to consistent alignment between compound filter and applied filters, but is a change from what has probably been there since #507

PR #3781, opened 2023-05-29
prettier: sync with aap ui

himdel:
we already have semi and single-quote enabled, adding printWidth to match https://github.com/ansible/ansible-ui/blob/main/.prettierrc.json


himdel:
holding off until there's less open PRs, this will mean everybody is rebasing

PR #3783, opened 2023-05-30
UI API bindings deduplicate

MilanPospisil:


himdel:
uhh, this is going the wrong way I think..

* a repository can be a pulp_ansible repository or a pulp_container repository .. hence AnsibleRepository
* all our API classes are named SomethingAPI

=> rename Repository to AnsibleRepositoryAPI please? (And use AnsibleRepositoryType)

MilanPospisil:
> 

Lets close it, its better to do it from scratch and make agreement what should be the result of the next PR.

PR #3784, opened 2023-05-30
Respect `display_repositories` feature flag for collection/namespace/role filters list

himdel:
Issue: AAH-2326

Filters for legacy roles, collections search, namespace detail and legacy namespaces all use CollectionFilter,
but the logic of whether to display repository filter was based only on static ignoreParams.

Use the feature flag too, hiding the Repository filter when `display_repositories` is off.

(4.7 doesn't have the flag, not backporting)


himdel:
Cc @jctanner

PR #3787, opened 2023-05-30
Repository list - add columns, filters

himdel:
Issue: AAH-2271, AAH-2340

* [x] Combine sync status and last sync into a single column. This should still show the last sync status and date. The date could be a popover. If no remote is set on the repository, this column should say something like "no remote". If a remote is set, but a sync has never been performed this should say something like "never synced".

|before|after|
|-|-|
|||

* [x] Add a "Labels" column that includes the pipeline and hide from search label
* [x] Add a "Private" column that indicates if the repo is private or not

* [x] Add a "Pipeline" filter that lets the user select approved or staging repositories
* ~~Add a "Private" filter~~ (if the API doesn't support this, you can skip it for now)
  * https://beta-galaxy.ansible.com/api/v3/swagger-ui/#/Repositories%3A%20Ansible/repositories_ansible_ansible_list
* [x] Add a "Remote" filter that lets a user select None or a pick a remote from the list of configured remotes.

### DIFF
--- a/CHANGES/2271.feature
+++ b/CHANGES/2271.feature
@@ -1,0 +1,1 @@
+Repository list - add columns, filters
--- a/CHANGES/2340.bugfix
+++ b/CHANGES/2340.bugfix
@@ -1,0 +1,1 @@
+Repositories, Remotes - clean filter text on search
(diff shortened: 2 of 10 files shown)


MilanPospisil (approved):
LGTM :+1:

PR #3843, opened 2023-06-09
Fix translations

MilanPospisil:


himdel:
ok, if this is only missing extract&compile, we can close this, it will happen automagically over the weekend :+1:

PR #3878, opened 2023-06-19
[stable-4.6] Use only namespace in ee owners tab

jerabekjiri:
Issue: AAH-2122

Filter out the namespace part if the container is in the format `<namespace>/<name>`

### DIFF
--- a/CHANGES/2122.bugfix
+++ b/CHANGES/2122.bugfix
@@ -1,0 +1,1 @@
+Filter out namespace name from container name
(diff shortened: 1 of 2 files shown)


himdel:
So.. the problem is that the namespace apis expect the namespace name instead of the whole container name, and the namespace is identical to container name when no slashes, and identical to the first component when there are slashes?

If so, :+1: .. but, what happens in 4.7 and master?

(Any chance the `namespace/name` would still work when properly escaped? (if it works in 4.7 and not 4.6, that could be the reason))

jerabekjiri:
@himdel 
> So.. the problem is that the namespace apis expect the namespace name instead of the whole container name, and the namespace name is identical to the container name when no slashes, and identical to the first component when there are slashes?

yes, that's exactly correct :+1: 

> If so, +1 .. but, what happens in 4.7 and master?

It uses different APIs and therefore works correctly https://github.com/ansible/ansible-hub-ui/blob/master/src/containers/execution-environment-detail/execution_environment_detail_access.tsx#L214

>(Any chance the namespace/name would still work when properly escaped? (if it works in 4.7 and not 4.6, that could be the reason))

I don't think that would help, since this depends on the submitted container name when the container is created, which we can't really control.

himdel (approved):

himdel:
Ah, thanks! Different APIs explain it :)

PR #4009, opened 2023-07-19
Remove alert from Task list view

bmclaughlin:
Removed check for `core.view_task` permission that displayed an alert.  All authenticated users are able to [list tasks](https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/access_control/statements/pulp.py#L583) and [retrieve task details](https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/access_control/statements/pulp.py#L588) .

Issue: AAH-2438

### DIFF
--- a/CHANGES/2438.bug
+++ b/CHANGES/2438.bug
@@ -1,0 +1,1 @@
+Removed alert from task view page stating users could only see their tasks.  All authenticated users can list and retrieve task details.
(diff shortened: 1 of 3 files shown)


himdel (approved):
LGTM :+1: 

thanks for fixing the tests too :)

(ignore insights tests, those are currently broken for unrelated reasons)

PR #4012, opened 2023-07-21
Repository/Remote delete: wait for task before requery, redirect to list screen instead of 404 after delete on detail

himdel:
For both Ansible Repositories and Ansible Remotes,

before, on a list screen: delete triggers a pulp task and reloads the list - the list may update before the delete task finished
after: UI waits for pulp task to finish before reloading

before, on a detail screen: delete triggers a pulp task and reloads the detail - can either update before deletion or successfully 404
after: UI waits for pulp task to finish before redirecting to list screen


himdel:
Cc @nixocio

nixocio (approved):

PR #4106, opened 2023-08-07
Add "eda" tag to list of filterable tags

himdel:
matches https://github.com/ansible/galaxy-importer/pull/233

allows users to see "eda" in the list of tags to filter by

### DIFF
--- a/src/constants.tsx
+++ b/src/constants.tsx
@@ -40,6 +40,7 @@ export class Constants {
     'application',
     'cloud',
     'database',
+    'event_driven_ansible_content',
     'infrastructure',
     'linux',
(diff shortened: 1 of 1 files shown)


src/constants.tsx
```diff
@@ -40,6 +40,7 @@ export class Constants {
     'application',
     'cloud',
     'database',
+    'event_driven_ansible_content',
```
bmclaughlin, line 43: This was updated to `eda` in https://github.com/ansible/galaxy-importer/pull/233

himdel: Thanks, updated :)

PR #4107, opened 2023-08-08
Add autocomplete=off to login form fields, password fields

himdel:
Fixes AAH-2625

This PR:

* forks the patternfly 4 `LoginForm` component, to allow adding `autoComplete=off`
* adds `autoComplete=off` to both the username and password fields on the login form, to match https://github.com/ansible/awx/pull/14232
* adds `autoComplete=off` to any other password fields in forms

([docs](https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion#the_autocomplete_attribute_and_login_fields))


himdel:
tested in firefox, chrome

PR #4122, opened 2023-08-14
[stable-4.7] Rename master to main in workflows and docs

himdel:
and sometimes to stable-4.7 when it makes sense to prefer over main

---

AAH-2643

4.2: #4125
4.5: #4124
4.6: #4123
4.7: #4122
master/main: #4126

### DIFF
--- a/.github/workflows/automerge.yml
+++ b/.github/workflows/automerge.yml
@@ -2,7 +2,7 @@ name: Automerge
 
 on:
   pull_request_target:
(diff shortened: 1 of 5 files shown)


himdel:
(Intentionally) remaining master references:

```
src/containers/legacy-roles/legacy-role.tsx
162:            description={t`The role is versionless and will always install from the head/main/master branch.`}
```

PR #4142, opened 2023-08-21
Approval -  fix when more than 10 staging repos

himdel:
Issue: AAH-2366

`MultipleRepoSelector` deals with displaying a list of repositories and allowing the user to select some,
used by `ImportModal` - single select from list of staging/non-pipelined repos,
by `ApproveModal` - multiselect from list of approved repos,
and `CopyCollectionToRepositoryModal` - multiselect of all repos.

None of those screens (and also Approvals list itself) now expect to be able to load all the repositories, relying on targeted queries for count, for the right repo when only one, and only listing all in the list screen, paginated.

This also disables the "Select all" checkbox, only leaving the "Select page" one, and deselects;
changes `CopyCollectionToRepositoryModal` to use `MultipleRepoSelector` instead of a custom copy;
merge common bits of `ApproveModal` & `CopyCollectionToRepositoryModal` into a new `MultiRepoModal`;
simplify approval status badge to just read repository pipeline without extra queries;
pass repositories as whole object instead of just names where useful.


nixocio (approved):
👍 Good job @himdel.

PR #4145, opened 2023-08-26
screenshot tests: add more screens, blackout time, enable legacy

himdel:
Change screenshot folder from `cypress/screenshots/test_name.js/` to `cypress/screenshots/test_suite_name/test_name.js/` (test_suite meaning all the `cypress/e2e/` subdirs used in test matrix).

Wrap our `DateComponent` in a `<time>` element, following https://microformats.org/wiki/datetime-design-pattern,
blackout `time` elements in non-failure cypress screenshots - we can test more screens.

Enable legacy roles for the screenshots run, allowing us to test those screens.


himdel:
(merging and cleaning the cache, so the cron action can recreate the new cache for #4143)

PR #4159, opened 2023-08-30
Import modal: better error for Invalid file format.

himdel:
we're checking the mime-type before allowing collection upload, this mostly works, but in some cases (see AAP-15541), it can fail unexpectedly.

This should make debugging that case easier, by including the detected type in the error message:


himdel:
might be a good idea to make the message ignorable, as per https://developer.mozilla.org/en-US/docs/Web/API/File/type, we might not want to rely on the type detection too much...

PR #3708, opened 2023-05-09
Synclist deprecation

ShaiahWren:
This pr addresses some of the issues in https://issues.redhat.com/browse/AAH-2335.

PR #4198, opened 2023-09-09
Token insights - restore MyDistribution query, show separate synclist url

himdel:
Restores bits from #3708, #3346 and #3893.

`TokenInsights` now queries  `_ui/v1/my-distributions` to find a `synclist` distribution `base_path`, and display that:

(TODO success .. might have to test on stage)


himdel:
TODO also add validated url

PR #4216, opened 2023-09-11
Search page

himdel:
AAH-2697

modelled after the Dispatch page (#4090), but replacing the 404 bits with a search bar

design: 

untitled-2023-09-11-1554

---

empty, loading, error:

landing page w/ search:


jctanner:
Technically it should probably show a section for v1 namespaces AND for v3, if it's going to show namespaces at all.

jctanner (approved):

rochacbruno (approved):

PR #4283, opened 2023-09-20
RoleRatings, CollectionRatings - display scores on role list/detail and collection detail

himdel:
AAH-2289 & AAH-2652

When available, display rating scores on role list detail & collection detail screens.

Role list:

Ratings tooltip (on hover):

Ratings popover (on click):

Role detail header:

Collection list (item only, not card):

Collection detail header:


config/community.dev.webpack.config.js
```diff
@@ -1,5 +1,8 @@
 const webpackBase = require('./webpack.base.config');
 
+const collectionRatings = require('../static/scores/collection.json');
+const roleRatings = require('../static/scores/role.json');
```
rochacbruno, line 4: So you are using only these 2 files? in the next update I can include only the 2?

himdel: As long as they stay small enough, absolutely :)

Chrome reports the transferred size (gzipped) as 16.8 kB and 40.9 kB, so it's faster to download and cache the whole thing than do per-item requests.

nixocio (approved):
@himdel There is a cypress test failing. Not sure this one is relevant. Other than that good job.

image

himdel:
Thanks, yup, but this time it's a non-change:

```
screenshot screenshots-main/__signature-keys.png changed: 292
292.555
```

`diff__signature-keys.png`:

diff--__signature-keys

PR #4286, opened 2023-09-21
Link legacy to provider namespace

jctanner:
In the legacy namespace summary and legacy role lists, "Provided by" now links to the v3 namespace where RBAC for the legacynamespace is actually managed. If no v3 namespace is linked, it'll be an empty string ... which is fine because we're not supposed to have that situation and it should be fixed for the user.

### DIFF
--- a/src/api/response-types/legacy-namespace.ts
+++ b/src/api/response-types/legacy-namespace.ts
@@ -3,6 +3,7 @@ export class LegacyNamespaceListType {
   url: string;
   summary_fields: {
     owners?: { username: string }[];
+    providers?: { id: number, name: string, pulp_href: string}[];
   };
   created: string;
   modified: string;
@@ -17,6 +18,7 @@ export class LegacyNamespaceDetailType {
   url: string;
   summary_fields: {
     owners?: { username: string }[];
+    providers?: { id: number, name: string, pulp_href: string}[];
   };
   created: string;
   modified: string;
--- a/src/components/legacy-namespace-list/legacy-namespace-provider.tsx
+++ b/src/components/legacy-namespace-list/legacy-namespace-provider.tsx
@@ -1,0 +1,41 @@
+import React from 'react';
+import {
+    Text,
+    TextContent,
+    TextVariants
+} from '@patternfly/react-core';
+import { Trans } from '@lingui/macro';
+import { Link } from 'react-router-dom';
+
+
+interface IProps {
+    id: number;
+    name: string;
+    url: string;
+}
+
+export class ProviderLink extends React.Component<IProps> {
+
+    constructor(props) {
+        super(props);
+    }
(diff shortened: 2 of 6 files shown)


src/utilities/legacy-namespace.ts
```diff
@@ -0,0 +1,31 @@
...
+      id: null,
+      name: null,
+      url: null,
+    };
+  }
+
+  return {
+    id: provider.id || null,
+    name: provider.name || null,
+    url: provider.name ? `/namespaces/${provider.name}` : null,
```
himdel, line 29: urls for Link need to be produced by `formatPath`, otherwise they won't work in all the environments..

```suggestion
    url: provider.name ? formatPath(Paths.legacyNamespace, { namespaceid: provider.name })` : null,
```

(`import { Paths, formatPath } from 'src/paths';`)

src/api/response-types/legacy-namespace.ts
```diff
@@ -3,6 +3,7 @@ export class LegacyNamespaceListType {
   url: string;
   summary_fields: {
     owners?: { username: string }[];
+    providers?: { id: number; name: string; pulp_href: string }[];
```
jerabekjiri, line 6: Legacy Namespaces API returns `provider_namespaces`

src/components/legacy-namespace-list/legacy-namespace-provider.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { Link } from 'react-router-dom';
+
+interface IProps {
+  id: number;
+  name: string;
+  url: string;
+}
+
+export class ProviderLink extends React.Component<IProps> {
+  constructor(props) {
```
jerabekjiri, line 13: just a nitpick, but you can remove the constructor. It's not required and only needed when you initialize local state in the component.

src/components/legacy-role-list/legacy-role-item.tsx
```diff
@@ -5,16 +5,15 @@ import {
...
-  TextVariants,
 } from '@patternfly/react-core';
 import React from 'react';
 import { Link } from 'react-router-dom';
 import { LegacyRoleDetailType } from 'src/api';
 import { DateComponent, DownloadCount, Logo, Tag } from 'src/components';
+import { ProviderLink } from 'src/components/legacy-namespace-list/legacy-namespace-provider';
 import { Paths, formatPath } from 'src/paths';
 import { chipGroupProps } from 'src/utilities';
+import { getProviderInfo } from 'src/utilities/legacy-namespace';
```
jerabekjiri, line 16: nitpick: You could import from `/src/utilities/` like most of the other utilities do.

jctanner:
cypress test failures are unrelated.

PR #4302, opened 2023-09-26
Namespace - show name, not company (outside insights)

himdel:
for non-insights mode, just show namespace name always, instead of preferring namespace.company for the title

Affects collection header (logo alt), collection list item, collection card, namespace header and namespace card.

Issue: AAH-2722

| | | | |
|-|-|-|-|
|Before||||
|After||||

| | |
|-|-|
|Before||
|After||

### DIFF
--- a/CHANGES/2722.fix
+++ b/CHANGES/2722.fix
@@ -1,0 +1,1 @@
+Show namespace name, not company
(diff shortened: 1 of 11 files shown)


rochacbruno (approved):

nixocio (approved):
@himdel cypress screenshot is also failing. Other than that 👍

himdel:
Thanks, yeah, with repositories it's a false positive, but with namespaces it's a good failure .. noticed an actual change - since we're removing the company field on the list screen :)

PR #4305, opened 2023-09-27
Ensure community users can delete their collections

bmclaughlin:
Community users do not have the model permission `ansible.delete_collection`, instead we rely on community users having object permissions on their namespace to determine if they have permissions to delete collections. 

Issue: AAH-2632

Links
----------------
PR (merged) implementing this same change in the api:
https://github.com/ansible/galaxy_ng/pull/1900

### DIFF
--- a/CHANGES/2632.bug
+++ b/CHANGES/2632.bug
@@ -1,0 +1,1 @@
+Ensure beta-galaxy users can delete their collections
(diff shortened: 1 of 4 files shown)


himdel (approved):
LGTM :+1: 

(though I haven't tested this yet)

Maybe we don't need the `ai_deny_index` (=is community) logic, because this should work like this (`hasPermission(...) || hasObjectPermission(......)`) in all the modes? Unless I'm missing something community specific?
(Either way LGTM, we can add this for standalone in a second PR when there's less things to worry about :))

Except some of the test failures still look real...

PR #4366, opened 2023-10-03
Language switcher

himdel:
Issue: AAH-2730; [discord](https://forum.ansible.com/t/how-can-i-change-language-in-new-galaxy-ng-site/1223)

This adds a language switcher to standalone navbar:

closed

opened, using browser default language

manualy switched

manually switched back

(reset to browser defaults leads back to second screenshot :))

(the `(current)`, `(browser default)` and `Reset to browser defaults` strings would be translated once we have the right strings in the catalog)

This uses the language-switching mechanism introduced in #3225, updating it to not pass ?lang to API calls.

### DIFF
--- a/CHANGES/2730.misc
+++ b/CHANGES/2730.misc
@@ -1,0 +1,1 @@
+community: add language switcher
--- a/src/components/index.ts
+++ b/src/components/index.ts
@@ -117,6 +117,7 @@ export { DetailList } from './shared/detail-list';
 export { Details } from './shared/details';
 export { DownloadCount } from './shared/download-count';
 export { ExternalLink } from './shared/external-link';
+export { LanguageSwitcher } from './shared/language-switcher';
 export { LoginLink } from './shared/login-link';
 export { MultiRepoModal } from './shared/multi-repo-modal';
 export { MultiSearchSearch } from './shared/multi-search-search';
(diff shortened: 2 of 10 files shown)


himdel:
options: 

* use translated language names ("English, Spanish" & "Ingles, Espanol")
* use original language names ("English, Espanol")
* use flag emojis ("🇬🇧, 🇪🇸") (possibly problematic for languages used in multiple countries)

felixfontein:
> * use translated language names ("English, Spanish" & "Ingles, Espanol")
> * use original language names ("English, Espanol")

I would use the original language names. The chance that someone figures out that this is the language selector and finds a language they understand is a lot higher if they can at least read the names of the languages they understand :)

(Just think of the UI being in a language and script you do not know, and you cannot even recognize one single language name since they are all in an unknown script. For example look at https://www.wikidata.org/wiki/Q1860#sitelinks-wikipedia and see in how many of the languages you can recognize that the article is about the English language :) )

> * use flag emojis ("🇬🇧, 🇪🇸") (possibly problematic for languages used in multiple countries)

Using flag emojis is a huge antipattern. There are countries with multiple languages (Canada, Switzerland, ...), and as you mentioned there are languages that are used in multiple countries (English, French, Spanish, German, ...).

You can find some concrete examples here: https://www.flagsarenotlanguages.com/blog/why-flags-do-not-represent-language/

> * ISO 639-1 ("en", "es")

That would work, but it looks a bit ugly ;-) What you can always do is add the ISO 639-1 code in brackets next to the language name (like `English (EN)`) - then you can even translate the language name without causing too many headaches (though mostly only because this site has a technical audience - I would still stick to untranslated language names).

himdel:
Updated to use original language names

PR #4368, opened 2023-10-05
Roles - update filters, add sort

himdel:


himdel:
(fixed by https://github.com/ansible/ansible-hub-ui/pull/4424)

PR #4391, opened 2023-10-12
Use a single variable for all access.redhat.com docs URLs

himdel:
EEs were the only one linking further down into the docs, but that link 404s now (https://access.redhat.com/documentation/en-us/red_hat_ansible_automation_platform/2.4/html-single/managing_containers_in_private_automation_hub/index)

=> adding a `UI_DOCS_URL` configuration var, set to https://access.redhat.com/documentation/en-us/red_hat_ansible_automation_platform/


nixocio (approved):
💯

PR #4392, opened 2023-10-12
Remote form: warn about signed_only enabled & missing requirements for Galaxy remote

himdel:
Issue: AAH-2748

also related to #4385 and thus AAH-2360

When editing the community remote (matched by Galaxy URL), warn about `signed_only` enabled and `requirements_yaml` empty.

Previously there were no warnings, these are valid settings that just don't work together with community galaxy.


himdel:
Cc @nixocio , you may want to add these to ansible-ui Remotes in some form?

PR #4424, opened 2023-10-22
Remove duplicate trailing slash on fetching legacy role versions and content

jctanner:
The UX on beta is fetching `/api/v1/roles/{id}/versions//` and `/api/v1/roles/{id}/content//`. On a local devstack the API routing is handling this fine and returning the data, but in the hosted environments it's giving 404's.


jerabekjiri (approved):
LGTM, I'm surprised we didn't notice this before :face_with_spiral_eyes:

PR #4425, opened 2023-10-22
Show the last+newest legacyrole version in the roles list.

jctanner:


jctanner:
Actually, the api is supposed to be reversing this list and only showing up to 10 per on the list page.

PR #4439, opened 2023-10-24
Drop unused cypress transitive dependency

himdel:
We depend on cypress in `test/`,
but this is a transitive dependency of frontend-components-config,
and we're currently stuck on an older version of fcc..
=> removing that copy of cypress.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1935,16 +1935,6 @@
                 "node": ">=6.9.0"
             }
         },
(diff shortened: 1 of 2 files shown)


himdel:
(check this has no effect on crc stage)

himdel:
(backports not necessary, only master depends on `frontend-components` (as opposed to frontend-components-config))

PR #4440, opened 2023-10-24
Roles: prefer namespace over github_user

himdel:
Follows #4423 

use `namespace` instead of `github_user`/`username` where applicable
(and a bit more legacy -> standalone renaming)


src/containers/ansible-role/role-detail.tsx
```diff
@@ -262,6 +264,7 @@ class AnsibleRoleDetail extends React.Component<RouteProps, RoleState> {
       );
     }
 
+    // TODO?
```
himdel, line 267: this one is :+1:

PR #4441, opened 2023-10-24
Dispatch: redirect on 1 item

himdel:
https://forum.ansible.com/t/improvements-to-new-galaxy-s-404-error-pages-and-redirects/1888

when the dispatch page finishes loading both collection results and role results,
and there is exactly 1 result total,
go there directly.


nixocio (approved):
LGTM

ShaiahWren (approved):
Works:)

PR #4442, opened 2023-10-25
Docs - render object choices

himdel:
https://github.com/ansible/ansible/pull/81951#issuecomment-1772123886

In docs renderer, `choices` used to be an array of string values, that's still permissible,
but now we also accept an object of `{ value: description }` instead of the array.

This also changes the rendering of choices to be closer to https://docs.ansible.com/ansible/devel/collections/ansible/builtin/config_lookup.html#parameter-on_missing
=> monospace the choice value, "`value`: description"

Before - Array:

Before - Object:

Now - Array:

Now - Object:

(also fixed the error message...

)

### DIFF
--- a/src/components/render-plugin-doc/render-plugin-doc.tsx
+++ b/src/components/render-plugin-doc/render-plugin-doc.tsx
@@ -1,5 +1,6 @@
+import { Trans, t } from '@lingui/macro';
 import { dom, parse } from 'antsibull-docs';
-import React from 'react';
+import React, { ReactNode } from 'react';
 import {
   PluginContentType,
   PluginDoc,
@@ -29,6 +30,18 @@ interface IProps {
     section: string,
   ) => React.ReactElement;
   renderWarning: (text: string) => React.ReactElement;
+}
+
+function Choice({ children }: { children: ReactNode }) {
+  return <pre style={{ display: 'inline-block' }}>{children}</pre>;
+}
+
+function Legend({ children }: { children: ReactNode }) {
+  return (
+    <div style={{ display: 'inline-block', verticalAlign: 'top' }}>
+      {children}
+    </div>
+  );
 }
 
 export class RenderPluginDoc extends React.Component<IProps, IState> {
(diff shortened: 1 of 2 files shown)


src/components/render-plugin-doc/render-plugin-doc.tsx
```diff
@@ -667,21 +675,27 @@ export class RenderPluginDoc extends React.Component<IProps, IState> {
...
-                  {c === defaul ? (
+                  {c === defaultChoice ? (
                     <span className='blue'>{c} &nbsp;&larr;</span>
                   ) : (
                     c
                   )}
+                  {legends[c] ? (
+                    <>
+                      {' - '}
+                      {legends[c]}
```
felixfontein, line 686: `legends[c]` can be either a string or a list of strings (if it's something else it is broken), and it can contain Ansible markup.

I would probably normalize it to arrays in `parseOptions`, and do something like
```
            {legends[c].map((d) => (
              <p>{this.applyDocFormatters(d)}</p>
            ))}
```
with it. Or maybe (to avoid strange formatting) don't add `<p>`, and replace `</p>` by `<br>`? :shrug:

himdel:
@felixfontein can you re-check please?



I updated choices to monospace the value, and show the potentially multiline (& formatted) description after.

src/components/render-plugin-doc/render-plugin-doc.tsx
```diff
@@ -644,47 +657,86 @@ export class RenderPluginDoc extends React.Component<IProps, IState> {
...
                   )}
+                  {this.renderLegend(legends[c])}
                 </li>
               ))}
             </ul>
           </div>
         ) : null}
 
-        {defaul && !choices.includes(defaul) ? (
+        {defaultChoice && !choices.includes(defaultChoice) ? (
```
felixfontein, line 731:
```suggestion
        {defaultChoice !== undefined && !choices.includes(defaultChoice) ? (
```
in case the default is `0`, `false`, or `''`.

felixfontein (commented):
Besides that, LGTM!

felixfontein (commented):
I haven't tested this and I'm not that familiar with React, but the way I understand the code it looks good :)

felixfontein (approved):
(Should have been an 'approve' review :D )

PR #4457, opened 2023-10-30
Execution environment form fixups

himdel:
Noticed when reviewing https://github.com/ansible/ansible-ui/pull/1054

When editing an EE:

* tags to include/exclude allows adding an empty tag
  * => remove falsy tags when adding
* "and a up to one (slash)" -> "and zero or one (slash)"
* clicking on the LabelGroup "N more" and "show less" buttons submits the form (https://github.com/ansible/ansible-ui/pull/1054#issuecomment-1786179561)
  * => wrap LabelGroup to fix that, and inline chipGroupProps
  * (we should fix this *in* pf5, but this fixes it for pf4)


nixocio (approved):
LGTM

PR #4458, opened 2023-10-31
Import modal - respect display_repositories, hide selector when no choices

himdel:
When the `display_repositories` feature flag is disabled, don't show the "Staging Repos"/"All Repos" radios, only allow staging.

When there is only 1 candidate choice within the current list, don't show the whole list, only the preselected repo.

Thus...

`display_repositories == false, staging count == 1`:

`display_repositories == true, staging count == 1`:

`display_repositories == false, staging count > 1`:

`display_repositories == true, staging count > 1`:

### DIFF
--- a/src/components/import-modal/import-modal.tsx
+++ b/src/components/import-modal/import-modal.tsx
@@ -1,5 +1,12 @@
 import { t } from '@lingui/macro';
-import { Button, Modal, Radio } from '@patternfly/react-core';
+import {
+  Button,
+  Flex,
+  FlexItem,
+  Label,
+  Modal,
+  Radio,
+} from '@patternfly/react-core';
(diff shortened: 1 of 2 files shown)


gundalow:
`display_repositories == false, staging count == 1:` screenshot is a lot clearer, thank you.

jctanner (approved):

src/components/import-modal/import-modal.tsx
```diff
@@ -1,5 +1,13 @@
 import { t } from '@lingui/macro';
-import { Button, Modal, Radio } from '@patternfly/react-core';
+import {
+  Button,
+  Flex,
+  FlexItem,
+  Label,
+  LabelGroup,
```
nixocio, line 7: This one should be imported from https://github.com/ansible/ansible-hub-ui/pull/4457 once this one is merged.

PR #4507, opened 2023-11-11
collection list - use HubListToolbar, add sort

himdel:
Follows #4368, #896, #4506, #4520, #4525
Related to AAH-2802, AAH-850

Adds `CardListSwitcher` support to `HubListToolbar`,
revamp `CollectionFilter` -> `collectionFilter` to return props for `HubListToolbar`,
replace `Search` & `NamespaceDetail` toolbars with `HubListToolbar`.

And allow sorting collections by name, namespace, pulp_created and version.


src/components/collection-list/collection-filter.ts
```diff
@@ -0,0 +1,74 @@
...
+      inputType: 'select' as const,
+      options: [
+        { id: 'true', title: t`Signed` },
+        { id: 'false', title: t`Unsigned` },
+      ],
+    },
+  ].filter(Boolean);
+
+  // TODO name, date & download counts?
+  const sortOptions = [{ title: t`Name`, id: 'name', type: 'alpha' as const }];
```
himdel, line 64: TODO more sorts

src/components/collection-list/collection-filter.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  return {
+    filterConfig,
+    sortOptions,
+    typeaheads: {
+      repository_name: loadRepos,
```
himdel, line 70: TODO verify/fix typeaheads

PR #4540, opened 2023-11-20
Release 4.9.0

jerabekjiri:
No-Issue


himdel (approved):

himdel:
There's a bunch of new files in `CHANGES` so you'll want redo the towncrier parts anyway, but also...

there is no AAH-3715 .. `CHAGES/3715*` was added in #3715 but I guess we should remove this from the changelog? (or figure out how to reference issues without jira).

jerabekjiri:
> Oh, and one more thing .. did we forget to clean up CHANGES/ on master after cutting 4.8? https://github.com/ansible/ansible-hub-ui/pull/3715 was already in 4.8.0 so it probably shouldn't even be in 4.9.0 changelog - guessing there may be more?

I might have forgotten about this. Also `CHANGES.rst` is missing 4.8.0 changes for some reason. I'll try to fix that

himdel (approved):

PR #4874, opened 2024-02-29
Namespace form: add warning that Logo URL must point to a jpeg/png

himdel:
When SVGs are used, the backends uploads the svg to S3, which treats it as application/octet-stream, which, coming from a different domain, the browser rejects as an invalid request for an `` tag. These would still work when not using S3 (or when telling s3 about the right type on upload), but the users can try at their own risk - adding the warning here.


himdel:
marking as draft, pending final resolution backend-side
re-evaluate the backport-4.9 label after

himdel:
Made part of #4866 , closing this one

PR #5106, opened 2024-06-10
Documentation link - move from left nav to top nav in standalone mode

himdel:
Issue: AAP-25057

Documentation link - move from left nav to top nav in standalone mode
(community Documentation link is still supposed to be the last item in left nav AFAIK - #4991)


himdel:
@tiyiprh does this match your expectations? :)

tiyiprh:
@himdel Looks good! Can we just make the documentation link the first item to be consistent with the other services? So it would be Documentation, Customer Support, Training, and About for Hub

himdel:
Thanks, sure can! :) Updated :)

PR #5276, opened 2024-10-15
Fix webpack-dev-server options

himdel:
Bump @redhat-cloud-services/frontend-components-config-utilities from 3.2.2 to 3.2.3 switched from using `onBeforeSetupMiddleware` to `setupMiddlewares` (https://github.com/RedHatInsights/frontend-components/pull/2070)

so we don't need that fix anymore,
but it was also removing devServer.https, which is still used, deleting as before, with fixed condition.

(stable branches are not affected, no frontend-components-config)

### DIFF
--- a/config/webpack.base.config.js
+++ b/config/webpack.base.config.js
@@ -226,15 +226,8 @@ module.exports = (inputConfigs) => {
   );
 
   // webpack-dev-server 5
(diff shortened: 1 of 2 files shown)


himdel:
Merging to get dev env working again.

PR #5277, opened 2024-10-15
[stable-4.10] CI fixups

himdel:
* pr-checks: install python to unbreak pip install
* cypress: remove community test breaking the matrix check


himdel:
(merge once 4.10 is unfrozen, should unblock dependabot PRs too)

PR #5292, opened 2024-10-31
[stable-4.9] AboutModal: remove links

himdel:
Issue: AAP-34966
Closes #5291

This is an alternative to #5291 - that one fixes the repo link when using tags,

This one removes all external links from the about modal instead.


himdel:
Discussed with John, merging this one, for 4.9 specifically.

PR #5415, opened 2025-01-21
Plugin option choices: both 'bool' and 'boolean' are valid types for plugins, and should be treated the same

felixfontein:
This bug was reported in https://github.com/ansible-collections/community.general/pull/9573#discussion_r1922838554.

Example: https://galaxy.ansible.com/ui/repo/published/community/general/content/inventory/iocage/#parameters option "leading_separator", which is a boolean with type `boolean` with default `false`. "trailing_separator", on the other hand, is a boolean of type `bool` with default `true`.

### DIFF
--- a/CHANGES/5415.bug
+++ b/CHANGES/5415.bug
@@ -1,0 +1,1 @@
+Collection documentation: treat options of type `boolean` the same as options of type `bool`.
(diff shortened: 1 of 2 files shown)


himdel (approved):
:+1: 

There's one more `bool` mention in `documentedType`, but that one already behaves the same for `bool` and `boolean`. So looks like you found all of them :)

felixfontein:
> There's one more `bool` mention in `documentedType`, but that one already behaves the same for `bool` and `boolean`.

Yeah, that's why I didn't adjust it.

felixfontein:
@himdel @prat98 thanks for reviewing and merging!

PR #5416, opened 2025-01-21
Plugin deprecation messages: process markup

felixfontein:
Fixes #5328.

PR #5417, opened 2025-01-21
Fix zero width of spacer columns indenting suboptions and sub-return values

felixfontein:
The CSS from table.css (coming from some dependency, I guess) overrides `width` with `0` due to the column being `:empty`. This makes the indent almost disappear (you can still see it by noticing that the left border line is a tiny bit thicker).

For example, check out the return values here:

Before:
image

After:
image

### DIFF
--- a/CHANGES/5417.bug
+++ b/CHANGES/5417.bug
@@ -1,0 +1,1 @@
+Collection documentation: fix zero width of spacer columns indenting suboptions and sub-return values.
--- a/src/components/render-plugin-doc.scss
+++ b/src/components/render-plugin-doc.scss
@@ -21,7 +21,9 @@
   width: 100%;
 
   .spacer {
-    width: 20px;
+    // The !important is needed here to avoid the value being overridden
+    // by table.css injected by some dependencies.
+    width: 20px !important;
     border-top: 0;
     border-bottom: 0;
   }


oraNod (approved):
Thanks @felixfontein

felixfontein:
@himdel are you still the right person to ping? I have this and two other bugfix PRs (#5416, #5415) that are waiting for reviews.

himdel:
> are you still the right person to ping? 

@felixfontein I .. have no idea :) Probably not.

Buut ... looking good, restarted tests (those look like just ordinary sporadic failures, nothing caused by this PR) aand approved :)

himdel (approved):

himdel:
@felixfontein Approved all of them :) (One note, all of these issues will likely be present in the aap ui repo as well, might want to fix it there too :))

felixfontein:
> @felixfontein I .. have no idea :) Probably not.

Hmm, do you have any idea who is responsible for Galaxy / this repo now?

> (One note, all of these issues will likely be present in the aap ui repo as well, might want to fix it there too :))

Is there a separate repo for AAP UI that basically does similar things than this one? I assumed that AAP basically uses this repo as well as upstream...

(But then I never used AAP, so :shrug:)

himdel:
None, my apologies. I can try asking around.

That would have been true up to and including 2.4, but the later versions' UI is an unrelated effort. You can see some of it in https://github.com/ansible/ansible-ui/, after which development moved to a private repo. Which likely limits your ability to fix anything there :)

samccann:
@felixfontein thanks for this PR!  We're asking around for who can review and merge so bear with us for a bit :-)

felixfontein:
@samccann thanks!

@prat98 thanks for merging this! Can you also take a look at #5416 and #5415?

felixfontein:
@himdel @prat98 does this and the other PRs need to be cherry-picked to one or more of the `stable-x` branches? Or how do we get the fixes to galaxy.ansible.com?

himdel:
@felixfontein No need to backport, the stable branches are unrelated to galaxy ... and the three merged PRs are already running on galaxy.ansible.com :)

PR #5471, opened 2025-03-27
Fix rendering of empty return doctrings

jborean93:
The existing code checked whether the return docstring was null but does not check if it is iterable. If a collection with 'RETURN: {}' was set it would fail to render the entire docstring. This ensures that both null and {} can be parsed correctly.


prat98 (approved):
LGTM 🚀

PR #5484, opened 2025-04-21
Update galaxy docs link

samccann:
Not sure when the url stopped working, but someone pointed it out to me so here's a working url.


prat98:
LGTM

